PR #3, opened 2017-04-23
Fix styles typo

rosszurowski:
Just stumbled across this codebase a spotted a typo in the styles for `@vx/responsive`. Not sure what impact this'll have, but I figured, if it's in there, it should probably be spelt right :)


hshoff:
Good catch!

PR #17, opened 2017-05-08
Add basic usage to README

Flaque:
Forgive me if this the wrong place to put this. 

If we have a really basic example of how to use VX, it might make it a little easier for new folks to get started using it.

I took the SimpleBar chart in the gallery and stripped it down of everything I could so it would be simple enough to just glance over without causing too much information overload.


hshoff:
Lovely! Thank you 🙌

Issue #9, opened 2017-05-08
[docs] document api of all packages


hshoff:

Flaque:
Hey, this might be a stupid question, but where's the repo for the docs?

hshoff:
Great question. My current plan is to add docs to each `packages/**/Readme.md` then write a script that builds those to a new dir called `docs/` at the vx repo root. Open to suggestions for alternative solutions!

Flaque:
I looked into [react-docgen](https://github.com/reactjs/react-docgen) a bit for linking docs to code. But react-docgen assumes a `class` structure for it's objects and doesn't really work all that well with the existing code. It might be a future thing, but I'm just gonna write docs in the Readme's for the moment.

Flaque:
Is `brush` "stable"-ish right now? There's no longer a gallery version of it. Should I just hold off on writing the docs on it for now?

trainorpj:
Do you need help writing some of the documentation?  I could submit a PR for the [boxplots](https://github.com/hshoff/vx/tree/master/packages/vx-boxplot) docs, if nobody has that action.

I've been watching this project for a while, and I like it a lot.  I'm interested in helping out, if you'll have me!

Flaque:
Do it! :D

williaster:
closed @trainorpj by https://github.com/hshoff/vx/pull/102

williaster:
🙈 didn't mean to close the entire issue

kandros:
These are the packages with missing documentation

- [ ] brush
- [ ] clip-path
- [ ] brush
- [ ] drag
- [ ] event
- [ ] hierarchy
- [ ] responsive
- [ ] tooltip
- [ ] zoom

trainorpj:
I wrote the docs for the boxplot component, and I'm happy to contribute to other utilities/components

(I was a bit swamped before, but I have some time now)

trainorpj:
I just hacked together a little site we can use to keep track of progress on the docs: 

[https://trainorpj.github.io/vx-docs-progress/](https://trainorpj.github.io/vx-docs-progress/)

I figured there are so many packages/components it'd be nice to keep track of things at a high level.

[Link to the repo](https://github.com/trainorpj/vx-docs-progress/blob/master/README.md). Feel free to make changes.  I used d3 but didn't use vx... 😬

hshoff:
Very helpful, thank @trainorpj! I'm hoping for all green by v1

hshoff:
[...]

PR #18, opened 2017-05-13
Add docs for `@vx/point`. Issue #9

Flaque:
I can combine more things into one pull request if you would like. I wanted to make sure that this was the right sort of style that you wanted for docs. Let me know if there's anything I can do to make it better. ❤️


hshoff:
Looks great, thank you!

PR #19, opened 2017-05-13
Add docs for `@vx/scale` package

Flaque:
In general, I just linked to the d3 docs. This isn't the greatest, but I think it's fine for now, right? If not I can add better docs for each one here too.


hshoff:
Woo! Thanks! I'll probably circle back and add a paragraph in the intro about scales map domains to ranges, but that's not a blocker.

hshoff:
#9

PR #22, opened 2017-05-13
Add docs for several different packages.

Flaque:
I added some simple docs explaining an example for annotation and a markdown table for properties similar to what [Material-UI](http://www.material-ui.com/#/components/badge#properties) has for it's docs. 

Since working with markdown tables is kind of annoying, I did two things. First I got the [markdown-table-editor](https://atom.io/packages/markdown-table-editor) package for Atom. Then I [wrote a python script](https://gist.github.com/Flaque/beac1aff6ef313875cc6ef8043faf928) to take the function call params and fill out the name, type, and defaults of the table so there was less busy work.

### DIFF
--- a/packages/vx-annotation/Readme.md
+++ b/packages/vx-annotation/Readme.md
@@ -1,5 +1,52 @@
 # @vx/annotation
 
+
+## LinePathAnnotation
+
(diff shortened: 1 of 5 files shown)


hshoff:
👏 Looks great, thank you!

Flaque:
❤️

PR #24, opened 2017-05-15
Add Marker, Mock-data, and Text Docs 

Flaque:
Issue #9

❤️


hshoff:
Looks great! Once again thank you @flaque for your help documenting vx!

Side note for those reading: the current @vx/marker will turn into the svg <marker /> in the future.

PR #31, opened 2017-05-20
pass in axis label components explicitly

williaster:
@hshoff see what you think of this. Because the number of props for the `Axis` components was getting unwieldy, this PR updates them to expect explicit label components to enable full control over their appearance (though simple tweaks like color now require more code as you can see in the `demo -> axis.js`).

I got rid of some props like `tickK` because it seemed like there was mixed logic for using the `tickK` that was passed in for some computations but still inferring orientation with the `ORIENTATION` symbol...which seemed like it could cause bugs with some combos. If users wanted to customize the `tickTransform` they can still do this.

Tested the params by customizing in all directions:

image

### DIFF
--- a/packages/vx-axis/src/axis/Axis.js
+++ b/packages/vx-axis/src/axis/Axis.js
@@ -19,10 +19,8 @@ export default function Axis({
   tickFormat,
   tickStroke = 'black',
   tickLength = 8,
-  tickLabelPadding = 14,
-  tickOffset = 0,
+  tickLabelOffset = 14,
   tickTransform,
-  axisPadding = 0.5,
   hideAxisLine = false,
   hideTicks = false,
   hideZero = false,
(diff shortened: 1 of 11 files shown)


hshoff:
@williaster very excited about this changeset thank you! 🎉 

want to rebase from vx:master and push to make travis ci happy?

packages/vx-axis/src/axis/Axis.js
```diff
@@ -4,38 +4,45 @@ import { Line } from '@vx/shape';
...
   hideTicks = false,
   hideZero = false,
+  labelComponent = (
+    <text
+      textAnchor="middle"
+      fontFamily="Arial"
+      fontSize={10}
+      fill="black"
+    >
+      default label
```
hshoff: let's nuke the `default label` text, it was a bad decision on my part to have a default label text. In the case where folks want to have no label they have to put `<AxisLeft label={''} />` which is super annoying and unnecessary.

packages/vx-axis/src/axis/Axis.js
```diff
@@ -58,72 +68,66 @@ export default function Axis({
...
+    const labelFontSize = labelComponent.props.fontSize || 10;
+    const tickLabelFontSize = tickLabelComponent.props.fontSize || 10;
+
+    const labelProps = {
+      transform: horizontal ?  '' : `rotate(${tickSign * 90})`,
+      x: horizontal ? range1 / 2 : tickSign * (range0 / 2),
+      y: horizontal ?
+        (tickSign * (tickLength + tickLabelOffset + tickLabelFontSize + (!isTop ? labelFontSize : 0)))
+        : (isLeft ? 1 : -1) * tickSign * (tickLength + tickLabelOffset),
+    };
```
hshoff: i'm not opposed to keeping this as is, but it'd be great to turn this into a `getTransfrom()` function passing in the needed props... maybe then we could pass in a `labelTransform()` prop that let folks override it with their own thing if they wanted... not sold on that idea, just throwing that out there.

williaster:
@hshoff I think master still does not have the tests? 
https://github.com/hshoff/vx/blob/master/package.json#L10

hshoff:
@williaster derp, my fault. let merge my tests branch

hshoff:
@williaster ready for a pull

williaster:
all right this should be good for prime time 🎉

hshoff:
👍

PR #32, opened 2017-05-20
Add script to compile Markdown into HTML docs

Flaque:
- A script to compile Markdown to HTML docs -> Issue #9

This might not be exactly what you meant, but it might help get the docs onto the website for the moment. Let me know your thoughts, I'm not attached to anything in particular and can do something different if you'd like. 

# This PR adds 
- A new top-level folder `scripts`
- A script `docs` that compiles each package's README into an HTML doc with the same name inside a folder called `docs` inside `vx-demo`.
- Alters the `docs.js` file in `vx-demo` to link to each html page.

# Upsides
- It's really easy to make docs! Woo! `$ node scripts/docs/index.js`
- The docs are in the website immediately so they're easy for people who are interested in VX to view right now! :D 

# Downsides

## Static (not React)
The HTML docs are static individual files stored inside of the `static` folder (so mixing react and static stuff).

## No Links
The docs don't have individual links to other docs (or even back to the home page). 

--- 
# Next steps
A better way to do this would probably be to use something like [react-markdown](https://github.com/rexxars/react-markdown) to compile the markdown docs into react components.

### DIFF
--- a/scripts/docs/package.json
+++ b/scripts/docs/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "docs",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "fs-jetpack": "^1.0.0",
+    "marked": "^0.3.6"
+  }
+}
--- a/scripts/docs/package.json
+++ b/scripts/docs/package.json
@@ -7,7 +7,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "author": "",
-  "license": "ISC",
+  "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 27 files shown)


Flaque:
@hshoff Anything this needs?

hshoff:
@flaque sorry haven't had a chance to give this the review that it deserves. I'll take a look this weekend 👍

Flaque:
@hshoff No worries <3 Just wanted to make sure

scripts/docs/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "docs",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
```
hshoff: Mind making this MIT just to be consistent with the rest of the licenses?

Flaque: Will do! :D

PR #42, opened 2017-05-24
[Axis] add `rangePadding`

williaster:
@hshoff in working with bar charts that don't have band scales (e.g., for time), you have to compute the bar width manually and also the offset from the tick position (see [this discussion](https://stackoverflow.com/questions/18835053/d3-js-calculate-width-of-bars-in-time-scale-with-changing-range) on StackOverflow).

Even if you compute this width and offset the `x` position of the _bars_, the scale itself doesn't account for the offset so the `Axis` is missing padding that you get for free with band scales:

I tried a few different ways of doing this including overriding the `Axis` position function,  but I think simply padding the range is a reasonable way to approach it (e.g., VictoryCharts [supports a `domainPadding` prop](http://formidable.com/open-source/victory/docs/victory-chart/#domainpadding)). Let me know what you think.

No `rangePadding` / before:
screen shot 2017-05-23 at 3 50 50 pm

with `rangePadding` = `barWidth / 2`:
screen shot 2017-05-23 at 3 55 21 pm

### DIFF
--- a/packages/vx-axis/src/axis/Axis.js
+++ b/packages/vx-axis/src/axis/Axis.js
@@ -13,6 +13,7 @@ export default function Axis({
   orientation,
   top = 0,
   left = 0,
+  rangePadding = 0,
   stroke = 'black',
   strokeWidth = 1,
   strokeDasharray,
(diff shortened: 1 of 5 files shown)


hshoff:
👍

PR #46, opened 2017-05-31
[vx-responsive] add withParentSize enhancer

williaster:
This PR adds a new `withParentSize` enhancer to `vx-responsive` to allow a chart to get its parent size if `screenSize` doesn't suffice. Let me know if you think a different approach would be better.

Tested with `@data-ui/xy-chart` and with `jest`, yellow is the HOC's container:
withparentsize

```js
const ResponsiveXYChart = withParentSize(({ parentWidth, children, ...rest }) => (
  <XYChart
    theme={theme}
    width={parentWidth}
    height={parentWidth / 1.9}
    {...rest}
  >
    {children}
  </XYChart>
));
```


hshoff:
🙌  LGTM

PR #52, opened 2017-06-04
[Bug fix] Move `scaleUtc` into it's own function

Flaque:
# What the bug was
Currently there's a bug where adding a `scaleUtc=true` to your `scaleTime` function will cause JS to crash:
``` js
// Crashes! 🚨🚨
scaleTime({
  scaleUtc=true, 
  ...
});
```

This was happening because the `time.js` file was calling `scale.scaleUtc();`, which isn't a function defined by d3. However, d3 does have a function called `scaleUtc` which works exactly like the `scaleTime` function but instead returns time vars in UTC format.  

# What this PR does
1. Removes the `scaleUtc` option from the `scaleTime` function in `vx-scale`.
2. Adds a new scale: `scaleUtc` to a `utc.js` file.
3. Updates readme to reflect changes.
4. Adds sanity test for` scaleUtc`.

### DIFF
--- a/packages/vx-scale/test/scaleUtc.test.js
+++ b/packages/vx-scale/test/scaleUtc.test.js
@@ -1,6 +1,6 @@
 import { scaleUtc } from '../src';
 
-describe('scaleTime', () => {
+describe('scaleUtc', () => {
   test('it should be defined', () => {
     expect(scaleUtc).toBeDefined()
   })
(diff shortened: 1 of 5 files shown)


packages/vx-scale/test/scaleUtc.test.js
```diff
@@ -0,0 +1,7 @@
+import { scaleUtc } from '../src';
+
+describe('scaleTime', () => {
```
hshoff: `scaleTime` => `scaleUtc`

Flaque: Oops! Sorry about that. Fixed now.

hshoff:
🙌

Issue #54, opened 2017-06-08
how i can  install 'vx'?


jockw:
when i install vx use 'npm install vx ' ,i am find that  the 'vx' i had installed didn't your 'vx',so how can i install your 'vx'

williaster:
Vx is broken down into multiple packages (`@vx/axis`, `@vx/shape`, etc.) with the idea that you install what you need and can minimize bundle size. 

Also note that because these are all name-spaced packages you need a `@` before vx. so something like `npm install @vx/axis` should work.

hshoff:
Hi @jockw thanks for checking out vx. As @williaster said, vx is a library of packages so you only need to install what you need. 

I'll give this some thought. Going to close this in favor of https://github.com/hshoff/vx/issues/33

jockw:
thankyou

hshoff:
To lower the friction to get started with vx, there is now `@vx/vx`

```bash
npm install --save @vx/vx react react-dom
```

```js
import { scaleLinear, AreaClosed, withTooltip } from '@vx/vx';
console.log(scaleBand);
console.log(AreaClosed);
console.log(withTooltip);

// or put everything under a vx namespace
import * as vx from '@vx/vx';
console.log(vx.scaleBand);
console.log(vx.AreaClosed);
console.log(vx.withTooltip);
```

PR #57, opened 2017-06-08
Add install instructions

Flaque:
We should add an example of how to install the individual packages on the front-facing README to account for issues like #54.


hshoff:
👍

PR #59, opened 2017-06-09
[axis] fix label transform

williaster:
This PR fixes a bug with the (y-axis) label transform logic. For y-axes you often have a pattern where you set the following domain and range to correct for the upper-right `(0,0)` svg coord:

```javascript
domain = [min val, max val];
range = [height, 0] 
// y-axis has `min val` at the bottom and `max val` at the top
```
If you want to have the min value at the _top_ of the scale you actually invert the range to `[0, height]`. The current y-label transform assumes the former and this PR fixes it.

Before:
screen shot 2017-06-09 at 3 02 10 pm

After:
screen shot 2017-06-09 at 3 04 05 pm

Verified that current transforms still work as expected:
screen shot 2017-06-09 at 3 05 54 pm
screen shot 2017-06-09 at 3 05 40 pm

### DIFF
--- a/packages/vx-axis/src/utils/labelTransform.js
+++ b/packages/vx-axis/src/utils/labelTransform.js
@@ -14,11 +14,11 @@ export default function labelTransform({
 
   let x, y, transform = null;
   if (orientation === ORIENT.top || orientation === ORIENT.bottom) {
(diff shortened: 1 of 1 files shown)


hshoff:
👍

PR #61, opened 2017-06-10
[ClipPath] add ClipPath package

williaster:
Let me know what you think about the utility of something like this. I think it's less useful than the `Gradients` package but on par with the `Group` package esp since it gets around creating the `<defs>` etc.

Before:
no-clip

After with `<RectClipPath id="clip"width={innerWidth} height={innerHeight} />`:
with-clip

### DIFF
--- a/packages/vx-clip-path/package.json
+++ b/packages/vx-clip-path/package.json
@@ -23,7 +23,7 @@
     "charts",
     "svg"
   ],
-  "author": "@hshoff",
+  "author": "Chris Williams @williaster",
   "license": "MIT",
   "bugs": {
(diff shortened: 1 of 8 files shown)


Flaque:
This is super cool.

packages/vx-clip-path/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "keywords": [
+    "vx",
+    "react",
+    "d3",
+    "visualizations",
+    "charts",
+    "svg"
+  ],
+  "author": "@hshoff",
```
williaster: 👍  my github username will haunt me foreva 👻

PR #63, opened 2017-06-11
[pattern] Add tests for pattern circles

Flaque:
# Changes
- Moves the embedded anonymous function that created circles into it's own separate function so that it can be easily tested. 
- Added tests for createCircles

# Testing done
- Ran tests.
- Ran Demo to prove that everything still looks fine visually: 

What Demo Looks Like
(Still looks normal)


hshoff:
👍

PR #66, opened 2017-06-14
Replace lodash per-method packages with scoped imports

lencioni:
The per-method packages are zero-dependency modules, so they often
include a bunch of extra code compared to importing from lodash/foo. We
can save some bundle size by using these partial includes instead.

Additionally, if I recall correctly, the per-method packages are all
deprecated and no longer updated, so this is the way to go for lodash
moving forward.


hshoff:
👍

PR #70, opened 2017-06-20
add strokeDashoffset prop to LinePath

jacek213:
This PR adds `strokeDashoffset` prop to `LinePath` component which is helpful for transitions.


hshoff:
👍 will also add support for additional props re: https://github.com/hshoff/vx/issues/71

PR #74, opened 2017-06-25
Get min range for area chart baseline

aocenas:
If range of scale does not start at 0, chart would extend beyond bounds of the scale when yScale(0) is used. This fixes it by taking first element of scales range (which should be the lowest y point on screen)

I was not able to test the code, trying to run tests resulted in lots of errors across all packages. Also even if technically correct I am not sure if this does not brake something outside of my usecase.


hshoff:
Looks good!

PR #91, opened 2017-07-07
[vx-bounds] add package with 'withBoundingRects()' HOC and tests

williaster:
This PR adds the `@vx/bounds` package which contains a `withBoundingRects()` higher-order component that is meant to help with detecting whether a given component is overflowing its parent container. This HOC passes the following props to the component which it wraps:

```javascript
const rectShape = PropTypes.shape({
  top: PropTypes.number.isRequired,
  right: PropTypes.number.isRequired,
  bottom: PropTypes.number.isRequired,
  left: PropTypes.number.isRequired,
  width: PropTypes.number.isRequired,
  height: PropTypes.number.isRequired,
});

export const withBoundingRectsProps = {
  getRects: PropTypes.func,
  rect: rectShape,
  parentRect: rectShape,
};
```
**Tests**
Tested with jest and with `@vx/tooltip` in a separate repository:

**Naive tooltip (before`withBoundingRects()`)**
naive-tooltip

**Smarter tooltip (after `withBoundingRects()`)**
smart-tooltip

### DIFF
--- a/packages/vx-bounds/.babelrc
+++ b/packages/vx-bounds/.babelrc
@@ -1,0 +1,19 @@
+{
+  "presets": ["es2015", "react", "stage-0"],
+  "plugins": [],
+  "env": {
+    "development": {
+      "plugins": [
+        ["react-transform", {
+          "transforms": [{
+            "transform": "react-transform-hmr",
+            "imports": ["react"],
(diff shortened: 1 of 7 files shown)


hshoff:
👏 fantastic work @williaster!

PR #95, opened 2017-07-13
[shape] Add tests for Arc, AreaClosed, & Line

Flaque:
Also fixed a bug where `rgba()` was treated as a function because it wasn't in quotes and therefore would cause an `undefined` error.

### DIFF
--- a/packages/vx-shape/src/shapes/AreaClosed.js
+++ b/packages/vx-shape/src/shapes/AreaClosed.js
@@ -13,7 +13,7 @@ export default function AreaClosed({
   strokeDasharray,
   strokeWidth = 2,
   stroke = 'black',
-  fill = rgba(0,0,0,0.3),
+  fill = "rgba(0,0,0,0.3)",
   curve,
   ...restProps,
 }) {
--- a/packages/vx-shape/package.json
+++ b/packages/vx-shape/package.json
@@ -31,13 +31,14 @@
   },
   "devDependencies": {
     "babel-jest": "^20.0.3",
+    "d3-array": "^1.2.0",
     "enzyme": "^2.8.2",
     "jest": "^20.0.3",
+    "react": "^15.0.0 || 15.x",
     "react-addons-test-utils": "^15.5.1",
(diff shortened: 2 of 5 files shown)


packages/vx-shape/src/shapes/AreaClosed.js
```diff
@@ -13,7 +13,7 @@ export default function AreaClosed({
   strokeDasharray,
   strokeWidth = 2,
   stroke = 'black',
-  fill = rgba(0,0,0,0.3),
+  fill = "rgba(0,0,0,0.3)",
```
williaster: nice catch. not sure if we've thought about adding `.eslint` for the library but I think [single quotes are preferred](https://github.com/airbnb/javascript#strings--quotes) for strings

Flaque: Oh good point! Sorry about that. 

Side question: Is there an beautifier for the Airbnb JS style? Also, would anyone mind if I added the ESLint checks to the travis build?

williaster: beautifier as in an IDE plugin of sorts? I'm not totally sure about that one but harry might know. 

I think adding linting to the root  of the `vx` repo and running it for travis could be a good call for catching things like this (and shouldn't require adding it to every package). the number of errors flagged on the initial pass can be a little rough though and it will flag things like `propTypes` + `defaultProps` not being defined for components but could add them to the ignore list to start. Happy to give this a shot as well / recently set it up for the @data-ui repo.

I'd imagine @harry has probably given this all some thought already, though?

Flaque: Oh, I might beat you to it. 😅

Flaque: Though I did steal your .eslintrc 😜

williaster: 🙌

Flaque: Ho boy. 
screen shot 2017-07-13 at 8 31 59 pm

PR #96, opened 2017-07-14
Double Quotes to Single Quotes

Flaque:
In reference to [this comment](https://github.com/hshoff/vx/pull/95). 

Since this was essentially already CR'd I'm just gonna merge this after a day or so if no one has any issues with it.


hshoff:
No need to wait. LGTM!

PR #97, opened 2017-07-14
Add eslint

Flaque:
Config from @williaster's brilliant dataui.


Flaque:
I can split this into two or just take the eslint if the --fix is too much for one review. We also could use something like [eslint-nibble](https://github.com/IanVS/eslint-nibble) to do this one at a time. Although this feels more like ripping off a bandaid.

packages/vx-axis/src/utils/center.js
```diff
@@ -1,7 +1,7 @@
 export default function center(scale) {
-  var offset = scale.bandwidth() / 2;
+  let offset = scale.bandwidth() / 2;
```
williaster: I think this actually still needs to be `var` because of the closure, did you test the gallery locally to see if there were no errors?

packages/vx-demo/components/gallery.js
```diff
@@ -104,7 +104,7 @@ export default class Gallery extends React.Component {
                 <div className="details">
                   <div className="title">Lines</div>
                   <div className="description">
-                    <pre>{`<Shape.Line />`}</pre>
+                    <pre>{'<Shape.Line />'}</pre>
```
williaster: not sure if the backticks were intentional for styling?

packages/vx-pattern/src/patterns/Circles.js
```diff
@@ -15,7 +15,7 @@ export function createCircles({
   stroke,
   strokeWidth,
   strokeDasharray,
-  className
+  className,
 }) {
   return corners.map(([cornerX, cornerY]) =>
```
williaster: surprised this one didn't complain about no `(` + `)`

packages/vx-responsive/src/enhancers/withParentSize.js
```diff
@@ -28,7 +28,7 @@ export default function withParentSize(BaseComponent) {
 
     resize(event) {
       if (this.container) {
-        var boundingRect = this.container.getBoundingClientRect();
+        const boundingRect = this.container.getBoundingClientRect();
```
williaster: this might be another instance where you need a `var` so it can be used in the closure

hshoff:
Awesome thanks @Flaque, I'm going to be out of town and away from computer this weekend. I will take a closer look and give it a proper review on Monday. No need to split it up.

packages/vx-axis/src/utils/center.js
Flaque: I did not. :'( I might have had too much trust in the --fix. 

That was a dumb mistake. From now on I will always test the gallery locally!

As for the var. Don't lets work in "child" scopes, but not in above scopes?

.eslintignore
```diff
@@ -0,0 +1,5 @@
+**/build/*
+**/node_modules/*
+**/coverage/*
+**/_gh-pages/*
```
Flaque: Good point!

packages/vx-axis/src/utils/center.js
Flaque: Example: https://repl.it/J2dE

Flaque:
I went back and made sure everything still ran in the gallery. No problems as far as I can tell.

packages/vx-responsive/src/enhancers/withParentSize.js
Flaque: I think you can still use const here since the closure is contained within the same closure that `var` is defined in.

In general, I think `let` and `const` work the same way most variable assignments work in other languages, but `var` is the nasty one :'( 

[There's a good SE question about it here.](https://softwareengineering.stackexchange.com/questions/274342/is-there-any-reason-to-use-the-var-keyword-in-es6)

williaster:
Awesome, thanks for functionally testing the UI! 👏

packages/vx-responsive/src/enhancers/withParentSize.js
williaster: I think you're right about the scope, played with it for a while on babel + locally. I remembered having an issue with something similar in a different repo but I think the root cause was something else. thanks for the SE link 👍

Flaque:
I'll go through and resolve the conflicts tonight.

hshoff:
So this got me thinking more than I expected. 

The gist is, linting is great for large codebases with a large number of engineers contributing all the time. At this time, vx finds itself mostly driven 4 contributors at the moment so I think we can get by without adding a big new step like this for now. Linting has a time and place and i'm not sure if vx is ready for it at this stage of development. 

I've been using prettier to format on save for the past few weeks and it's been lovely. But that's mostly because i've lived and breathed the airbnb eslint config for the past 4 years and most of the lint fixes we're seeing are my personal preferences vs the airbnb config. 

For now I think we're fine without it. I don't want to slow down/trip up new contributors just yet. Before the v1 release we can run prettier on the whole codebase and add in the lint config + build step then and we can tend the garden from v1 on. 

Would love to hear your thoughts on this. I've been thinking about it a lot since the weekend.

Flaque:
I love this plan. Prettier is wonderful <3

hshoff:
Yeah let's close this for now and we'll bring the discussion back to life closer to `v1`

Flaque:
Sounds good!

williaster:
this sounds like a good plan to me 👍

Issue #69, opened 2017-06-18
Could we make a vx-util?


Flaque:
There's a couple repeated instances around VX that could probably be put into a separate package. 

For example, [additionalProps in heatmap](https://github.com/hshoff/vx/blob/936d58ab684e65434d5332a5c9c479a36d86d376/packages/vx-heatmap/src/util/additionalProps.js) and [additionalProps in shape]. 

Is there any reason why these aren't in their own package?

hshoff:
> Is there any reason why these aren't in their own package?

Just laziness at the moment. Good point, we should do this. 

There's some more work I'd like to do with `additionalProps` so it's more explicit which props get passed on. And I might revisit the api for how data get's passed to thunks re: https://github.com/hshoff/vx/issues/50#issuecomment-306431147 but I haven't gotten to these two things just yet.

PR #105, opened 2017-07-23
[geo] add package geo

nschnierer:
I want to create a new package **vx/geo** with the ability to draw geoJSON/TopoJSON with different projections. Under the hood the d3-geo library.

This pull request isn't ready to merge now. The documentation isn't complete and a example is missing. I want to finish it in few days. However I think it's good time to get feedback and to review my code.

First example:
bildschirmfoto 2017-07-23 um 02 07 36
The example for vx-demo will be more beautiful ;)

```js
render() {
  const width = 800;
  const height = 600;

  const world = topoJsonClient.feature(topology, topology.objects.units).features;

  return (
      <Mercator
        data={world}
        scale={110}
        translate={[width / 2, height / 2]}
        fill="#bbbbbb"
        projectionFunc={projection => {
          console.log('Get projection to use invert', projection);
        }}
        onClick={data => event => {
          console.log(`Clicked country: ${data.id}`);
        }}
      />
    </svg>
  );
}
```

I really like the vx concept. It's so much easier without d3. Good work ;)


packages/vx-geo/src/components/Projection.js
```diff
@@ -0,0 +1,80 @@
...
+          />
+          {centroid && centroid(path.centroid(feature), feature)}
+        </g>
+      )}
+      {projectionFunc && projectionFunc(currProjection)}
+    </Group>
+  );
+}
+
+Projection.propTypes = {
```
Flaque: <3 <3 <3 Proptypes! :D

packages/vx-geo/src/components/Projection.js
```diff
@@ -0,0 +1,80 @@
...
+  scale: PropTypes.number,
+  translate: PropTypes.array,
+  center: PropTypes.array,
+  rotate: PropTypes.array,
+  precision: PropTypes.number,
+  fitExtent: PropTypes.array,
+  fitSize: PropTypes.array,
+  centroid: PropTypes.func,
+  className: PropTypes.string
+};
```
Flaque: I might be wrong about this, but does Javascript let you add stuff to something "after" you export it? If so => COOL! :D

Flaque:
This is super super cool! Let me know if you need any help with anything <3

packages/vx-geo/src/components/Projection.js
```diff
@@ -0,0 +1,80 @@
...
+  const path = geoPath().projection(currProjection);
+
+  return (
+    <Group className={`vx-${projection}-group`}>
+      {data.map((feature, i) =>
+        <g key={`${projection}-${i}`}>
+          <path
+            className={classnames(`vx-${projection}`, className)}
+            d={path(feature)}
+            {...additionalProps(restProps, feature)}
```
hshoff: we will likely want to pass the `centroid` and maybe the index `i` in the `data` part here.

something like we did with `<Arc />` in `@vx/shape`: https://github.com/hshoff/vx/blob/master/packages/vx-shape/src/shapes/Arc.js#L44

packages/vx-geo/src/util/additionalProps.js
```diff
@@ -0,0 +1,8 @@
+import callOrValue from './callOrValue';
+
+export default function additionalProps(restProps, data) {
+  return Object.keys(restProps).reduce((ret, cur) => {
+    ret[cur] = callOrValue(restProps[cur], data);
+    return ret;
+  }, {});
+}
```
hshoff: this comment is just a reminder for myself to follow up on https://github.com/hshoff/vx/issues/69

packages/vx-geo/src/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as Albers } from './components/Albers';
+export { default as Mercator } from './components/Mercator';
+export { default as Orthographic } from './components/Orthographic';
```
hshoff: minor thing, could we rename the `./components/`dir to `./projections/`? this will help match convention with the other packages.

packages/vx-geo/src/components/Projection.js
```diff
@@ -0,0 +1,80 @@
...
+          />
+          {centroid && centroid(path.centroid(feature), feature)}
+        </g>
+      )}
+      {projectionFunc && projectionFunc(currProjection)}
```
hshoff: What's `projectionFunc` for? just an outlet for rendering out something based on the currProjection?

packages/vx-geo/src/components/Projection.js
hshoff: yup! should be fine https://github.com/hshoff/vx/blob/master/packages/vx-pattern/src/patterns/Pattern.js

packages/vx-geo/src/components/Projection.js
nschnierer: Ok, good idea.

packages/vx-geo/src/components/Projection.js
nschnierer: @hshoff Might be helpful to use invert (https://github.com/d3/d3-geo#projection_invert) x and y coordinates to longitude and latitude. Not a good solution because you can set projection settings with props like scale, translate. Maybe its possible to place it with additionalProps on <Group ... />?

packages/vx-geo/src/components/Projection.js
nschnierer: I like the pattern to place it after it (like https://github.com/facebook/prop-types#usage). Maybe it's better to export the function after that, but I'm not sure.

packages/vx-geo/src/index.js
nschnierer: @hshoff Yeah, sure. It think projections will be great. Is it a good idea to place the Projection.js  into this folder. There is no export for Projection.

nschnierer:
@hshoff @Flaque Thanks a lot for the feedback ;) I will finish it soon.

nschnierer:
## ```vx/geo```
Finished the work. I think that is a good result for the first release ;)

### ```<Mercator />```
bildschirmfoto 2017-07-24 um 00 53 30
The demo uses a topology with low resolution (110 metre, ~213kb).

#### ```<Orthographic />```
vx-geo-orthographic

#### ```<Albers />```
vx-geo-albers


I hope the color of the demo is fine ✌️ 

**Next steps:**
* Add graticule
* Add more projection's
* Add pictures to readme
* Rethinking about the projectionFunc property 🤔

packages/vx-geo/src/projections/Projection.js
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { Group } from '@vx/group';
+import additionalProps from '../util/additionalProps';
+import { geoOrthographic, geoAlbers, geoMercator, geoPath } from 'd3-geo';
+
+// TODO: Implement all projections of d3-geo
+const projectionMapping = {
+  orthographic: geoOrthographic(),
```
williaster: if there are multiple instances on a page of one type of projection, would singleton instances result in mixed or lingering config / props being set on them (e.g., if one set `clipAngle` and another didn't, would they both have the specified `clipAngle`)? 

you could get around it by making these factories as opposed to singletons?

packages/vx-geo/src/projections/Projection.js
```diff
@@ -0,0 +1,90 @@
...
+                ...feature,
+                index: i,
+                centroid: c
+              })}
+            />
+            {centroid && centroid(c, feature)}
+          </g>
+        );
+      })}
+      {/* TODO: Maybe find a different way to pass projection function to use for example invert */}
```
williaster: yeah it's a bummer there is no `projection.copy()` utility analogous to `scale`s 🤔

packages/vx-geo/src/projections/Projection.js
nschnierer: Oh, good point. I think factories are better. So, maybe like this:
```
const projections = () => ({
  orthographic: geoOrthographic(),
  // ...
});
```

hshoff: arrow looks good to me. good catch @williaster!

nschnierer: It's done 👍

packages/vx-geo/src/projections/Mercator.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import Projection from './Projection';
+
+export default function Mercator({ ...restProps }) {
```
hshoff: should have caught this in the first review but this is an unnecessary rest here.

let's go with:

```js
export default function Mercator(props) {
  return <Projection projection="mercator" {...props} />;
}
```

and well need to update the other projections. this isn't a blocker. happy to change this after a merge.

nschnierer: Ugh... Solved it. @hshoff

hshoff:
LGTM!

Flaque:
Open Source is great <3

williaster:
This is a great add, nice work @nschnierer 💯!

PR #108, opened 2017-07-26
[tooltip] add `<TooltipWithBounds />` and PropTypes to `@vx/tooltip` exports

williaster:
This PR adds a`<TooltipWithBounds />` component that wraps the current `<Tooltip />` and is aware of its bounds via the `@vx/bounds` HOC. It also adds some more `PropTypes` to the package.

I found myself wanting a tooltip with this functionality and thought it'd be a useful add here but let me know if you think it's too specialized.

Tested functionally in a linked `@data-ui`

tooltip-with-bounds


hshoff:
🙌

Issue #93, opened 2017-07-12
[npm] What should we do about lockfiles?


Flaque:
Right now it doesn't seem that there's any `package-lock.json` files in any of the packages. But the `lerna exec npm install` will create a lockfile in each package. 

Npm claims that we should commit these. Is there any reason we haven't? 

Could we either add them or add them to the .gitignore?

andyfangdz:
Whether lockfiles make sense for libraries is debatable, but npm seems pretty opinioned on this -- according to the manual they cannot be published. Therefore, we could safely ignore them.

https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md
>One key detail about `package-lock.json` is that it cannot be published, and it will be ignored if found in any place other than the toplevel package. It shares a format with npm-shrinkwrap.json(5), which is essentially the same file, but allows publication. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.

Flaque:
Cool, do we want to add them to the git ignore then?

andyfangdz:
I'm cautiously positive that they can be ignored.

/cc @hshoff

hshoff:
@ljharb any thoughts on how libraries should treat `package-lock.json`?

ljharb:
@hshoff Only apps should have lockfiles. Libs/packages should add `package-lock=false` to `.npmrc`, and gitignore `package-lock.json`, `npm-shrinkwrap.json`, and `yarn.lock`, full stop.

hshoff:
@ljharb makes sense thank you!

PR #114, opened 2017-08-08
Add minimal tests to vx-gradient components

Ffloriel:


hshoff:
Hi @Ffloriel thanks for checking out vx and adding more coverage!

A couple things for this to land:
- Let's hold off on the `package-lock.json` files until https://github.com/hshoff/vx/issues/93 is resolved
- I see you renamed `test() => it()` let's keep `test()` for now to stay consistent with the other packages

Once those are done, I'd be happy to merge.

Ffloriel:
Hi @hshoff 
I was actually trying to be consistent when I changed to it (2 files were using `it`) but I didn't look at the other packages. It makes more sense to use `test` in that situation. I will make the requested changes

hshoff:
@Ffloriel ah that was my fault. thanks for the updates! looks good to me.

PR #116, opened 2017-08-11
add_touchPoint

arusantimo:
Made for mobile use event. : )


hshoff:
Looks good to me. I imagine we'll want to support multiple touches later, but that can come later. This is likely good for most use cases.

PR #121, opened 2017-08-16
fix typos

kandros:
I assume the html in pages are generated by these files, so there no need to edit them


hshoff:
Good catch, thanks!

Issue #124, opened 2017-08-16
sync demo with documentation


kandros:
I noticed  the documentation website misses some components and some are outdated
May I suggest to execute script to generate docs in CI?

I'm looking for others missing

PR #125, opened 2017-08-16
sync vx-demo site documentation with packages

kandros:
as referenced in #124 the demo is not in sync with the documentation

what about create static`/docs` links dynamically so we don't risk to miss them in the future?


kandros:
I inserted all the links to pages that have a readme page with something other than the install instructions

hshoff:
🏆 thank you!

Issue #33, opened 2017-05-21
Improve usability: aggregated package or cli


andyfangdz:
Creating this issue here as a note of what we discussed on Friday:

Right now, to use vx, one has to pull in each of the dependency as individual repositories. To improve the usability of the package (e.g. upgrading to a new version of vx), we can consider two approaches:

1. Create an "aggregated" package, e.g., `@vx/vx`, as an indexing package of all the components in vx. This is the same approach as lodash (https://github.com/lodash/lodash#installation).

 2. Create a CLI that manages `@vx/*`.

PR #131, opened 2017-08-18
[vx] add one stop install pkg @vx/vx

hshoff:
Fixes https://github.com/hshoff/vx/issues/33. Thanks to @andyfangdz for the suggestion.

If you're not worried about bundle size or you just want to play around with vx packages without having to install each one you can now do:

```bash
npm install --save @vx/vx
```

and then use it like so

```js
import { scaleLinear, AreaClosed, withTooltip } from '@vx/vx';
console.log(scaleBand);
console.log(AreaClosed);
console.log(withTooltip);

// or put everything under a vx namespace
import * as vx from '@vx/vx';
console.log(vx.scaleBand);
console.log(vx.AreaClosed);
console.log(vx.withTooltip);
```


williaster:
this is great! 🎉

PR #136, opened 2017-09-03
add readme for vx/hierarchy

trainorpj:
I have a few notes/questions:
1. It appears that `className` is passed as a prop to both components, but is never actually called, so I didn't include it in the props table.  Am I missing something here?
2. The props `size`, `nodeSize`, and `separation` are all passed to the respective d3 layouts, and that's where the defaults are being set (not in the vx components).  Should we mention this, or is it not a big deal?
3. Should this readme include more details on `nodeComponent` and `linkComponent`?  Is there a plan to have defaults for these, like in [vx/network](https://github.com/hshoff/vx/tree/master/packages/vx-network)?

[Link to components for convenience.](https://github.com/hshoff/vx/tree/master/packages/vx-hierarchy/src/hierarchies)

Let me know if anything needs clarification!

### DIFF
--- a/packages/vx-hierarchy/Readme.md
+++ b/packages/vx-hierarchy/Readme.md
@@ -3,3 +3,35 @@
 ```
 npm install --save @vx/hierarchy
 ```
+
+This package contains two components, `<Tree />` and `<Cluster />`, both of which render trees as a node-link diagram.
+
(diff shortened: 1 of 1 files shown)


hshoff:
Awesome thank you @trainorpj!

1. This is a bug. Will fix.
2. Also a bug. Will fix.
3. Good catch, should probably default them like `@vx/network`

Issue #143, opened 2017-09-10
Add module support for tree shaking


dylanmoz:
Add support for the `"module"` field in package.json so that webpack/rollup can do proper tree shaking. If you use @vx currently with Rollup, you will get an error by default due to the only availability being commonjs that it cannot understand, so you have to add all the exports of each sub-library of @vx explicitly in rollup.config.js:

https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports

The result is that if you import one part of a `@vx/shape`, for example, you have to import its entirety.

I believe this can be as simple as building each library twice: Once as it currently is (default es5 transpiling by babel), and a second build into another folder, such as `module`, where the `modules` flag is turned off for babel in the babelrc:
```
"presets": [["es2015", { "modules": false }], "react", "stage-0"],
```
https://babeljs.io/docs/plugins/preset-es2015/#options

hshoff:
Ah yes, i've been putting off getting this going. I'll take a swing at this soon. Will definitely have it sorted out by v1.0.0.

ljharb:
The "module" field is nonstandard and should not be used.

"Proper tree-shaking" should not, and does not, require ES Modules; it can work equally well with (static, top-level) CJS - although popular bundlers may not have chosen to build support for that yet.

Using deep paths is always the best practice.

Separately, if you want to follow the actual thing that node will be doing (which is what the ecosystem will also end up doing), you should generate `.mjs` files alongside all your `.js` files, and the `.mjs` files need to use only syntax that's in node v8.5 (or whatever version ends up first landing unflagged ESM support).

techniq:
We might also want to look into adding `pure-module: true` to our modules for Webpack 4.  See d3 [issue](https://github.com/d3/d3/issues/3131) for details

ljharb:
@techniq i believe it's now `sideEffects: false`.

hshoff:
more info: https://github.com/webpack/webpack/tree/next/examples/side-effects
excited for this.

knoopx:
Three shaking won't work using absolute paths to components on the app side unless the library also uses absolute paths within the library code.

ljharb:
@knoopx tree-shaking isn’t *needed* when using deep paths (absolute paths - starting with / - are never a good idea)

knoopx:
Yes I was referring to "deep paths". The library should use deep paths too in order to be able to tree shake on the app side.

knoopx:
For example I'm using only these modules:

```
[...]

PR #144, opened 2017-09-10
ES6 Module Support Proposal

dylanmoz:
Relates to https://github.com/hshoff/vx/issues/143

One way of doing it. I couldn't find a simpler way of changing the `modules: false` field for two different builds. The idea would be that this format would be applied across all the libraries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 build
+module
 bundle.js
 test/tmp
 *.log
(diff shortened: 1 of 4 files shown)


amilajack:
What's the status of this?

hshoff:
Hi @amilajack see https://github.com/hshoff/vx/issues/143

For now, it's recommended to use deep path imports:

```js
import LinePath from '@vx/shape/build/shapes/LinePath';
```

closing this for now, but will continue tracking in https://github.com/hshoff/vx/issues/143

Issue #139, opened 2017-09-04
update to vx/hierarchy


trainorpj:
This is in reference to #136, which concerns the `<Tree />` and `<Cluster />` components.  [Links to components for convenience](https://github.com/hshoff/vx/tree/master/packages/vx-hierarchy/src/hierarchies).

@hshoff addressed these concerns, but I wanted to list them here so we don't lose track of them.  These apply to both components.

1. `className` is passed as a prop, but is never used in the component.
* **Solution:** That depends where className should be... I'm not sure, but probably in the `<Group />` component?
2. `size`, `nodeSize`, and `separation` are all passed as props, but they are passed to their respective d3 layouts (`d3.tree()` and `d3.cluster()`).  The defaults are not set in the components, but the d3 layouts sets defaults automatically.
* **Solution:** Set the defaults to be whatever we want, even if we hard-code them to be the d3 defaults... or just leave it be.  I'm not sure what's best.
3. `nodeComponent` and `linkComponent` are passed as props, but they have no defaults.  You can see that [vx/network](https://github.com/hshoff/vx/tree/master/packages/vx-network/src) has default components for these.
* **Solution**: Just use the same defaults that are used in `vx/network`.  They can just be put in `vx/hierarchy/src`.

I'm happy to make a PR for any of these!  Just give me the word and I'll go ahead and do it.  I can keep track of the docs when changes are made.

These aren't urgent problems, but they're worth keeping track of 😄

PR #146, opened 2017-09-10
Update hierarchy comps

trainorpj:
This is related to #139 

This PR does the following (each statement applies to both `<Tree />` and `<Cluster />` components:

1. Adds `className` to each `<Group />` component.
* Is this where the `className` prop should go?  I didn't see any other reasonable place to put it.
2. Adds `<DefaultLink />` and `<DefaultNode />` components to each component, along with basic tests for each.
* These render just fine--let me know if there are issues with them.
* I don't know much about testing components, so let me know if we need more comprehensive tests right now 🃏 

This PR **does not**

3. Add defaults to the `size`, `nodeSize`, and `separation` props.
* The defaults are set in the corresponding d3 functions.   The questions is, are we setting defaults at the highest or lowest level they're called at?  
  * For example, there is no default for the `top` prop in either component--the prop gets passed to `<Group />` which **does** have a default for `top`.  So the default is set there (that's what I mean by "lowest level". I think this is the best way to do it.
4. Update the Readme.  
* I can do this once the above points are addressed (let me know and I can add them before you merge 👍)

This is my first contribution to the codebase outside of documentation, so let me know if there's something I missed.  For example, it doesn't have today's commits to master (I probably need to do that, right? 😬).

### DIFF
--- a/packages/vx-hierarchy/src/hierarchies/Cluster.js
+++ b/packages/vx-hierarchy/src/hierarchies/Cluster.js
@@ -25,7 +25,7 @@ export default function Cluster({
   const links = data.links();
   const descendants = root.descendants();
   return (
-    <Group top={top} left={left} className={className}>
+    <Group top={top} left={left} className={cx("vx-cluster", className)}>
       {linkComponent &&
         links.map((link, i) => {
           return (
@@ -37,7 +37,7 @@ export default function Cluster({
       {nodeComponent &&
         descendants.map((node, i) => {
           return (
-            <Group key={`cluster-node-${i}`} className={className}>
+            <Group key={`cluster-node-${i}`}>
               {React.createElement(nodeComponent, { node })}
             </Group>
           );
--- a/packages/vx-hierarchy/src/hierarchies/Cluster.js
+++ b/packages/vx-hierarchy/src/hierarchies/Cluster.js
@@ -1,7 +1,7 @@
-import React from 'react';
-import cx from 'classnames';
-import { Group } from '@vx/group';
-import { cluster as d3cluster } from 'd3-hierarchy';
+import React from "react";
+import cx from "classnames";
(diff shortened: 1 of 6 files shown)


hshoff:
Looks good to me!

packages/vx-hierarchy/src/hierarchies/Cluster.js
```diff
@@ -23,21 +25,23 @@ export default function Cluster({
...
-        );
-      })}
-      {nodeComponent && descendants.map((node, i) => {
-        return (
-          <Group key={`cluster-node-${i}`}>
-            {React.createElement(nodeComponent, { node })}
-          </Group>
-        );
-      })}
+    <Group top={top} left={left} className={className}>
```
hshoff: let's change this to

```js
className={cx('vx-cluster', className)}
```

that way it has a default class and folks can add their own

packages/vx-hierarchy/src/hierarchies/Tree.js
```diff
@@ -23,21 +25,23 @@ export default function Tree({
...
-        );
-      })}
-      {nodeComponent && descendants.map((node, i) => {
-        return (
-          <Group key={`tree-node-${i}`}>
-            {React.createElement(nodeComponent, { node })}
-          </Group>
-        );
-      })}
+    <Group top={top} left={left} className={className}>
```
hshoff: same here

```js
className={cx('vx-tree', className)}
```

packages/vx-hierarchy/src/hierarchies/Cluster.js
trainorpj: Got them both.  I also found a (small) bug in my code, adding className where it has no business being 🐛

trainorpj:
I'll update the readme when this is merged (I need to have working links to the components).

PR #161, opened 2017-10-01
Vx axis tests

Ffloriel:
- add unit test with scaleBand instead of scaleLinear
- tickStroke has a default value, making stroke unreachable (line 135)?


hshoff:
Thanks for this @Ffloriel! I need to take a look at why an unrelated test is broken. Will get this merged soon.

Issue #162, opened 2017-10-03
Collapsible tree example


techniq:
I derived a simple example from https://vx-demo.now.sh/trees available here: https://codesandbox.io/s/n3w687vmqj

I'm not overly happy with the `forceUpdate()` and would like to have it animate similar to https://bl.ocks.org/mbostock/4339083, but it's a start

techniq:
@sghall - any tips on how to use react-move to produce something similar to https://bl.ocks.org/mbostock/4339083?

techniq:
@hshoff / @sghall - It seems like to be able to make this work we would need to be able to add a `<NodeGroup />` [here](https://github.com/hshoff/vx/blob/master/packages/vx-hierarchy/src/hierarchies/Tree.js#L28-L45)

Maybe if `<Tree/>` / `<Cluster/>` receives a children function, we call it instead?  Such as:

```js
<Tree
  top={margin.top}
  left={margin.left}
  root={hierarchy(data, d => d.isExpanded ? d.children : null)}
  size={[
    height - margin.top - margin.bottom,
    width - margin.left - margin.right
  ]}
>
  ({ links, descendants }) => (
    // magic here
  )
</Tree>
```

hshoff:
@techniq yup that’s a great solution. I’ve been busy with interactions, will get to this when i can. Happy to a review a PR if anyone wants to take a swing at this.

techniq:
@hshoff No promises, but I'll see if I can get a PR pushed up soonish (should be fairly trivial) but currently this is me just kicking around vx so my other tasks have higher priority.

Seems we would want the same with [vx-network](https://github.com/hshoff/vx/tree/master/packages/vx-network) and maybe other layouts.  I've only recently been getting into d3 beyond the basics so I'm not aware of all the different layouts and their intricacies and how vx has integrated them.

I would love to hear if @sghall has any feedback on this with regards to react-move.

Lastly, I wonder if the child function / render prop should be embraced whole-heartedly throughout vx, in place of some HOCs (such as replacing `withTooltip`, `withParentSize`, `withScreenSize`, ...).  I've had good success with it in my [react-fetch-component](https://github.com/techniq/react-fetch-component) and [downshift](https://github.com/paypal/downshift), and I know Michael Jackson has been a big proponent of it [lately](https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce)

techniq:
[...]

PR #163, opened 2017-10-05
[hierarchy] add render prop to <Tree /> + <Cluster />

hshoff:
More control over rendering. Fix for https://github.com/hshoff/vx/issues/162

```js
<Tree root={someHierarchy}>
  {({ links, descendants, data, root }) => {
    // up to you
  }}
</Tree>
```

cc @techniq


techniq:
👍 thanks.  I'll try to take a look at this with react-move in the next few days and let you know how it goes.  I'm going out of town this weekend so it might be early next week before I get a chance.

PR #173, opened 2017-10-12
Add Pack, Partition, and Treemap hierarchical layouts.  Closes #172

### LINKED ISSUE #172: Additional hierarchy layouts
I created three new vx components for the remainder of the d3-hierarchy layouts.

## `<Pack />`
- [Component and example](https://codesandbox.io/s/y3rozj8m4x)
- [d3 example ported from](https://bl.ocks.org/mbostock/4063530)

## `<Partition />`
- Grid
  - [Component and example](https://codesandbox.io/s/m36nojzlx)
  - [d3 example ported from](https://bl.ocks.org/mbostock/2e73ec84221cb9773f4c)

- Sunburst
  - [Component and example](https://codesandbox.io/s/jp2zww0knv)
  - [d3 example ported from](https://bl.ocks.org/mbostock/4063423)

## `<Treemap />`
- [Component and example](https://codesandbox.io/s/w74309l5p7)
- [d3 example ported from](https://bl.ocks.org/mbostock/4063582)

## Recommended API change
- Only pass `data` to children render function for all hierarchy components instead of calling `data.links()`, `data.descendants()`, `data.leaves()` and passing each of them explicitly.  I haven't measured, but I'm expecting an incurred cost for each call, and don't expect a user to need both `descendants` or `leaves` in the same visualization, for example.

I'll try to submit a PR with these 3 new layout components to vx-hierarchy package, but maybe someone else could create some prettier examples for the gallery than my d3 ported ones?  It would also probably be good to put some hierarchical data in vx-mock (maybe the `flare.json` that d3 uses).  It would also be nice to probably create a data generator for this type of data.

I'm next going to start looking at creating an orthogonal link, useful for [brackets](http://bl.ocks.org/jdarling/2503502) and possible a curved layout (see this [vega](https://vega.github.io/vega/examples/radial-tree-layout/) example and change the link type).  A 45deg [elbow](https://bl.ocks.org/tlfrd/3dedf0369f7d67843be17961550218d9) would be nice as well.  I'm not sure if all of these should be separate components like `LinkHorizontal`, `LinkVertical`, and `LinkRadial` or just a `type` prop (or something) such as `<LinkHorizontal type="orthogonal" />`.  Thoughts?

techniq:
I also updated `Tree` and `Cluster` to only pass a single param to the child render function.  Based on [this](https://github.com/d3/d3-hierarchy#hierarchy):

image

I thought it made sense to swap `data` and `root` names, so now a layout component takes in a `data` prop (instead of `root`) and the child render function receives a `root` prop (instead of `data`, along with removing `descendants`, `links`, etc).

This also makes it more consistent with `BarGroup`, `BarStack`, `Projection`, etc.  I can be talked out of this, but it seemed like it made sense, and since we were already making a breaking change with the number of params passed to the children function, I thought it was worth changing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,18 +6,30 @@
   "scripts": {
     "test": "lerna exec npm install && jest",
     "docs": "node ./scripts/docs/index.js",
-    "prepare-release":
-      "git checkout master && git pull --rebase origin master && npm run docs && lerna updated",
+    "prepare-release": "git checkout master && git pull --rebase origin master && npm run docs && lerna updated",
     "release": "npm run prepare-release && lerna publish --exact"
   },
   "jest": {
-    "projects": ["<rootDir>/packages/*"],
+    "projects": [
+      "<rootDir>/packages/*"
+    ],
     "collectCoverage": true,
     "coverageDirectory": "<rootDir>/coverage",
-    "coveragePathIgnorePatterns": ["/node_modules/"],
-    "coverageReporters": ["text", "lcov"]
+    "coveragePathIgnorePatterns": [
+      "/node_modules/"
+    ],
+    "coverageReporters": [
+      "text",
+      "lcov"
+    ]
   },
-  "keywords": ["react", "d3", "visualization", "vx", "charts"],
+  "keywords": [
+    "react",
+    "d3",
+    "visualization",
+    "vx",
+    "charts"
+  ],
   "author": "@hshoff",
   "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


hshoff:
This is great. Happy to merge it. Tests are failing on travis right now, so don't worry about that. I'll get that fixed soon https://github.com/hshoff/vx/issues/171.

Let's keep it root over data. My reasoning is you're getting back a hierarchy root node that has all the special methods like `descendants`, `links` that wouldn't be on a raw data object like an array of objects. Once that change is made I'll get this merged. Make some new tiles/add mock data and cut a `0.0.143` release.

techniq:
Done

PR #183, opened 2017-10-31
🚀 [shape] add `<Area />` and tests

williaster:
#### :rocket: Enhancements
This PR adds a new `<Area />` component (+ tests) to the `@vx/shape` package to support confidence bands, which the current `<AreaClosed />` component doesn't support. I left the current `<AreaClosed />` component in addition to the new component to avoid breaking changes and because I think there's merit having specialized components for this.

Tested with jest and functionally / visually with a linked `@data-ui` library:
screen shot 2017-10-31 at 11 00 32 am

### DIFF
--- a/packages/vx-shape/build/index.js
+++ b/packages/vx-shape/build/index.js
@@ -67,6 +67,15 @@ Object.defineProperty(exports, 'LinkRadial', {
   }
 });
 
+var _Area = require('./shapes/Area');
+
+Object.defineProperty(exports, 'Area', {
+  enumerable: true,
+  get: function get() {
+    return _interopRequireDefault(_Area).default;
+  }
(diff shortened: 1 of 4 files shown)


packages/vx-shape/build/index.js
```diff
@@ -67,6 +67,15 @@ Object.defineProperty(exports, 'LinkRadial', {
   }
 });
 
+var _Area = require('./shapes/Area');
```
williaster: not sure how the build file got in, but looks like it's been there for a while. let me know if you want me to remove it here!

hshoff: 🤦‍♂️ my fault. I'll fix this.

PR #184, opened 2017-10-31
[axis] less restrictive tickValue propTypes

williaster:
#### :bug: Bug Fix
not realllly a bug, but along those lines this PR makes all `<Axis* />` component `tickValues` `propTypes` less restrictive. currently if you were to specify an array of dates for a timeline axis you would get console warnings.


hshoff:
😍

Issue #174, opened 2017-10-12
Support additional link types


techniq:
Creating a separate issue to track, mentioned in #172.

It would be nice to support more link types, similar to vega (see [tree example](https://vega.github.io/vega/examples/tree-layout/), [radial tree example](https://vega.github.io/vega/examples/radial-tree-layout/)), such as:

## Orthogonal
- Useful for [brackets](http://bl.ocks.org/jdarling/2503502)
- Another [example](http://bl.ocks.org/goodmami/fd03b250588e1e1143bd)
![image](https://user-images.githubusercontent.com/177476/31474840-486e8d4e-aecb-11e7-94b4-c69f604ba279.png)

![image](https://user-images.githubusercontent.com/177476/31474806-19a3072e-aecb-11e7-9e87-8fadc14a190c.png)
![image](https://user-images.githubusercontent.com/177476/31474784-fc8f4350-aeca-11e7-81ff-14d2de8f55ca.png)

## Curve
- [example](https://bl.ocks.org/mbostock/4600693)
![image](https://user-images.githubusercontent.com/177476/31474770-eb1c3d9e-aeca-11e7-8ed8-64035f75ffff.png)
![image](https://user-images.githubusercontent.com/177476/31474815-28e535ae-aecb-11e7-82d4-81d622fae82e.png)

also possible
## Elbow (45deg)
- [example](https://bl.ocks.org/tlfrd/3dedf0369f7d67843be17961550218d9) 
- useful for annotations / call outs
![image](https://user-images.githubusercontent.com/177476/31474917-b9b7c344-aecb-11e7-9c29-1d5da9b29133.png)

## Implementation
I'm not sure if all of these should be separate components like `LinkHorizontal`, `LinkVertical`, and `LinkRadial` (`LinkHorizontalOrthogonal` ?) or by adding a `type` prop (or something) such as `<LinkHorizontal type="orthogonal" />` where type could be `line`, `diagonal`, `curve`, `orthogonal`, `elbow`, etc.

[...]

PR #194, opened 2017-11-09
Add support for step, curve, and line links.  Issue #174

techniq:
#### :rocket: Enhancements
Adds new Step, Curve, and Line links for cartesian (vertical and horizontal) and radial/polar, while retaining the original links ("diagonal") as is (no renaming).

See this [codesandbox](https://codesandbox.io/s/n3w687vmqj) for examples.  We could also update our [tree](http://localhost:3000/trees) example to allow switching types or possibly create a few new examples.  See issue #174 for possible inspirations as well.

I plan to still implement an Elbow type as well (see this [example](https://bl.ocks.org/tlfrd/3dedf0369f7d67843be17961550218d9) and the elbow example of [react-annotation](http://react-annotation.susielu.com/)) but it probably won't be soon-ish.


packages/vx-shape/src/shapes/link/curve/LinkRadialCurve.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const sx = sr * sc;
+    const sy = sr * ss;
+    const tx = tr * tc;
+    const ty = tr * ts;
+
+    const dx = tx - sx;
+    const dy = ty - sy;
+    const ix = 0.2 * (dx + dy);
+    const iy = 0.2 * (dy - dx);
```
hshoff: should the `0.2` be configurable here?

packages/vx-shape/src/index.js
```diff
@@ -3,9 +3,18 @@ export { default as Pie } from './shapes/Pie';
 export { default as Line } from './shapes/Line';
 export { default as LinePath } from './shapes/LinePath';
 export { default as LineRadial } from './shapes/LineRadial';
-export { default as LinkHorizontal } from './shapes/LinkHorizontal';
-export { default as LinkVertical } from './shapes/LinkVertical';
-export { default as LinkRadial } from './shapes/LinkRadial';
+export { default as LinkHorizontal } from './shapes/link/diagonal/LinkHorizontal';
+export { default as LinkVertical } from './shapes/link/diagonal/LinkVertical';
+export { default as LinkRadial } from './shapes/link/diagonal/LinkRadial';
```
hshoff: Will probably need to export `./shapes/link/diagonal/{LinkHorizontal, LinkVertical, LinkRadial}` from `./shapes/{LinkHorizontal, LinkVertical, LinkRadial}` else it's a breaking change for folks deep linking to the components

techniq: I tried to fix this but was running into problems (I don't know if it was from Lerna or something else).

For example, I recreated `src/shapes/LinkHorizontal.js` and added `export * from './link/diagonal/LinkHorizontal';` to re-export from this location, but when I tried up update the reference in `vx-demo` `tree.js` from

```js
import { LinkHorizontal } from '@vx/shape';
```
to
```js
import LinkHorizontal from '@vx/shape/shapes/LinkHorizontal'
```
I'm getting a dependency error.  

image

I tried running `npm run build` within `vx-shape` and restarting `npm run dev` on `vx-demo.

Any ideas?

hshoff: does this work? 

```js
// src/shapes/LinkHorizontal.js
export { default as LinkHorizontal } from './shapes/link/diagonal/LinkHorizontal';
```

packages/vx-shape/src/shapes/link/curve/LinkRadialCurve.js
techniq: I played around with this but I didn't see much benefit as this is being built using [cubic bezier curves](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths#Bezier_Curves) (`C` path command) that I ported from Vega.  We would probably need to expose both of these `0.2` separately, but I wasn't sure what to call them, etc.

I took a quick stab at using a simpler quadratic curve (`Q` path command) where a single "midpoint" would define the curve but I was having trouble creating the same curve.

I've kicked around the idea in my head a bit and part of me thinks exposing `<HorizontalLink curve={curve} />`, `<VerticalLink curve={curve} />` and `<RadialLink curve={curve} />` and allow passing the curve function similar to show [d3-shape](https://github.com/d3/d3-shape#curves) does this.  This is basically what my `line` functions are doing, such as...

```js
const step = (source, target) => {
    const sa = x(source) - Math.PI / 2;
    const sr = y(source);
    const ta = x(target) - Math.PI / 2;
    const tr = y(target);

    const sc = Math.cos(sa);
    const ss = Math.sin(sa);
    const tc = Math.cos(ta);
    const ts = Math.sin(ta);
    const sf = Math.abs(ta - sa) > Math.PI ? ta <= sa : ta > sa;
    
    return `
      M${sr * sc},${sr * ss}
      A${sr},${sr} 0 0,${sf ? 1 : 0}
      ${sr * tc},${sr * ts}
      L${tr * tc},${tr * ts}
    `;
  };
```

I haven't looked to much into the details on how to implement this (looks like you you're supposed to use the [custom curves](https://github.com/d3/d3-shape#custom-curves) api).  This might also be overkill for 2 points...

I still think exposing the simple `LinkHorizontalStep` is the way to go, this is more for the customization beyond the defaults...

packages/vx-shape/src/index.js
techniq: It does not.

After that change, I also tried importing using

```js
import LinkHorizontal from '@vx/shape/shapes/LinkHorizontal'
```
and
```js
import LinkHorizontal from '@vx/shape/LinkHorizontal'
```

hshoff: No worries, i'll take a look after merging.

PR #198, opened 2017-11-13
 [responsive] add `<ParentSize />` component. fixes #192

### LINKED ISSUE #192: [responsive] manually trigger resize
Currently when using [withParentSize](https://github.com/hshoff/vx/blob/master/packages/vx-responsive/src/enhancers/withParentSize.js), if the parent changes size NOT from resizing the window (ex. show/hide side drawer that resizes the width of the main content), the width is not recalculated.

Need a way to force the component call `resize()` (based on a prop change, or expose the actual `resize` function to the user).  Maybe if we use a render function (which would be nice), something like...

```jsx
<ParentSize>
{({ parentWidth, parentHeight, resize }) => (
  ...
  <button onClick={resize} />
)}
</ParentSize>
```

or maybe add a `componentWillReceiveProps` to ParentSize and any prop change will cause a resize to be called.  In my case, I have a `drawerWidth` prop that could trigger this.

```jsx
<ParentSize drawerWidth={drawerWidth}>...</ParentSize>
```

hshoff:
responsive-resizeobserver

#### :rocket: Enhancements

- [responsive] add `<ParentSize />` component that uses [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) polyfill. fixes: https://github.com/hshoff/vx/issues/192

#### :house: Internal

- [demo] update gallery tiles to use `<ParentSize />` 
- [demo] add /responsive gallery tile + page

### DIFF
--- a/packages/vx-demo/components/gallery.js
+++ b/packages/vx-demo/components/gallery.js
@@ -33,6 +33,7 @@ import Streamgraph from '../components/tiles/streamgraph';
 import Pack from '../components/tiles/pack';
 import Treemap from '../components/tiles/treemap';
 import Radar from '../components/tiles/radar';
+import Responsive from '../components/tiles/responsive';
 
 const items = [
   '#242424',
@@ -53,7 +54,6 @@ export default class Gallery extends React.Component {
   constructor(props) {
     super(props);
   }
-
   render() {
     const detailsHeight = 76;
     return (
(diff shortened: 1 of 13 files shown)


techniq:
🎉

williaster:
💯 !!!

PR #204, opened 2017-11-21
[tooltip] fix tootlip with bounds offset

williaster:
#### :bug: Bug Fix
This PR fixes a pretty bad and somewhat esoteric bug for `<TooltipWithBounds />` in the `@vx/tooltip` package. Since [my PR](https://github.com/hshoff/vx/pull/193) that introduced optional padding/offsets for this component, it's been acting weird on the edges of containers:

@conglei and I debugged this earlier today and what we think is happening is that the tooltip `width` is being compressed on the edge of charts (when `word-wrap` is set to allow wrapping) like the pic below, making its `width` smaller and `height` larger (note this gif has boundary detection disabled). 

This smaller `width` and larger `height` _are_ reflected in the `withBoundingClientRect`s we get, but it seems because the tooltip is actually offset from the edge of the chart it does not visually appear to change dimensions. This means that our calculation for the offset includes a smaller-than-actual-`width` and larger-than-actual-`height`, making the tooltip go all over (if you pay close attn to the broken gif, it's consistent with this).

The fix was to switch to using css transforms instead of `top` and `left`. We set `top` to `0` meaning that the `withBoundingClientRect`s dimensions are always in sync with the rendered tooltip:

cc @techniq

### DIFF
--- a/packages/vx-tooltip/src/tooltips/TooltipWithBounds.js
+++ b/packages/vx-tooltip/src/tooltips/TooltipWithBounds.js
@@ -36,7 +36,7 @@ function TooltipWithBounds({
   }
 
   return (
-    <Tooltip style={{ top, left, ...style }}>
+    <Tooltip style={{ top: 0, transform: `translate(${left}px, ${top}px)`, ...style }}>
       {children}
     </Tooltip>
   );


techniq:
LGTM, and should have better perf by not requiring a layout or paint (https://csstriggers.com/transform) although I haven't benchmarked it.

Issue #16, opened 2017-05-08
[text] fix <TextWrap />


hshoff:
Make it easier to have text autowrap in svg. https://github.com/hshoff/vx/blob/master/packages/vx-text/src/text/TextWrap.js

techniq:
I originally [created](https://github.com/recharts/recharts/pull/243) something like this for the recharts library (original [issue](https://github.com/recharts/recharts/issues/198)) but recently decided to extract it out into it's own repo/library as [react-svg-text](https://github.com/techniq/react-svg-text) mostly because I saw vx's `TextWrap` with it's TODO.

I haven't started using vx myself (yet), but thought you might want to use this.  It still needs some work before I release it on npm (see TODO section on README).  It also provides more features than just word wrapping (changing the `verticalAnchor` is very useful).  Check out the [demo](https://techniq.github.io/react-svg-text/) to get a feel for it.

hshoff:
@techniq very nice! great work

techniq:
@hshoff I release `react-svg-text` today - https://www.npmjs.com/package/react-svg-text.

I need to come up with a better testing strategy (see TODO section on README about jsdom / jest).  I also haven't done much through manually testing so hopefully I didn't miss anything to glaring.

Anyways, feel free to integrate this into `vx` as you see fit, and if you come across anything you need, feel free to submit a issue or PR.

techniq:
fyi - I fired up a quick codepen and noticed I had forgot to add the `main` entry point in `package.json`.  I fixed this and pushed out `0.1.1`

hshoff:
Awesome! thanks for this @techniq!

techniq:
Very welcome. Thank you for vx :)

hshoff:
fixed in: https://github.com/hshoff/vx/pull/208

PR #208, opened 2017-12-06
Initial port of react-svg-text to vx.  Issue #16

techniq:
#### :boom: Breaking Changes

- Removes `<TextWrap>`, `<TextOutline>` and `<TextBackground>` components, which were incomplete

#### :rocket: Enhancements

- Adds new `<Text>`, with the following features
  - Word-wrapping (when width prop is defined)
  - Vertical alignment (verticalAnchor prop)
  - Rotation (angle prop)
  - Scale-to-fit text (scaleToFit prop)

This is mostly just a direct port of my [react-svg-text](https://github.com/techniq/react-svg-text) component, which I plan to deprecate in favor of `@vx/text` (since it can be installed standalone as well).

--- 

A few points of discussion before merging/releasing:
- The tests rely on the ability to query the DOM for an element's size (using `getComputedTextLength`).  Since jest currently only supports jsdom, which lacks this feature (although [puppeteer](https://github.com/xfumihiro/jest-puppeteer) support might be coming soon), I've commented out all the tests so as to not break the build
- I wasn't sure of the best way to demo the capabilities of the component in a concise manner (like the other tiles) so for now I just ported my Demo page as it's currently the best way to understand what the component provides.  Feel free to replace and/or remove the demo, although I find it useful when developing
  - Btw, I noticed I'm receiving the following error for the `Fit` example
image
    - `Fit` works fine when enabling at the top, but the standalone example is throwing this error (I'm assuming from next.js / SSR.  This might be an actual bug, but I was having trouble identifying the cause, especially if it's only during SSR.
- Should we replace all/most/some of our `<text>` usage with `<Text>`, especially for things like Axis which pass `tickLabelPropsObj`?
- I'm not sure if I like `angle` or `rotate` as the name for the prop to rotate.  Thoughts?
- I was having trouble using a 2.x version of reduce-css-calc, even after they resolved this [issue](https://github.com/MoOx/reduce-css-calc/issues/33), although I'm getting a different error now.  This not a big issue, and we're using the same version as `react-svg-text`, but I was hoping to update it while making the transition.  At some point I'll try to submit another issue to work through the current issue, but it's not a make/break to get this into vx.

Sorry for the lengty PR message, I mostly wanted to get some notes down while they were fresh.  A lot of these items above we can work through after merging if you'd like as well, or I can make some changes before merging.


packages/vx-text/package.json
```diff
@@ -29,7 +29,9 @@
   },
   "homepage": "https://github.com/hshoff/vx#readme",
   "dependencies": {
-    "classnames": "^2.2.5"
+    "classnames": "^2.2.5",
+    "lodash": "^4.17.4",
+    "reduce-css-calc": "^1.3.0"
```
hshoff: Should update the author field to your name 😄

hshoff:
Wonderful! Thanks for this @techniq! This is a huge improvement and I'm excited to welcome your excellent `react-svg-text` work to the vx family! 

> I've commented out all the tests so as to not break the build

makes sense and sounds good for now.

> Btw, I noticed I'm receiving the following error for the Fit example

I'll take a look after merging and see what's up.

> Should we replace all/most/some of our <text> usage with <Text>, especially for things like Axis which pass tickLabelPropsObj?

Yeah probably where it makes sense/simplifies the API. We can do this in a later PR. 

> I'm not sure if I like angle or rotate as the name for the prop to rotate. Thoughts?

I'm fine with leaving it as `angle` right now and see if folks get caught up on it. 

> I was having trouble using a 2.x version of reduce-css-calc

No problem, something we can keep an eye on.

PR #213, opened 2017-12-11
Fix center and add pointRadius, innerRef

msathis:
#### :boom: Breaking Changes
None.
- 

#### :rocket: Enhancements
Added pointRadius and fixed center.
-

#### :memo: Documentation

-

#### :bug: Bug Fix
Fixed center typo.
- 

#### :house: Internal

-

### DIFF
--- a/packages/vx-geo/src/projections/Projection.js
+++ b/packages/vx-geo/src/projections/Projection.js
@@ -79,7 +79,7 @@ export default function Projection({
             <path
               className={cx(`vx-geo-${projection}`, className)}
               d={path(feature)}
-              ref={ref => innerRef && innerRef(ref, feature, i)}
+              ref={innerRef && innerRef(feature, i)}
               {...additionalProps(restProps, {
                 ...feature,
                 index: i,
--- a/packages/vx-geo/src/projections/Projection.js
+++ b/packages/vx-geo/src/projections/Projection.js
@@ -59,7 +59,6 @@ export default function Projection({
 
   if (pointRadius) path.pointRadius(pointRadius);
 
-
   return (
     <Group className={`vx-geo`}>
       {graticule &&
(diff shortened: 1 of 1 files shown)


hshoff:
LGTM thanks!

packages/vx-geo/src/projections/Projection.js
```diff
@@ -47,14 +49,16 @@ export default function Projection({
...
-  if (center) currProjection.translate(center);
+  if (center) currProjection.center(center);
   if (rotate) currProjection.rotate(rotate);
   if (precision) currProjection.rotate(precision);
   if (fitExtent) currProjection.fitExtent(...fitExtent);
   if (fitSize) currProjection.fitSize(...fitSize);
 
   const path = geoPath().projection(currProjection);
 
+  if (pointRadius) path.pointRadius(pointRadius);
```
kachkaev, line 60:
```suggestion
  if (Number.isFinite(pointRadius)) path.pointRadius(pointRadius);
```

👋 folks! Do you if `pointRaidus` can be `0`? Would it be a valid input? I can change this in https://github.com/airbnb/visx/pull/1767 if that makes sense.

williaster: hey @kachkaev  according to the `d3-geo` docs this should be a 2-element numerical array https://d3js.org/d3-geo/projection#projection_center

that's consistent with our `@visx/geo` types / docs https://airbnb.io/visx/docs/geo#Projection_center

so I don't think `0` is valid.

kachkaev: I think it’s related to https://d3js.org/d3-geo/path#path_pointRadius (one number)

williaster: 🤦 🤦 🤦  sorry no idea how I misread that one. 100% agree with you/thanks for adding in your new PR 🙏

PR #217, opened 2017-12-20
Added description and example of each component and enhancer

crcarlo:
#### :memo: Documentation
I separated the documentation with a section for every enhancer and component. 
Enhancers' description have been improved by describing which props will the component get.
Added description and usage example of `ParentSize` previously missing.


hshoff:
LGTM thanks!

Issue #222, opened 2018-01-09
Issue with scalePower


katerineknox:
Hi,
I was just trying to use the scalePower function and came across an error.
`TypeError: scale.base is not a function`
Looks like it's trying to set base when it should be using exponent.
Happy to raise a pull request to fix and update the api to use exponent instead - since it's a small change I'm assuming I shouldn't break anything.

williaster:
Hi @katerineknox! Thanks for checking out vx. 

I think you're right that this should be `scale.exponent` as opposed to `.base`. A PR would be fantastic if you're up for it! Agree re the minor api change 👍

PR #223, opened 2018-01-10
issue #222 update to scalePower api to take in exponent instead of ba…

katerineknox:
#### :bug: Bug Fix

Fix for Issue #222. Update to scalePower api to take in exponent instead of base and pass this through to d3 scalePow. Added in test and updated ReadMe.md


hshoff:
LGTM, thank you @katerineknox!

Issue #228, opened 2018-01-14
Debounce option for ParentSize


Zaitsev:
We used a high CPU-load math for render SVG  wrapped in Parentsize component, so it will be nice to debounce resize events, so SVG will be re-rendered after some timeout  after resize stopped

williaster:
thanks for checking out vx Zaitsev! 

seems useful and similar to [similar functionality that exists within `withScreenSize` and `withParentSize](https://github.com/hshoff/vx/search?utf8=%E2%9C%93&q=lodash%2Fdebounce&type=)` (which we are [planning to deprecate](https://github.com/hshoff/vx/issues/220) for `v1` btw). 

PR's welcome :) but otherwise can get to it relatively soon 👍

PR #241, opened 2018-02-07
Added debounceTime prop to ParentSize

AlexJuarez:
Here is a proposed solution for #228 

Just adds debounceTime prop with a default of 300ms.

Note, resize-observer-polyfill waits on requestAnimationFrame between invocations. https://github.com/que-etc/resize-observer-polyfill/blob/a3ae98bcd34e972b92d9e40e8b974a75399503e8/dist/ResizeObserver.js#L214


williaster:
this LGTM and I tested it locally 👍thanks @AlexJuarez! 🎉

PR #254, opened 2018-03-16
remove dependeny on resize-observer-polyfill in ParentSize

sto3psl:
## Motivation

To keep bundles slim and give the user of this library a little bit more control I suggest removing the dependency of `resize-observer-polyfill` from `ParentSize`.  
There is a lot of discussion if library authors should ship with polyfills included or not, that's why this is just a suggestion that would make my team at work switch even faster to vx 😄. I think it's fine to tell the user of the library to include a polyfill if needed to save bytes. I'm happy to discuss this!

#### :boom: Breaking Changes

- `ParentSize` doesn't use `resize-observer-polyfill` anymore.

#### :memo: Documentation

- Added a note in the `ParentSize` readme.

Thank you for this library btw. I'm currently convincing my team to switch to vx for our visualizations and it is great to use, whether or not this change happens.

### DIFF
--- a/packages/vx-demo/components/tiles/responsive.js
+++ b/packages/vx-demo/components/tiles/responsive.js
@@ -1,6 +1,11 @@
 import React from 'react';
 import { ParentSize } from '@vx/responsive';
 import Lines from './lines';
+import ResizeObserver from 'resize-observer-polyfill';
+
+if (!window.ResizeObserver) {
+  window.ResizeObserver = ResizeObserver;
+}
 
 function Nav() {
   return (
--- a/packages/vx-responsive/Readme.md
+++ b/packages/vx-responsive/Readme.md
@@ -54,6 +54,8 @@ let chartToRender = withParentSize(MySuperCoolVxChart);
 
 You might do the same thing using the `ParentSize` component.
 
+*Note: `ParentSize` uses [ResizeObserver](https://developers.google.com/web/updates/2016/10/resizeobserver) which is currently only supported in Chrome 64 and newer ([caniuse](https://caniuse.com/#feat=resizeobserver)). If you need support for other browsers you need a polyfill like `resize-observer-polyfill` which is only a `npm install -S resize-observer-polyfill` away.*
+
 ### Example:
 ``` js
 import { ParentSize } from "@vx/responsive";
--- a/packages/vx-responsive/src/components/ParentSize.js
+++ b/packages/vx-responsive/src/components/ParentSize.js
@@ -1,7 +1,6 @@
 import debounce from 'lodash/debounce';
 import PropTypes from 'prop-types';
 import React from 'react';
-import ResizeObserver from 'resize-observer-polyfill';
 
 export default class ParentSize extends React.Component {
   constructor(props) {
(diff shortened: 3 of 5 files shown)


hshoff:
Thanks for the PR @sto3psl! Definitely agree this is the happy path in the future. Will land this in an upcoming release.

packages/vx-demo/components/tiles/responsive.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import { ParentSize } from '@vx/responsive';
 import Lines from './lines';
+import ResizeObserver from 'resize-observer-polyfill';
+
+if (!global.ResizeObserver) {
+  global.ResizeObserver = ResizeObserver;
```
hshoff: should this be `window && !window.ResizeObserver` instead of `global`?

packages/vx-responsive/Readme.md
```diff
@@ -54,6 +54,8 @@ let chartToRender = withParentSize(MySuperCoolVxChart);
 
 You might do the same thing using the `ParentSize` component.
 
+*Note: `ParentSize` uses [ResizeObserver](https://developers.google.com/web/updates/2016/10/resizeobserver) which is currently only supported in Chrome 64 and newer ([caniuse](https://caniuse.com/#feat=resizeobserver)). If you need support for other browsers you need a polyfill like `resize-observer-polyfill` which is only a `npm install -S resize-observer-polyfill` away.*
+
```
hshoff: 👍

packages/vx-responsive/src/components/ParentSize.js
```diff
@@ -1,63 +1,62 @@
...
+
+export default class ParentSize extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { width: 0, height: 0, top: 0, left: 0 };
+    this.resize = debounce(this.resize.bind(this), props.debounceTime);
+    this.setTarget = this.setTarget.bind(this);
+  }
+  componentDidMount() {
+    this.ro = new ResizeObserver((entries, observer) => {
```
hshoff: We should probably add an existence check here and thow an error saying to include the polyfill to fail early and so folks know why it's not working.

hshoff:
Hi @sto3psl sorry for the delay on this! I wanted to chat with some folks about approaches for handling polyfills in libraries. 

The nice part of `resize-observer-polyfill` is that it's not destructive. So if the ResizeObserver is already present it won't mutate anything: https://github.com/que-etc/resize-observer-polyfill/blob/master/src/index.js#L5

So then it comes a question of bundle sizes and library responsibility. 

It's true that out of the box you'll be including more js then needed if the polyfill is unnecessary. Though, this can be solved at the bundle tool level. You should be able to alias out the polyfill so it's not included in the bundle. 

My current thinking is to continue including the polyfill for ease of use, not having to respond to issues about xyzResizeObserver polyfills, and leaning to letting app developers decided how to slim down the bundle size. 

It's a tradeoff i know, but I see value in ease of use and correctness by default over extra js. Would love to hear your thoughts.

loopmode:
I think the best approach is to do include polyfills, but provide info and opt-out instructions.
So, including the polyfill has many benefits - allows quick start, helps new and unexperienced users.
Then provide a section in the readme/docs about how to exclude it from bundling, with examples for the popular bundlers. I guess just webpack is a good start. 

Most of the users will be happy about it and not even know or care. 
The fewer users who do care, e.g about bundle sizes, they have a more special requirement plus the experience. They do check docs for such details, and they can configure e.g. webpack.
So, it would be somewhat pareto :)

sto3psl:
> I think the best approach is to do include polyfills, but provide info and opt-out instructions.

I'd rather have instructions on how to include the polyfill in case I really need it. Removing code from a library should not be the concern of the consumer who, in most cases, doesn't have control of the libraries code. The idea of changing my bundlers config to exclude code from a third-party library doesn't speak well to me.

> So, including the polyfill has many benefits - allows quick start, helps new and unexperienced users.

I fully agree on the easier quick start for newcomers and maybe we can also teach them about the use of polyfills, by providing nice errors for them, if something goes wrong. Giving easy understandable warnings and information on how to solve the problem is, in my opinion, more helpful than not confronting them with the problem at all.

In this specific case with `ResizeObserver` I do agree on keeping the polyfill because browser support is so sparse.

loopmode:
Thinking about it, I actually agree.
I guess a user who is using react to draw svg with vx on top of d3 is most likely deep enough in the matter to be capable of understanding the need for and the procedure of including a polyfill.

sto3psl:
Going to close this now and maybe come back if 1 more browser supports `ResizeObserver` 😬

Issue #267, opened 2018-04-09
@vx/axis ticklabelprops dy not working


maganuk:
After update to 0.0.159 or 0.0.160 from 0.0.158, setting the dy in ticklabelprops on the left or right axis doesnt seem to work. 

`tickLabelProps={() => ({
        dy: '-0.4em',
        textAnchor: 'start'
      })}
`

hshoff:
Thanks for the bug report @maganuk! We replaced axis tick labels with `@vx/text` in `v0.0.159`. Looks like we need to update`@vx/text` to support `em/rem` values for `dx/dy`

sto3psl:
I would be happy to come up with a PR for this. The calculation for `em` to `px` should be pretty straightforward with the font size of the `<Text />` element. But for `rem` I'm not quite sure. Would `getComputedStyle(document.documentElement).fontSize` be the correct way? It returns the font size in `px` which should equal `1rem`.

techniq:
@sto3psl based on a [quick search](https://stackoverflow.com/a/42769683/191902) that should work.  Might be better to do this in userland and pass the pixel value to `<Text dy={...} />`?

sto3psl:
@techniq I agree. I think supporting `em` is fine and if one needs `rem` it's not hard to do in userland. There can even be a hint in the docs how to work with `rem`.

hshoff:
This plan sounds good to me

hshoff:
fixed by https://github.com/hshoff/vx/pull/271

PR #271, opened 2018-04-16
fix: tickLabelProps dy and dx work again

sto3psl:
This should fix the bug mentioned in https://github.com/hshoff/vx/issues/267.
I had several problems with this which made me come to the solution of using transforms for this bug fix.  

1. `px` values for `dx` and `dy` were fine and worked.
2. `em` values were fine as long as `fontSize` was explicitly provided for `<Text />`. Without `fontSize` there is no value to use for the transformation from `em -> px` without accessing the DOM.
3. `rem` units had the same problem that there is no way of transforming to `px` without the DOM.
4. I couldn't do `<text dy="1em">...</text>` because the child `<tspan>` defines a `dy` value that overwrites the one from `<text>`. I'm not quite sure why. [example](https://codepen.io/anon/pen/QmexEp)

In the end setting a CSS transform seemed like the easiest solution and has the nice benefit of supporting `rem`.

If there is a better solution please let me know, this took longer than I want to admit 😅.

#### :bug: Bug Fix

- `@vx/text` support `dx` and `dy` again

### DIFF
--- a/packages/vx-text/src/Text.js
+++ b/packages/vx-text/src/Text.js
@@ -138,22 +138,20 @@ class Text extends Component {
     }
 
     return (
-      <text
-        {...textProps}
-        textAnchor={textAnchor}
-        style={{
-          transform: `translate(${dx}, ${dy})`,
-          ...textProps.style
-        }}
-      >
-        {
-        wordsByLines.map((line, index) => (
-          <tspan x={x} dy={index === 0 ? startDy : lineHeight} key={index}>
-            {line.words.join(' ')}
-          </tspan>
-        ))
-      }
-      </text>
+      <svg x={dx} y={dy} fontSize={textProps.fontSize} style={{ overflow: 'visible' }}>
+        <text
+          {...textProps}
+          textAnchor={textAnchor}
+        >
+          {
+          wordsByLines.map((line, index) => (
+            <tspan x={x} dy={index === 0 ? startDy : lineHeight} key={index}>
+              {line.words.join(' ')}
+            </tspan>
+          ))
+        }
+        </text>
+      </svg>
     );
   }
 }
--- a/packages/vx-text/src/Text.js
+++ b/packages/vx-text/src/Text.js
@@ -103,8 +103,8 @@ class Text extends Component {
     } = this.props;
     const { wordsByLines } = this.state;
 
-    const x = textProps.x + dx;
-    const y = textProps.y + dy;
+    const x = textProps.x;
+    const y = textProps.y;
 
     let startDy;
     switch (verticalAnchor) {
@@ -139,10 +139,12 @@ class Text extends Component {
 
     return (
       <text
-        x={x}
-        y={y}
-        textAnchor={textAnchor}
         {...textProps}
+        textAnchor={textAnchor}
+        style={{
+          transform: `translate(${dx}, ${dy})`,
+          ...textProps.style
+        }}
       >
         {
         wordsByLines.map((line, index) => (


hshoff:
Clever solution! My only thought is how this might play with rotating the text. Will give this branch a go locally when I get a chance and report back

sto3psl:
screen shot 2018-04-18 at 12 04 27

In the screenshot I set `angle: -45`.

I checked myself and it did not work with rotation. My new commit fixes it. Let me know if there are any problems.

techniq:
Does this work on all (our supported) browsers?  I remember [reading](https://css-tricks.com/transforms-on-svg-elements/) that css transforms did not work on IE.  It would be nice to support IE 11 if possible, although I know it's on it's way out (github is dropping it in a few months).

sto3psl:
From [caniuse](https://caniuse.com/#search=transform) about IE11.
> Does not support CSS transforms on SVG elements (transform attribute can be used instead)

I can change it to use the transform attribute which means that `em` and `rem` don't work that nice anymore and would need to be calculated.

sto3psl:
The problem of transforming `em -> px` is only there if no `font-size` is set on `<Text />` (or stuff needs to go into the DOM which is not so nice). 

Would it be a good tradeoff to set a default `fontSize` prop of maybe `11px` that there is always a value to calculate `dx` and `dy` with?

That would get rid of the transforms, uses the correct svg attributes and works in supported browsers.

techniq:
@sto3psl Since adding `dx/dy` on `<text>` conflicts with the `<tspan>`, maybe we could wrap the entire thing in a `<g>` and apply `dx/dy` to `x/y`?

techniq:
Another thought is could you use [reduce-css-calc](https://github.com/MoOx/reduce-css-calc) like we are for `verticalAnchor` and convert it that way 

change:
```js
const x = textProps.x + dx;
const y = textProps.y + dy;
```
to
```js
const x = reduceCSSCalc(`calc(${textProps.x}) + ${dx}`);
const y = reduceCSSCalc(`calc(${textProps.y}) + ${dy}`);
```

sto3psl:
Wrapping `<text />` in a `<g />` could work. 

The problem with `reduceCSSCalc` is if `dx` or `dy` are em values and `textProps.x` or `textProps.y` are px values.

```
reduceCSSCalc(`calc(5px + 1em)`) -> 'calc(5px + 1em)'
```

And `'calc(5px + 1em)'` is not valid as `x` or `y` value.

techniq:
@sto3psl oh, I thought it handled mixed units, but I see it does not.  Hopefully the `<g>` route works.  Thanks for digging into the problem.

sto3psl:
Ok I played with svg a little and `<g>` does not work since it ignores `x` and `y`. We could wrap the `<text>` with a `<svg>` though and then we get `px,em,rem` support. According to the spec it's no problem if `svg` elements have `svg`s as child. See this codepen -> https://codepen.io/anon/pen/rvNGjz

techniq:
Our [Group](https://github.com/hshoff/vx/blob/master/packages/vx-group/src/Group.js) uses ```<g transform={`translate(${x}), ${y})`} />```.  Maybe try using the svg transform method on `g`?  You could also use the `<Group>` component to simplify it?  I'm assuming svg transforms understand em/rem (like css transform).

If this doesn't work, the `svg` might be a decent approach (would be odd to have a lot of nested SVG components, not sure if this could have any corner case problems).

sto3psl:
Svg transforms are unitless which means transforming `em -> px` again. The `<svg>` approach is really nice and Victory does a lot `<svg>` nesting too.

techniq:
@sto3psl sounds like `<svg>` is the way to go then 😄. Thanks again

sto3psl:
The `<svg>` seems to work nicely.

* The `style={{ overflow: "visible" }}` is necessary since the svg has no width and height and the text would otherwise not be visible. 
* The `fontSize` has to be set so that the browser calculates `em` values correctly.

Thanks for all the input and thoughts on this @techniq. If there is anything else let me know, this was quite fun to think about 😄.

techniq:
@sto3psl Thanks for digging in and finding a solution.  I haven't had a chance to pull it down locally (@hshoff if you could take a quick peak that would be awesome) but I have no reason to expect any issues :smile:

hshoff:
LGTM thanks for the contribution @sto3psl!

PR #272, opened 2018-04-17
[tooltip] expose optional containerProps in withTooltip, add tooltip readme

williaster:
These all apply to the `@vx/tooltip` package.

#### :rocket: Enhancements
- adds an optional `containerProps` as a second HOC "config" argument of `withTooltip(BaseComponent [, containerProps])`. This exposes a hook to enable users to customize any props on the container element.
- sets `width` and `height` to `inherit` by default on the container. I'm kind of on the edge with this and am open to removing it because it may be an edge case, but my reasoning is as follows: if a user is combining `withTooltip` and a responsive component like `ParentSize` like this example, they'll have to update the `withTooltip` container `div` to also have full width/height so this would be a "smart default".

#### `index.html`
```css
body, #root { width: 100%; height: 100%; }
```

#### `index.js`
```js
class ChartWithTooltip extends React.Component {
  // ...stuff

  render() {
    const { ... } = this.props;

    return (
      <ParentSize>
        {({ width, height }) => (
          <React.Fragment>
              // ...chart
            </svg>

            {tooltipOpen && (
              <Tooltip ...>...</Tooltip>
            )}
          </React.Fragment>
        )}
      </ParentSize>
    );
  }
}

// this would work out of the box if width/height are set to inherit, 
// otherwise the user would have to override the withTooltip container styles like so
// withTooltip(Chart, { style: { position: 'relative', width: 'inherit', height: 'inherit' } });
const ChartWithTooltip = withTooltip(Chart);

render(<ChartWithTooltip />, document.getElementById('root'));

```

#### :memo: Documentation
- Adds a readme for `@vx/tooltip` components + enhancer

### DIFF
--- a/packages/vx-tooltip/Readme.md
+++ b/packages/vx-tooltip/Readme.md
@@ -3,3 +3,105 @@
 ```
 npm install --save @vx/tooltip
 ```
+
+The `@vx/tooltip` package provides utilities for making it easy to add `Tooltip`s to a visualization and includes higher-order component (HOC) enhancers and Tooltip components.
+
+### Example:
+``` js
+import { withTooltip, TooltipWithBounds } from '@vx/tooltip';
+import { localPoint } from '@vx/event';
+
+class Chart extends React.Component {
+  handleMouseOver = (event, datum) => {
+    const coords = localPoint(event.target.ownerSVGElement, event);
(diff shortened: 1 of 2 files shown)


hshoff:
LGTM thanks for this contribution!

PR #298, opened 2018-06-04
[new build] use rollup, add es + umd builds. fixes #143

### LINKED ISSUE #143: Add module support for tree shaking

hshoff:
fixes: https://github.com/hshoff/vx/issues/143

- [x] adds `dist/vx-{name}.{umd,es}.js` files
- [x] adds `module` field to `package.json` that points to `dist/vx-{name}.es.js` for bundlers that support it (webpack, rollup, etc)
- [x] `main` field points to `dist/vx-{name}.umd.js`
- [x] removes [`react-fatigue-dev`](https://github.com/tj/react-fatigue-dev) + Makefile build (means you can build on windows now)
- [x] build targets  `"android": 30, "chrome": 35, "edge": 14, "explorer": 9, "firefox": 52, "safari": 8, "ucandroid": 1` (matching [airbnb-babel-preset](https://github.com/airbnb/babel-preset-airbnb/blob/master/index.js#L9-L17)
- [x] keeps `build/` with `cjs` babel build files for [deep paths](https://github.com/hshoff/vx/issues/143#issuecomment-367649444) (no breaking changes 🤞)


package.json
```diff
@@ -8,8 +8,7 @@
     "docs": "node ./scripts/docs/index.js",
     "prepare-release": "git checkout master && git pull --rebase origin master && npm run docs && lerna updated",
     "release": "npm run prepare-release && lerna publish --exact",
-    "format": "prettier \"{packages,scripts}/**/*.js\" --write",
-    "precommit": "lint-staged"
```
hshoff: TODO: add this back, it was blowing up on my Win10 box

williaster:
this is awesome! 🙌

package.json
sto3psl: This looks amazing!

Don't know if it applies but you can ignore the commit hooks if you do `git commit -m "message" -n`.

hshoff: @sto3psl awesome! didn't know I could do that... for context it was only blowing up when i did a commit on windows

hshoff:
going to merge this and release it as `0.0.165-beta.*`

PR #311, opened 2018-06-15
Support Custom Rendering For Pie Charts

psachs21:
I'd like to be able to animate a pie chart but for most animation frameworks, we need access to the children groups in order to modify the path parameters. By adding an optional render function, we can add animation to the pie chart while allowing users to choose their own animation framework.

#### :boom: Breaking Changes

None

#### :rocket: Enhancements

- Add optional render function as child of Pie to allow more fine grained control of rendering.


hshoff:
published in `v0.0.168` https://github.com/hshoff/vx/releases/tag/v0.0.168

PR #319, opened 2018-07-10
[area-closed] add optional y0 for custom area fills

rjatkinson2:
#### :rocket: Enhancements

- Adds optional y0 for custom area fills
- Primary Example: show negative area as integral
- I know typical pattern is for y0 to be an accessor passed to yScale, which isn't flexible enough in this case. Interested to hear thoughts and happy to refactor to fit project patterns.

```
const stock = appleStock.slice(800).map((val, index) => ( index < 250
  ? val
  : { ...val, close: -val.close })
);

....

<AreaClosed
   data={stock}
   xScale={xScale}
   yScale={yScale}
   x={xStock}
   y={yStock}
   y0={Math.min(yScale(0), yScale.range()[0])}
   strokeWidth={1}
   stroke={'url(#gradient)'}
   fill={'url(#gradient)'}
/>
```

negativeasintegral

### DIFF
--- a/packages/vx-shape/src/shapes/AreaClosed.js
+++ b/packages/vx-shape/src/shapes/AreaClosed.js
@@ -11,6 +11,7 @@ AreaClosed.propTypes = {
 export default function AreaClosed({
   x,
   y,
+  y0,
   xScale,
   yScale,
   data,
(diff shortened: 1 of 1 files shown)


techniq:
@rjatkinson2 Sounds similar to the existing [Threshold](https://vx-demo.now.sh/threshold), which allows you to also styling above/below differently.

rjatkinson2:
@techniq yea agreed and I originally went down the Threshold route. Threshold uses three Area charts, which seems superfluous for a single series. Additionally, I had trouble implementing Threshold without multiple data sets.

techniq:
@rjatkinson2 makes sense, and change is minor.  I'll let @hshoff review, but LGTM.

PR #327, opened 2018-07-25
<Pie /> Allow null sort callbacks

Gosha:
[`pie.sortValues(null)`](https://github.com/d3/d3-shape#pie_sortValues) turns off the default sorting, but before this PR `pie.sort` was never called with a `null` value. Hopefully this isn't a breaking change.

#### :bug: Bug Fix

- `<Pie pieSort={null} pieSortValues={null} />` isn't ignored


hshoff:
Sorry for the delay! just published https://github.com/hshoff/vx/releases/tag/v0.0.171

Issue #302, opened 2018-06-07
Heatmap - Allow for auto indexed rows


JacquiManzi:
First off, thanks for making this project- it’s exactly what I've needed.

So, I'm implementing a heatmap using VX and would like to provide a custom data structure. From reading the docs and source code, I believe this is the structure that is intended for the heatmap chart:

```
[{
  bin: Date, // The date we are applying for our x value (the column)
  bins: [{
    count: number, // This determines the color (heat) value
    bin: number // The row number for this bin in a column (this is unfortunately required)
 }]
```

The problem is that I would like to have my API return matching data structures for my charts to consume in order to prevent expensive data mapping on the client. My ideal structure looks like this:

```
[{
  bucket: Date, // The date we are applying for our X value (the column)
  values: [{
    title: string, // The descriptive bin value (not required by VX)
    count: number, // This determines the color (heat) value
  }]
}]
```

Luckily, VX allows me to override these props:
```
bin = d => d.bin,
bins = d => d.bins,
count = d => d.count
```

Which allows me to replace `bin` with `bucket` and `bins` with `values`, but I run into issues here since the yScale value is re-using the bin accessor function here:
`y={yScale(bin(b) + step) + gap}`

I think it’s valuable to allow for a custom row value (bin in this case) because it means you can have a custom order to items, but I would rather just use the index value of my values array. This way I wouldn’t need my API to provide a row value at all.

Could there potentially be a boolean prop that allows for an auto index like:
`const y = indexedRows ? j : bin(b);`

Which then can be used here:
`y={yScale(y + step) + gap}`

Ultimately this would allow me to do this:
```
<HeatmapRect
 bins={d => d.values }
 bin={d => d.bucket }
 indexedRows={true}
 data={data}
 x={0}
 xScale={xScale}
 yScale={yScale}
 colorScale={colorScale}
 opacityScale={opacityScale}
 binWidth={binWidth}
 binHeight={binHeight}
 step={dStep}
 gap={0}
/>
```
A quick example of what I'm considering is here: https://github.com/muxinc/vx/pull/1/files

hshoff:
Great idea @JacquiManzi! I went for a bit different approach that allows you to pass in a `yBin()` prop (it defaults to your `bin()` prop).

Both `yBin()` and `bin()` will get a second arg with the index of the datum.

instead of adding a boolean prop you would do:

```js
<HeatmapRect
 bins={d => d.values }
 bin={d => d.bucket }
[...]

PR #332, opened 2018-08-09
Cleaned up and simplified heatmaps

jens-ox:
The heatmap demo overflows on the top (see the white half circle):
white circle

This is due to the way the y-positions of the heatmap elements are calculated. This can be simplified by using the array lengths to calculate the positions (imho the cleaner way).

This is what the updated Heatmap demo looks like:

heatmap updated

----

#### :boom: Breaking Changes

- Different `vx-heatmap` API

#### :rocket: Enhancements

- Cleaner `vx-heatmap` API, better Heatmap layout
- Adds a separation attribute (margin between the two heatmaps) to the heatmap demo
- Removes usage of `data[0].bins.length` as reference bins length, calculates maximum

#### :bug: Bug Fix

- Fixes broken Heatmap demoo
- Removes hardcoded left margin of 5

### DIFF
--- a/packages/vx-demo/components/tiles/heatmap.js
+++ b/packages/vx-demo/components/tiles/heatmap.js
@@ -21,29 +21,29 @@ export default ({
     left: 20,
     right: 20,
     bottom: 110
-  }
+  },
+  separation = 20
 }) => {
   if (width < 10) return null;
 
   // bounds
-  const size = width > margin.left + margin.right ? width - margin.left - margin.right : width;
+  const size =
+    width > margin.left + margin.right ? width - margin.left - margin.right - separation : width;
   const xMax = size / 2;
-  const yMax = height - margin.bottom;
-  const dMin = min(data, d => min(y(d), x));
-  const dMax = max(data, d => max(y(d), x));
-  const dStep = dMax / data[0].bins.length;
+  const yMax = height - margin.bottom - margin.top;
+  const maxBucketSize = max(data, d => y(d).length);
   const bWidth = xMax / data.length;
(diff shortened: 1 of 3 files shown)


hshoff:
Hi @jens-ox thanks for the pull request! 

I agree this would simplify the API only this would come at the cost of control. The heatmap API used to be simpler but was changed to address https://github.com/hshoff/vx/issues/302 in https://github.com/hshoff/vx/pull/307. Explanation: https://github.com/hshoff/vx/issues/302#issuecomment-396785295.

90% of the time users will likely scale on the index of the bin. The other 10% may need to do something else to support their data. Supporting that last 10% is a tradeoff. So I think for now we'll keep the API as is for now. Happy to continue the discussion.

I'd love to land your heatmap demo tile fix. I agree it looks much better with your improvements!

jens-ox:
Hi @hshoff, thanks for your feedback! I don't see how this PR would pose a problem to scenarios like #302, as no `bin` function is used anymore. `HeatmapRect` would look like this:

```jsx
<HeatmapRect
  data={data}
  xScale={xScale}
  yScale={yScale}
  colorScale={colorScale2}
  opacityScale={opacityScale}
  bins={d => d.values}
  binWidth={bWidth}
  binHeight={bWidth}
  gap={2}
  onClick={data => event => {
    if (!events) return;
    alert(`clicked: ${JSON.stringify(data.bin)}`);
  }}
/>
```

The only difference to the example is the additional `bins` attribute.

https://gist.github.com/jens-ox/83e85812a959b969a8c9feace61b64c8 is the code of the Heatmap demo using the same data structure as @JacquiManzi.

PR #335, opened 2018-08-15
Avoid ResizeObserver loop limit exceeded

kristw:
#### :bug: Bug Fix

- Avoid `ResizeObserver` loop limit exceeded

> since `ResizeObserver` seems to be firing the callback immediately after `observe()` is called, this causes an infinite loop.

Reference: https://github.com/WICG/ResizeObserver/issues/38

Solution was to wrap the resize call in `requestAnimationFrame` to avoid infinite loop

Reference:
- https://github.com/cerner/terra-core/pull/1647/files
- https://github.com/cerner/terra-core/pull/1658/files

### DIFF
--- a/packages/vx-responsive/src/components/ParentSize.js
+++ b/packages/vx-responsive/src/components/ParentSize.js
@@ -9,22 +9,26 @@ export default class ParentSize extends React.Component {
     this.state = { width: 0, height: 0, top: 0, left: 0 };
     this.resize = debounce(this.resize.bind(this), props.debounceTime);
     this.setTarget = this.setTarget.bind(this);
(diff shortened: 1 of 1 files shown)


williaster:
nice  🙌 LGTM 🔁  

were we seeing this on dashboards? is it specific to browser versions? (I haven't seen it before, or haven't been aware of seeing it :)

kristw:
@williaster @michellethomas found it while working on integration test in Cypress.

hshoff:
published in `v0.0.172`: https://github.com/hshoff/vx/releases/tag/v0.0.172

PR #346, opened 2018-09-05
[development] add linting

### LINKED ISSUE #128: add linting
Well it sound funny that a project from `Airbnb JavaScript Style Guide creator` doesn't have linting :)

In italy we have a saying
>the shoemaker always wears worn-out shoes

 a reminder that those who carry out a particular profession take care of their clients needs but neglect their own

@hshoff can i proceed adding eslint using airbnb rules?

geekplux:
#### ✨ Development

- add linting
- fix #128 

If you run `npm run lint`, you would discover there are:

```bash
✖ 10557 problems (10488 errors, 69 warnings)
  6731 errors and 0 warnings potentially fixable with the `--fix` option.
```

lol 😂

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 build
+dist
 bundle.js
 test/tmp
 *.log
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,13 @@
+.DS_Store
+node_modules
+build
+bundle.js
+test/tmp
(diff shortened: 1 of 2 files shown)


.eslintignore
```diff
@@ -0,0 +1,13 @@
+.DS_Store
+node_modules
+build
```
hshoff: need to add `dist` here as well

geekplux: done

PR #349, opened 2018-09-06
Fix label transform computation

lucafalasco:
#### :bug: Bug Fix

- Fix `labelTransform` function in *@vx/axis* to correctly align the axis label when different values for the scale's *range* are specified, in particular when it does not start (or end) with a 0.

To give a visual example, take a scale defined as below(same as [this](https://github.com/hshoff/vx/blob/master/packages/vx-demo/components/tiles/axis.js#L43#L47), but with a slightly different range): 

```
const yScale = scaleLinear({
    range: [yMax, yMax / 2],
    domain: [0, max(data, y)],
    nice: true
  });
```
this is what we get with the current code(which basically uses a `Math.max(...range) / 2` to compute the center):  
screen shot 2018-09-06 at 21 46 29

while this is the same example with the fix: 
screen shot 2018-09-06 at 21 51 14

### DIFF
--- a/packages/vx-axis/src/utils/labelTransform.js
+++ b/packages/vx-axis/src/utils/labelTransform.js
@@ -14,7 +14,7 @@ export default function labelTransform({
     y,
     transform = null;
   if (orientation === ORIENT.top || orientation === ORIENT.bottom) {
-    x = Math.max(...range) / 2;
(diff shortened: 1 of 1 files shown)


williaster:
this LGTM, thanks for the fix @lucafalasco!! 🙌

PR #350, opened 2018-09-07
add back genBoxPlot.js to mock-data

ezy:
#### :bug: Bug Fix

genBoxPlot was previously added in pull request [#89]: 

- [mock data] add `genBoxPlot()` [#89](https://github.com/hshoff/vx/pull/89)

It's still referenced and required in the boxchart demo `packages/vx-demo/pages/boxplot.js`

Adding it back so it can be imported and avoid the package not-found error on `yarn start` when trying to use the import:

`import { genBoxPlot } from '@vx/mock-data';`

### DIFF
--- a/packages/vx-mock-data/src/generators/genBoxPlot.js
+++ b/packages/vx-mock-data/src/generators/genBoxPlot.js
@@ -1,0 +1,21 @@
+export default function genBoxPlot(number) {
+  const data = [];
+  let i;
+  for (i = 0; i < number; i += 1) {
+    const points = [];
+    let j;
+    for (j = 0; j < 5; j += 1) {
+      points.push(Math.random() * 100);
+    }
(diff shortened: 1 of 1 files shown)


hshoff:
This looks like a case of outdated gallery example code on my part.

The source for the boxplot gallery example is here: https://github.com/hshoff/vx/blob/master/packages/vx-demo/components/tiles/boxplot.js#L6 and it imports `genStats()`: https://github.com/hshoff/vx/blob/master/packages/vx-mock-data/src/generators/genStats.js#L7.

So instead I think all we need to do is update the example code string in `vx-demo/pages/boxplot.js` to match the code in `vx-demo/components/tiles/boxplot.js`

ezy:
Cool - I'll carve out some time to update the boxplot chart based on your comments and create a new pull request.

PR #363, opened 2018-09-28
Add rest props to internal div

giulioz:
#### :boom: Breaking Changes

- Nothing

#### :rocket: Enhancements

- Needed if used in flex layouts where 100% size breaks the layout.

### DIFF
--- a/packages/vx-responsive/src/components/ParentSize.js
+++ b/packages/vx-responsive/src/components/ParentSize.js
@@ -43,9 +43,9 @@ export default class ParentSize extends React.Component {
     this.target = ref;
   }
   render() {
-    const { className, children } = this.props;
+    const { className, children, ...rest } = this.props;
     return (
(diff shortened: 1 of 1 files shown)


packages/vx-responsive/src/components/ParentSize.js
```diff
@@ -43,9 +43,9 @@ export default class ParentSize extends React.Component {
     this.target = ref;
   }
   render() {
-    const { className, children } = this.props;
+    const { className, children, ...rest } = this.props;
```
hshoff: Super minor nit, but to match the rest of the codebase could we rename this `...restProps`?

giulioz: Done, sorry, we were in a rush at work and we missed this.
Thank you

hshoff:
LGTM, thanks for the contribution!

hshoff:
This is now published in `@vx/responsive@0.0.177`

PR #367, opened 2018-10-04
add scale type

lucafalasco:
#### :rocket: Enhancements

- Add `type` property as utility to scales to do specific actions or use specific interfaces from "agnostic" contexts.


williaster:
nice add! 🙌

PR #368, opened 2018-10-04
improve BarStack to work with all scales type

### LINKED ISSUE #288: BarStack doesn't work with ScalePoint
`BarStack` doesn't work with ScalePoint because the latter doesn't implement `innerPadding()` (or `outerPadding()` for that matter). This might or might not be a D3 issue,  and might or might not be by design, still it wasn't what I expected.

lucafalasco:
#### :rocket: Enhancements

- BarStack and BarStackHorizontal were currently broken if a scale other than `ScaleBand` was passed as scale prop; this enhance those components by checking for existence of specific interface functions and avoiding throwing errors with different scales.

#### :bug: Bug Fix

- fix #288


lucafalasco:
Yeah that's exactly the approach I initially considered.
The reasons I opted for the `type` property check here were essentially:

1. I thought it could be useful/clearer (together with #367 ) to implement `isScaleBand()` etc. util functions also in other scenarios.

2. If the d3 scales interfaces change, we just have to update the function that is executed and not the "check-for-existence" function, as the `type` remains the same.

Looking at it again though, I believe that, as you suggest, checking for feature support is probably better and more future proof if we add more scales implementing the same functions. Also, the smaller bundle is definitely a good point 😉

PR #370, opened 2018-10-05
Add optional flex prop to LegendLabel

jdmoody:
#### :rocket: Enhancements

- Add optional `flex` prop for overriding `flex` css property to LegendLabel

#### :bug: Bug Fix

- Allow overriding flex property to fix [IE flexbox bug](https://stackoverflow.com/questions/39192995/flex-container-wont-expand-to-fit-content-in-ie). IE 11 does not seem to expand to minimum content width when flex-basis is 0px (this is the case when using the shorthand `flex: 1`).

I encountered a bug with the LegendLabel layout on IE11, and setting the flex property to `flex: auto` fixed it. I was hesitant to change the style for all usages, so I added an optional `flex` prop to `Legend` and `LegendLabel`.


jdmoody:
Np! Thanks for making such a great visualization library 🙏

PR #382, opened 2018-11-15
Let y0 use the same way as y

callmedadaxin:
change

```
<AreaClosed y={d => d.a} y0={d => yScale(d.b)}/>
```

to

```
<AreaClosed y={d => d.a} y0={d => d.b}/>
```

#### :boom: Breaking Changes

- 

#### :rocket: Enhancements

-

#### :memo: Documentation

-

#### :bug: Bug Fix

- 

#### :house: Internal

-

### DIFF
--- a/packages/vx-shape/src/shapes/AreaClosed.js
+++ b/packages/vx-shape/src/shapes/AreaClosed.js
@@ -27,7 +27,7 @@ export default function AreaClosed({
 }) {
   const path = area()
     .x((...args) => xScale(x(...args)))
-    .y0(y0 || yScale.range()[0])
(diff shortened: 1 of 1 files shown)


hshoff:
HI @callmedadaxin thanks for checking out vx! Good catch with the inconsistency of some props. I've gone a done a big pass across all the components and packages with an eye for making all the APIs consistent. The PR is ready just yet but you can take a look here: https://github.com/hshoff/vx/compare/harry-breaking#diff-909d1f4a3bf190080a79346c2e8f0f04

Instead of a applying the scales to the accessors for you, you would use the scale in the accessor. This gives the user more control over how a thing is rendered vs making assumptions about how they most likely want it rendered.

Hope to have a PR with the changes soon!

Issue #328, opened 2018-07-31
Can you add event handlers to a VX linegraph?


kevinxia787:

hshoff:
yup! 

```js
onClick={data => event => console.log('click')}
```

kevinxia787:
@hshoff Thanks for the quick reply! Is there a way to isolate the point that I click on? Because if I do 

`onClick={data => event => console.log(data)}
`
I get the data array used to plot the line, not the specific entry.

hshoff:
Ah I see, it takes a bit more work to make that work on a LinePath. Check out this tutorial from @browniefed and the tooltip section for mapping an event position to a data point https://codedaily.io/tutorials/39/Make-a-Beautiful-Interactive-Bitcoin-Price-Chart-with-React-D3-and-VX#Tooltip-Handler

Achilles718611:
It doesn't work in recent vx version.

![image](https://user-images.githubusercontent.com/31845458/58066690-fe2de700-7bbc-11e9-8551-27b7c1b0870f.png)

hshoff:
@Achilles718611 Correct, there was a breaking change in [v0.0.181](https://github.com/hshoff/vx/releases/tag/v0.0.181) for details please see: https://github.com/hshoff/vx/pull/383

```diff
-  onClick={data => event => this._handleLineSelect(event)}
+  onClick={event => this._handleLineSelect(event)}
```

Achilles718611:
Thanks for your kind reply.
But I have to pass data in click event.

hshoff:
`data` is defined in the outer scope so it's still available to use in your event handlers:

```jsx
<LinePath
  data={data[key]}
  // ...
  onClick={event => {
    console.log('my data', data[key]);
    this._handleLineSelect(event);
  }}
/>
```

Achilles718611:
data is object which is consist of key and value(array).
So data[key] is array.
I would like to pass one element of array.
When I use 0.0.179, it's possible.
d => event => {
  .....
}
In 0.0.179 d is what I want to pass value.

hshoff:
Here's an example of how to accomplish this: https://codesandbox.io/s/bitcoinchart-syeeo?fontsize=14&module=%2Fsrc%2FApp.js

Achilles718611:
Got it.
Thank you.

Issue #199, opened 2017-11-15
Event handlers on pie charts?


erosenberg:
Hello, just wanted to know if it's possible to add event handlers to pie charts. Essentially, I want a label elsewhere on my page to update when I hover over a specific data set. Is this possible? I can work around it, but it would be nice functionality to have if not there already. #

hshoff:
Hi @erosenberg, thanks for checking out vx! Yes, events on a pie chart are possible. 

![vx-pie-events](https://user-images.githubusercontent.com/339208/32869269-e52822ba-ca2a-11e7-9b92-60a716001fb5.gif)

demo: https://codesandbox.io/s/mjp8lmvop?expanddevtools=1

the important bit:

```js
<Pie
  // ...
  onMouseEnter={data => event => {
    console.log("mouse enter", data.data.letter);
  }}
/>
```

erosenberg:
This worked excellently. Thanks, @hshoff ! Was that that somewhere in the docs and I just missed it?
Also, amazing job on this and thanks again for the speedy response!

libnine:
Any possibility we can get an updated example? I'm trying to do exactly as above but my `onMouseEnter` function can't seem to grab the section it hovers over. Ultimately, i'm trying to implement a tooltip with the underlying data.

Here's a demo: https://codesandbox.io/s/jolly-burnell-uxzh7?file=/src/App.js

hshoff:
@libnine sure thing, here ya go: https://codesandbox.io/s/strange-lehmann-vdoff?file=/src/App.js

libnine:
Thank you so much Harrison!

Issue #50, opened 2017-06-02
<Bar> onClick handler is called on render, not on click


donpinkus:
I'm noticing something odd - it seems that the `<Bar>` component's `onClick` handler is called when it is rendered. 

I noticed it while following the example in VX's gallery: https://vx-demo.now.sh/bars

Here is code to reproduce it. 

It has a `console.log` on render, when the SVG is clicked, and when the Bar is clicked. Notice that the console.logs for the `bar` being clicked are all present after each render.

```
import React, { Component } from 'react';
import { Bar } from '@vx/shape';
import { Group } from '@vx/group';
import { GradientTealBlue } from '@vx/gradient';
import { letterFrequency } from '@vx/mock-data';
import { scaleBand, scaleLinear } from '@vx/scale';
import { extent, max } from 'd3-array';

export default class VXBarTest extends Component {
  render(){
    console.log('render');
    const data = letterFrequency.slice(5);

    function round(value, precision) {
      var multiplier = Math.pow(10, precision || 0);
      return Math.round(value * multiplier) / multiplier;
    }

    // accessors
    const x = d => d.letter;
    const y = d => +d.frequency * 100;

    const width = 750;
    const height = 400;

    if (width < 10) return null;

    // bounds
    const xMax = width;
    const yMax = height - 120;

    // scales
    const xScale = scaleBand({
      rangeRound: [0, xMax],
      domain: data.map(x),
      padding: 0.4,
    });

    const yScale = scaleLinear({
      rangeRound: [yMax, 0],
      domain: [0, max(data, y)],
    });

    const bars = data.map((d, i) => {
      const barHeight = yMax - yScale(y(d));

      return (
        <Group key={`bar-${x(d)}`}>
          <Bar
            width={xScale.bandwidth()}
            height={barHeight}
            x={xScale(x(d))}
            y={yMax - barHeight}
            fill="rgba(23, 233, 217, .5)"
            onClick={ function(){ console.log('bar clicked'); }}
          />
        </Group>
      );
    });

    return (
      <svg
        width={width}
        height={height}
        onClick={ function(){ console.log('svg clicked'); }}>
        <GradientTealBlue id="teal" />
        <Group top={40}>
          {bars}
        </Group>
      </svg>
    );
  }
}
```

donpinkus:
I suspect this is some bug related to my code, if

hshoff:
[...]

Issue #220, opened 2018-01-03
Render props


techniq:
Creating a list of existing HOC components we should updated to have least have a render prop equivalent (feel free to add to the list)

- `<ScreenSize>`
  - Convert [withScreenSize](https://github.com/hshoff/vx/blob/master/packages/vx-responsive/src/enhancers/withScreenSize.js) similar to what we did with [ParentSize](https://github.com/hshoff/vx/blob/master/packages/vx-responsive/src/components/ParentSize.js)
- `<BoundingRects>`
  - Convert [withBoundingRects](https://github.com/hshoff/vx/blob/2cf9ade4c36337199fec9d1571f84e197fff3d83/packages/vx-bounds/src/enhancers/withBoundingRects.js)
- `<Tooltip>`
  - Convert [withTooltip](https://github.com/hshoff/vx/blob/4f128fb067fcda3bb264438e03b0b7ef345b6f15/packages/vx-tooltip/src/enhancers/withTooltip.js), [Tooltip](https://github.com/hshoff/vx/blob/4f128fb067fcda3bb264438e03b0b7ef345b6f15/packages/vx-tooltip/src/tooltips/Tooltip.js), and [TooltipWithBounds](https://github.com/hshoff/vx/blob/4f128fb067fcda3bb264438e03b0b7ef345b6f15/packages/vx-tooltip/src/tooltips/TooltipWithBounds.js)
  - Need to think how best these should compose together.  We might not need all three and just use composition of the components

hshoff:
thanks @techniq!

williaster:
We could possibly remove `TooltipWithBounds` if we converted `withBoundingRects` -> `BoundingRects` because things could be more composable:

```javascript
const top = ...;
const left = ...;

<BoundingRects>
  {({ rect, parentRect }) => (
    <Tooltip 
        style={{ 
          top: rect && parentRect && rect.right > parentRect.right ? (left - rect.width) : left, 
          left: rect && parentRect && rect.bottom > parentRect.bottom ? (top - rect.height) : top, 
          ...rest 
     }}>
  )}
</BoundingRects>
```

though this still requires the user to do the right rect comparison, which isn't hard but is basically boilerplate.

techniq:
@williaster could `BoundingRects` expose some helpers passed down as well, such as

const top = ...;
const left = ...;

```
<BoundingRects>
  {({ rect, parentRect, getTop, getLeft }) => (
    <Tooltip 
        style={{ 
[...]

PR #383, opened 2018-11-18
RFC: Consistency and Control

hshoff:
Breaking changes are painful. I try to avoid them whenever possible. Getting vx to `v1` means I need to make a larger breaking change than I am typically comfortable with as a library author. In short, I made a poor design decision in the early days of this library that I now wish to fix. I hope you'll agree with me that it's for the best and will lead to a brighter future with more consistency and control.  

### Summary

This pull request does the following:

- [x] removes two utility functions used by vx components: [`additionalProps()`](https://github.com/hshoff/vx/blob/master/packages/vx-shape/src/util/additionalProps.js) and [`callOrValue()`](https://github.com/hshoff/vx/blob/master/packages/vx-shape/src/util/callOrValue.js)
- [x] adds the more common react pattern of [render props](https://reactjs.org/docs/render-props.html#using-props-other-than-render) to components impacted by the removal of the above two utils.
- [x] removes scale props in favor accessors that return scaled values
- [x] adds more PropTypes
- [x] adds more tests
- [x] update all [vx-demo.now.sh/gallery](https://vx-demo.now.sh/gallery) tiles to new api

### Motivation

Almost 2 years ago I started working on a library of components that brought together `react` and `d3` at a lower-level than traditional charting libraries. Unfortunately, I made an api design choice that I've regretted for a long time now and it is time to rectify before releasing v1.

### Data Binding Before
Initially I thought the target audience would be existing `d3` developers that found themselves in a react codebase. So I went about making an api that would feel similar to d3 when it came to data binding.

Let's walk through an example of rendering out a series of bars that are different colors based on each datum's name. (`data = [{ name: 'captain gary' }, // ...]`)

```js
// d3
svg.append("g")
 .selectAll("rect")
 .data(data)
 .enter().append("rect")
  .attr("fill", datum => color(datum.name))
```
So one coffee-fueled morning I thought "fine I'll bind props to data in react if you pass a function as a prop"
 
```jsx {4}
// vx
<Bars
  data={data}
  fill={datum => color(datum.name)}
/>
```
But wait, what about props that are typically functions like event handlers? Say you wanted to click on each colored bar and alert the bar's name. Well you would just pass a function that returns your click handler with the datum as the first argument.

```jsx
// vx
<Bars
  data={data}
  fill={datum => color(d.name)}
  onClick={datum => event => {
   alert(d.name);
  }}
/>
```
🤦‍♂️ 🤦‍♂️ 

*__What's so bad about this?__* It no longer feels like `react`. vx components become _special_ in the sense you need to know about this silly pattern and it's not like other react components. This is confusing: https://github.com/hshoff/vx/issues/50 https://github.com/hshoff/vx/issues/199 https://github.com/hshoff/vx/issues/328 

### Data Binding After
Then came the realization that the target audience for `vx` was really `react` developers that needed data visualization without having to learn `d3`.

Seems like a confusing api tradeoff for the added ability of setting a prop based on a datum value. Along came render props. Let's take a look at what the new api looks like with render props.

```jsx
<Bars
  data={data}
  // ...
>
   {bars => {
    return bars.map((bar, i) => {
      return (
        <rect
          key={bar.name}
          fill={color(bar.name)}
          onClick={event => alert(bar.name)}
        />
      );
     })
   }}
</Bars>
```
This means vx components are no longer opaque boxes with special rules. You now have more control over what's rendered and it's more transparent how it works following common patterns in react land.

This is a change I've wanted to make for a long time. It took a lot boring work to make it happen, but I think it's finally ready.

### What about Hooks?

I was just as surprised as everyone else about the react [Hooks](https://reactjs.org/docs/hooks-overview.html) announcement at ReactConf. The day they were announced, I was finishing migrating all of the gallery tiles to the new api and I had the panic moment of, "oh no this new api is going to be dead-on-arrival". 

The Hooks announcement talked about how hooks can replace render props with the benefit of not introducing a false hierarchy to your render. This is great and I look forward to updating vx to include hooks. That said, Hooks are still a proposal available on an alpha release of react. Currently, vx supports react 15 and 16. vx will look to support hooks in a future release. I see the render props api as an immediate improvement for all users of vx where hooks would only support those lucky enough to be in a codebase on 16.7 alpha.

Initial thoughts are hooks are the future of react, so vx will support them when they are more widely available. 

### Accessors + Scales Before

In the beginning, I found myself often having to apply my scale on the accessor value. So I turned this into a pattern in vx. You would pass in your scale and your accessor separately and vx would apply them for you. It used look like this:

```jsx
 <LinePath
  data={data}
  xScale={xScale}
  yScale={yScale}
  x={x}
  y={y}
/>
```
behind the scenes vx would create a line path from d3-shape's line generator

```js
const path = line()
  .x((...args) => xScale(x(...args)))
  .y((...args) => yScale(y(...args)))
  .defined(defined)
.curve(curve);
```

Again this isn't transparent. Let's make it transparent.

### Accessors + Scales After

```jsx
<LinePath
  data={data}
  x={d => xScale(x(d))}
  y={d => yScale(y(d))}
/>
```
This api simplifies what vx is doing behind the scenes as well to:

```js
const path = line();
if (x) path.x(x);
if (y) path.y(y);
if (defined) path.defined(defined);
if (curve) path.curve(curve);
```
This gives transparency, consistency with the d3 api, and gives you more control over shape generation.

### Closing thoughts

I worry about the early adopters that are not going to be happy about having to update their code due to these breaking changes, but the longer the I wait to make this the bigger the hill will be to climb.

Would love to hear your thoughts.

#### :boom: Breaking Changes

- [glyph][breaking] rm additionalProps, add children as fn 
- [shape][breaking] rm additionalProps, add children as fn 
- [geo][breaking] rm additionalProps, add children as fn 
- [heatmap][breaking] rm additionalProps, add children as fn 
- [stats][breaking] rm additionalProps, add children as fn 
- [boxplot][breaking] rm additionalProps, add children as fn 
- [voronoi][breaking] rm additionalProps, add children as fn 
- [legend][breaking] rm additionalProps, add children as fn 

#### :house: Internal

- [demo] update gallery tile examples to new apis

cc @williaster @conglei @techniq


techniq:
Should be consider dropping our `HOCs` as part of this breaking change?  See: https://github.com/hshoff/vx/issues/220

sto3psl:
> But wait, what about props that are typically functions like event handlers?

I always kind of liked the `datum => event => {}` style. It worked well if one doesn't have to customize the rendered component that much. If I understand it correctly, with the new API, one has to always render the actual SVG and vx just provides the data. This makes a lot of sense in the low-level nature of vx.

For the repo I'm working on it isn't a big deal so go for it. I like it when things get more consistent and changes will reduce overall confusion for new users! We use vx since `v0.0.158`.

The scale and accessor change is really nice as well. More power to the user.

Nice job on the proposed changes and that's quite some work you had to do there 😅!

PR #384, opened 2018-11-23
Prevent exception when drag event target not under svg

madprog:
#### :bug: Bug Fix

- In order to have a continuous dragging even if my mouse gets out of the `<svg>` element, I attach `dragMove` and `dragEnd` to the `window` object in my code.
  However, these events are no longer guaranteed to have a `target` under a `<svg>`, provoking an exception in `@vx/event`:

      vx-event.es.js:15 Uncaught TypeError: Cannot read property 'ownerSVGElement' of undefined
          at localPoint (vx-event.es.js:15)
          at Drag.dragMove (vx-drag.es.js:121)
          at DragViewBase._this.handleMouseMove (DraggableSvg.tsx:75)

  This exception is fixed by the changes of this PR.


boy51:
Hi @madprog, I'm digging up this old PR because I'm currently facing the same problem. It seems however the implementation has changed a bit since this PR was opened. I made an issue about this: #1172 , maybe you could help me out since you seemed to have solved this?

So I believe you and I used the same approach, whcih is to attach event listeners to window instead of the draggable object. It understand back when you made this the problem was that it was looking for the SVG which may not be there when event.target is the window.

My problem is similar. The way I see it, when dragging outside the browser, we are forced to use window as `event.target`. Right? However, on mousedown there is no way to have window be the `event.target`. This results in mousedown and mousemove to have different event.targets, which results in an annoying jump whenever we pan, since it uses `getDOMRect` to calculate the pan.

So right now the only way I see to solve this is to either change the implementation to use `event.srcElement` or to implement some kind of offset in `dragMove`. Both are not very awesome and I'm wondering if there is a way to do this without changing implementations.

I'm not sure how to solve the problem that using `window.addEventListener` will return a `MouseEvent` instead of a `React.SyntheticEvent` which `dragStart` expects.

I'd seriously appreciate your input on how you solved this. You can also shoot me an email if you prefer. Thanks!

haiderRizvii (approved):

PR #399, opened 2019-01-04
[docs] generate api docs from prop-types, sync to website

hshoff:
Still need to document each prop for all packages, but automating the documentation of prop-types is big step forward.

screen shot 2019-01-03 at 9 23 34 pm
screen shot 2019-01-03 at 9 23 15 pm

#### :memo: Documentation

- [docs] use [react-docgen](https://github.com/reactjs/react-docgen) to generate docs from prop-types and comments.

#### :house: Internal

- [docs] add script to sync files -> docs -> readme -> [vx-demo.now.sh/docs](https://vx-demo.now.sh/docs)


packages/vx-axis/src/axis/AxisBottom.js
```diff
@@ -5,31 +5,106 @@ import Axis from './Axis';
...
+  /**
+   * The pixel value for the width of the lines.
+   */
   strokeWidth: PropTypes.number,
+  /**
+   * The pattern of dashes in the stroke.
+   */
   strokeDasharray: PropTypes.string,
+  /**
+   * The class name applied to each tick grou
```
williaster: copying some of these onto `AxisRadial` and noticed missing `p` 😝

PR #404, opened 2019-01-07
[axis] add `AxisRadial` component

williaster:
#### :rocket: Enhancements

This PR 
- adds a new `<AxisRadial />`  component to the `@vx/axis` package
- adds `propType` descriptions and tests for `<AxisRadial />`
- updates the `LineRadial` demo tile to also include the `<AxisRadial />` component:

### DIFF
--- a/packages/vx-axis/src/axis/AxisRadial.js
+++ b/packages/vx-axis/src/axis/AxisRadial.js
@@ -8,6 +8,7 @@ import { Text } from '@vx/text';
 import { curveCardinal } from '@vx/curve';
 
 import ORIENT from '../constants/orientation';
+import polarToCartesian from '../utils/polarToCartesian';
 import radialLabelTransform from '../utils/radialLabelTransform';
 import radialTickLabelTransform from '../utils/radialTickLabelTransform';
 import identity from '../utils/identity';
@@ -42,13 +43,21 @@ const propTypes = {
    */
   labelClassName: PropTypes.string,
   /**
-   * Pixel offset of the axis label (does not include tick length or tick label font size, which is accounted for automatically)
+   * Pixel offset of the axis label (does not include tick length or
+   * tick label font size, which is accounted for automatically)
    */
   labelOffset: PropTypes.number,
   /**
-   * Props applied to the axis label component.
+   * High-leevl positioning of axis label component. Can be
+   * refined or customized further using labelProps.
    */
-  labelOrientation: PropTypes.oneOf(['top', 'right', 'bottom', 'left', 'center']),
+  labelOrientation: PropTypes.oneOf([
+    ORIENT.top,
+    ORIENT.right,
+    ORIENT.bottom,
+    ORIENT.left,
+    ORIENT.center
+  ]),
   /**
    * Props applied to the axis label component.
    */
@@ -106,7 +115,8 @@ const propTypes = {
    */
   tickTransform: PropTypes.string,
   /**
-   * An array of values that determine the number and values of the ticks. Falls back to `scale.ticks()` or `.domain()`.
+   * An array of values that determine the number and values of the ticks.
+   * Falls back to `scale.ticks()` or `.domain()`.
    */
   tickValues: PropTypes.array,
   /**
@@ -120,7 +130,7 @@ const propTypes = {
   /**
    * For more control over rendering or to add event handlers to datum, pass a function as children.
    */
-  children: PropTypes.func,
+  children: PropTypes.func
 };
 
 export default function AxisRadial({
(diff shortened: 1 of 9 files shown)


packages/vx-axis/src/axis/AxisRadial.js
```diff
@@ -0,0 +1,287 @@
...
+  let format = scale.tickFormat ? scale.tickFormat() : identity;
+  if (tickFormat) format = tickFormat;
+
+  let tickLabelFontSize = 10; // track the max tick label size to compute label offset
+
+  const ticks = values
+    .map((val, index) => {
+      if (hideZero && val === 0) return null;
+
+      const angle = scale(val) - Math.PI / 2;
```
williaster: Maybe haven't had enough ☕️ but I don't fully understand why the `-🥧/2` is needed ... but with different scale ranges (e.g., `[-🥧, 🥧]`, `[0, 2*🥧]`), the axis was always off by 45° without this.

packages/vx-axis/src/utils/radialTickLabelTransform.js
```diff
@@ -0,0 +1,28 @@
+const CENTER_ALIGN_THRESHOLD = 1;
+
+export default function radialTickLabelTransform(fromPoint, toPoint) {
+  let textAnchor = 'middle';
+  let verticalAnchor = 'middle';
+  const dx = toPoint.x - fromPoint.x;
+  const dy = toPoint.y - fromPoint.y;
+
+  // offset the label in the same direction as the tick orientation
+  const x = toPoint.x + dx / 2;
```
williaster: this offset is somewhat arbitrary but gave decent results as you can see in the demo tile. we could possibly make it more configurable.

packages/vx-axis/src/axis/AxisRadial.js
```diff
@@ -0,0 +1,287 @@
...
+      return tickLabelPropsObj;
+    })
+    .filter(defined => !!defined);
+
+  const axisLineData = scale.ticks(50);
+  const angle = d => scale(d);
+
+  if (children) {
+    return (
+      <Group className={cx('vx-axis-radial', axisClassName)} top={top} left={left}>
```
williaster: all classes are the same as other `Axis*` components except this top-level one which should be sufficient to distinguish from other (linear) axes on the page.

packages/vx-axis/src/axis/AxisRadial.js
```diff
@@ -0,0 +1,287 @@
...
+      {!hideAxisLine && (
+        <LineRadial
+          className={cx('vx-axis-line', axisLineClassName)}
+          data={axisLineData}
+          angle={angle}
+          radius={radius}
+          stroke={stroke}
+          strokeWidth={strokeWidth}
+          strokeDasharray={strokeDasharray}
+          curve={curveCardinal}
```
hshoff: does this always have to be `curveCardinal`? should it be configurable via prop?

williaster: I think you'd probably want this most of the time, otherwise the axis wouldn't look like an arc.

In thinking about this more though, I changed the implementation of `GridRadial` to use `Arc` instead of `LineRadial`, that could simplify this as well and allow us to drop the `@vx/curve` dep. With `LineRadial` we're sort of doing the arc interpolation ourself but `Arc` should be best at that.

MrBlenny:
@williaster this looks great! Where did this PR get to?

This is exactly what I'm after in order to create some gauge-style components.
Screenshot from 2021-07-03 12-03-07

Is there some way I can help in order to get this merged?

williaster:
hey @MrBlenny 👋  the gauge looks great! 😍 

This PR was never completed/I lost bandwidth to get it merged. @sarathps93 implemented the `radial` versions of `@visx/grid` in #1007 , I'm not sure if this is still on his radar or not. If not, and you are interested in contributing this for others to use, I would happily review a PR!

packages/vx-axis/src/utils/radialLabelTransform.js
```diff
@@ -0,0 +1,32 @@
...
+    y = -radius - offset;
+  } else if (labelOrientation === 'right') {
+    x = radius + offset;
+    textAnchor = 'start';
+  } else if (labelOrientation === 'bottom') {
+    y = radius + offset;
+  } else if (labelOrientation === 'left') {
+    x = -radius - offset;
+    textAnchor = 'end';
+  }
```
haiderRizvii, line 29: use switch case

haiderRizvii (changes requested):

PR #405, opened 2019-01-08
[grid] add `GridRadial` and `GridPolar`

williaster:
#### :rocket: Enhancements

- This PR adds `<GridRadial />` and `<GridPolar />` components to `@vx/grid`, and updates the `RadialLine` demo to be implemented using them.

image

#### :bug: Bug Fix

- Also fixes a bug in the `@vx/shape/Arc` component where passing any prop (e.g., `innerRadius`) as `0` has no effect because of the `if/else` logic. This is needed for the implementation of `<GridRadial />` because in order to use `d3`'s `arc()` function without data, you must fully specify `inner/outerRadius` and `start/endAngle`, so I included it in this PR.

Will push the doc changes later tonight.


williaster:
to clarify what the grid on its own looks like

image

hshoff:
Thought about this more, how about we rename `<GridPolar/>` to `<GridAngle />`?

Then we could make a `<GridPolar />` which uses both `<GridAngle />` and `<GridRadial />` (like the cartesian coordinate counterpart `<Grid />` which uses `<GridRows />` and `<GridColumns />`).

williaster:
closed in favor of #1007

PR #423, opened 2019-02-12
Fixed broken BarStacks example

EugeneDraitsev:
#### :memo: Documentation

- fixed broken BarStacks example. Bar Stack Horizontal example works correct, but BarStack for some reason uses `({ barStacks })` instead of `barStacks`


MrMonsk:
This is inline w/ the previous https://github.com/hshoff/vx/pull/409 PR to fix the Horizontal

PR #431, opened 2019-03-10
use getDerivedStateFromProps in vx/text to stop strict warnings 

dagda1:
#### :rocket: Enhancements
vx/text gives the following warning when running in a strict react 16 build:

>main.js:10481 Warning: Unsafe lifecycle methods were found within a strict-mode tree:
    in StrictMode
    in Unknown

> componentWillMount: Please update the following components to use componentDidMount instead: Text

> componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Text

> Learn more about this warning here:
-

I've moved the initial call to `updateWordsByLines` into the constructor and used `getDerivedStateFromProps` instead of `componentWillReceiveProps`

### DIFF
--- a/packages/vx-text/src/Text.js
+++ b/packages/vx-text/src/Text.js
@@ -9,13 +9,11 @@ class Text extends Component {
     this.state = {
       wordsByLines: []
     };
-  }
 
-  componentWillMount() {
     this.updateWordsByLines(this.props, true);
   }
 
-  componentWillReceiveProps(nextProps) {
+  static getDerivedStateFromProps(nextProps) {
     const needCalculate =
       this.props.children !== nextProps.children || this.props.style !== nextProps.style;
     this.updateWordsByLines(nextProps, needCalculate);


hshoff:
Currently vx supports React 15 and 16. Updating these lifecycle methods is something that needs to happen for React 17 support and I was waiting for a React 17 release candidate to start work on upgrading all lifecycle methods.

If we landed this update in @vx/text it would me setting React peerDep to >=16.3.0 for @vx/text in the next release which would be inconsistent with the other packages. So I want to hold off on making these changes until ready to do it all at once.

packages/vx-text/src/Text.js
```diff
@@ -9,13 +9,11 @@ class Text extends Component {
...
-  }
 
-  componentWillMount() {
     this.updateWordsByLines(this.props, true);
   }
 
-  componentWillReceiveProps(nextProps) {
+  static getDerivedStateFromProps(nextProps) {
     const needCalculate =
       this.props.children !== nextProps.children || this.props.style !== nextProps.style;
```
sto3psl: The change is unfortunately not that easy. `getDerivedStateFromProps` doesn't have access to the component instance, which means no `this`. As a second, `getDerivedStateFromProps` needs to return a state object or null if nothing changes, much like `setState` because it runs right before `render`.

More info can be found in the React Docs.  
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

dagda1:
@sto3psl you are absolutely right, terrible oversight from me.  I opened [this PR](https://github.com/hshoff/vx/pull/394) originally but nobody responded.

I did notice the instance `this` first time round.  Out of curiosity, is that PR any better?

sto3psl:
@dagda1 I'm not quite sure just from looking at it. As @hshoff already mentioned, these lifecycles won't go anywhere in a reasonable timeframe. React 17 isn't coming soon and there have been no mentions from the React team. 

I'm not entirely sure if this will help with the warnings but you could try renaming the unsafe lifecycles to `UNSAFE_componentWillReceiveProps` and `UNSAFE_componentWillUpdate`.

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

denkristoffer:
@hshoff With the the latest [release of React 16.9](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html) is it still the plan to wait for React 17? The lifecycle methods have been renamed and will continue to work in 17 whenever it comes, which seems to be a long time away.

hshoff:
@denkristoffer now that `16.9` is out and folks are starting to see the warnings, I think it's time. In an upcoming release vx will have a react & react-dom peerDep of 16.3<= & =<16.9

Likely plan is to rename deprecated lifecycle methods to `UNSAFE_*` which will remove the warnings then follow up with migrating away from the unsafe methods.

PascalSenn:
@hshoff I'm not sure if it makes sense to use the prefixed methods. Those warnings are there for a reason. 
When there are no errors I assume that I can update react without major issues
It could happen that someone hasn't upgraded VX to the newest version but upgrades react once it's released
I would just leave them and replace them as you go

PR #458, opened 2019-05-06
Only call 'children' if it is a function

### LINKED ISSUE #397: Is preact still compatible?
Hello,

I tried to use vx/vx with preact.
I do like to use both of them.
But, most examples is not working.

```
vx-shape.es.js?83e7:418 Uncaught TypeError: children is not a function
    at Component.AreaClosed [as constructor] (vx-shape.es.js?83e7:418)
    at Component.doRender [as render] (preact.js?10a9:229)
    at renderComponent (preact.js?10a9:271)
    at setComponentProps (preact.js?10a9:248)
    at buildComponentFromVNode (preact.js?10a9:339)
    at idiff (preact.js?10a9:134)
    at innerDiffNode (preact.js?10a9:176)
    at idiff (preact.js?10a9:152)
    at innerDiffNode (preact.js?10a9:176)
    at idiff (preact.js?10a9:152)
```

Only this one is working fine.
https://github.com/hshoff/vx/blob/master/packages/vx-demo/components/charts/SimpleBar.js

Do I have to do something for children attribute?

This page for supporting preact shows 404 error.
https://zeit.co/deployments/vx-preact.now.sh?redirect=1

I deeply appreciate for your time and efforts for such the great library.

james-kraus-ck:
#### :rocket: Enhancements

No longer throws because preact passes an empty array instead of `undefined` when there are no children

#### :bug: Bug Fix

#397


williaster:
Sorry this was never merged. #507 now constrains the `children` prop to a function, which should provide better type safety. 

However if this still fails with `preact` we can merge this if we rebase on master. Will keep an eye on it / merge asap so it doesn't get stale.

williaster:
Going to close this since it's been inactive for so long

PR #468, opened 2019-06-13
feat: add TypeScript declaration for vx/responsive

kristw:
#### :rocket: Enhancements

Add typescript declaration for `@vx/responsive`. 

We have several typescript projects that use this library and ended up writing ad-hoc declaration files, so having this included in the package will be much better. I have read through all the jsx code and added types for all items in this package.

cc: @williaster @conglei


williaster:
going to close this for now and hope we can re-use the definitions in the [Typescript Project](https://github.com/hshoff/vx/projects/2)

PR #484, opened 2019-08-22
internal: Refactor to nimbus

williaster:
#### :house: Internal

This PR is a WIP for moving build configuration to `nimbus`, which is a soon-to-be open-sourced shareable build config for `jest`, `babel`, `eslint`, `prettier`, and `typescript`. This will help 
- eliminate duplicate configs across individual packages 
- move config to the root for consistency and easier upgrades
- enable easier support for `typescript` in the future 😉  

I will enumerate the changes as the PR evolves below, will try and convert a package at a time, I think the first couple were the hardest.

**Changes (dev)**
- Use `babel` not `rollup`
- Use `yarn` not `npm`
  - this will enable `workspaces` so that we can push all config to the root instead of duplicating across every package as is the case now
- Stricter linting rules (e.g., `.jsx` required for `React` files)

**Changes (consumers)**
- [breaking] Deprecate `build/` and `dist/`, use `lib/` and `esm/`
- [breaking] Deprecate umd builds

**General TODO**
- [x] point to open-source libs when published 
- [x] update docs
- [x] verify all package changes functionally with demo

**Packages to convert** 
- [x] `root`
- [x] `vx-annotation`
- [x] `vx-axis`
- [x] `vx-bounds`
- [x] `vx-brush`
- [x] `vx-curve`
- [x] `vx-event`
- [x] `vx-gradient`
- [x] `vx-heatmap`
- [x] `vx-marker`
- [x] `vx-pattern`
- [x] `vx-scale`
- [x] `vx-text`
- [x] `vx-voronoi`
- [x] `vx-chord`
- [x] `vx-demo`
- [x] `vx-geo`
- [x] `vx-tooltip`
- [x] `vx-group`
- [x] `vx-boxplot`
- [x] `vx-drag`
- [x] `vx-clip-path`
- [x] `vx-network`
- [x] `vx-legend`
- [x] `vx-glyph`
- [x] `vx-stats`
- [x] `vx-responsive`
- [x] `vx-threshold`
- [x] `vx-shape`
- [x] `vx-point`
- [x] `vx-mock-data`
- [x] `vx-hierarchy`
- [x] `vx-grid`
- [x] `vx-zoom` 
- [x] `vx-vx`

@hshoff @milesj @schillerk @techniq @kristw

PR #488, opened 2019-09-11
typescript(vx-group): re-write package in TypeScript

### LINKED ISSUE #510: Re-write `@vx/group` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off of #484, introducing `TypeScript` config and converting the `@vx/group` package (which is quite simple) as a proof of concept. The plan is to convert packages one by one from here 📈 

<p align="center">
  
</p>

Notes 
- `@types/*` related to testing were added to the root `devDeps`
- package-specific types should be added as `deps` in the relevant packages
- `@airbnb/config-babel` adds [`babel-plugin-typescript-to-proptypes`](https://github.com/milesj/babel-plugin-typescript-to-proptypes) when used with the `typescript` driver, which generates `PropTypes` for react component `type`s/`interface`s and requires us to keep the `prop-types` dep though it's no longer referenced in source code. Example compiled `.js` file below:

<p align="center">
  
</p>

Closes #510 

@hshoff @milesj @schillerk @techniq 

Fixes https://github.com/hshoff/vx/projects/2#card-26383081


package.json
```diff
@@ -31,6 +31,7 @@
     "precommit": "lint-staged",
     "setup": "yarn run build",
     "test": "yarn run jest",
+    "type": "nimbus typescript --build --reference-workspaces",
```
milesj: Should only emit declarations, otherwise the TS build would overwrite the Babel build.

williaster: 🤦‍♂ done

williaster: oops not compatible with `--reference-workspaces`

package.json
```diff
@@ -31,7 +31,7 @@
     "precommit": "lint-staged",
     "setup": "yarn run build",
     "test": "yarn run jest",
-    "type": "nimbus typescript --build --reference-workspaces",
+    "type": "nimbus typescript --emitDeclarationOnly --reference-workspaces",
```
milesj: `--build` is necessary for TS project refs :P

packages/vx-group/package.json
```diff
@@ -34,6 +35,8 @@
     "react": "^15.0.0-0 || ^16.0.0-0"
   },
   "dependencies": {
+    "@types/classnames": "^2.2.9",
+    "@types/react": "^16.9.2",
```
williaster: noting that we should make this less strict

williaster:
whoops, looks like I failed to update the branch this should be merged into after `chris--nimbus` was merged. should I open another PR for the same into master?

PR #490, opened 2019-09-16
typescript(vx-point): Re-write package in TypeScript

### LINKED ISSUE #491: Re-write `@vx/point` in TypeScript

williaster:
#### 🚀 Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/point` package to `TypeScript`.

Closes #491 

**Testing**
- [x] CI
- [x]  Generates .d.ts files

@hshoff @schillerk


packages/vx-point/src/Point.ts
```diff
@@ -1,4 +1,7 @@
 export default class Point {
+  public x: number = 0;
```
williaster: @hshoff I'm not sure if you intended these to be `public` or `private`

williaster: I think they need to be `public` because much of the codebase references them as `point.x/y` 👍

PR #493, opened 2019-09-17
typescript(vx-event): Re-write package in TypeScript

### LINKED ISSUE #492: Re-write `@vx/event` in TypeScript

williaster:
#### 🚀 Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/event` package to `TypeScript`. 

The functions in this package were a bit gnarly/confusing to follow with different call signatures and duplicative, so I consolidated and cleaned them up with type guards.

Depends on #490 (`@vx/point`) for types, closes #492.

**Testing**
- [x] CI Depends on #490 (@vx/point) for types
- [x]  Generates .d.ts files
- [x] tested functionally 
  - [x] `/areas` and `/zoom-i` in `/gallery` (which use `@vx/event`)
  - [x] modified `/areas` to test the two call signatures of `localPoint`


packages/vx-event/src/typeGuards.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// functional definition of TouchEvent
+export function isTouchEvent(event?: Event): event is TouchEvent {
+  return !!event && 'changedTouches' in event;
+}
+
+// functional definition of event
+export function isEvent(event?: Event | Element): event is Event {
+  return !!event && 'target' in event;
```
milesj: `event instanceof Event`?

williaster:
(noting that the build is failing because of the missing `@vx/point` types; could `@ts-ignore` to remove this dependency but `@vx/point` is quite straight forward)

packages/vx-event/src/typeGuards.ts
williaster: I wasn't 100% sure if that would play well with `React` `events`, I'll give it a try functionally and update if it works 👍

williaster: looks like `event instanceof Event` is false for `React` `events`

packages/vx-event/src/localPointGeneric.ts
```diff
@@ -0,0 +1,33 @@
+import { Point } from '@vx/point';
+import { isSVGElement, isSVGGraphicsElement, isSVGSVGElement } from './typeGuards';
+import getXAndYFromEvent from './getXAndYFromEvent';
+
+export default function localPoint(node: Element, event: MouseEvent | TouchEvent) {
```
williaster: should update all of the event types to also accept `React.MouseEvent | React.TouchEvent`

packages/vx-event/src/typeGuards.ts
```diff
@@ -0,0 +1,27 @@
+export function isElement(elem?: Element | Event): elem is Element {
```
williaster: should update all of the event types to also accept `React.MouseEvent | React.TouchEvent`

packages/vx-event/src/typeGuards.ts
milesj: Yeah React events are `SyntheticEvent`, which have a property `nativeEvent`, that points to the actual `Event`.

williaster: 👍 I added `@types/react` to this package so we could support `React.MouseEvent | React.TouchEvent`, and updated this check to include `instance of` on either the passed `event` or `event.nativeEvent`. Seems like a good solution.

packages/vx-event/src/localPoint.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default function localPoint(nodeOrEvent: Element | EventType, maybeEvent?: EventType) {
+  if (isElement(nodeOrEvent) && maybeEvent) {
+    return localPointGeneric(nodeOrEvent, maybeEvent);
+  }
+  if (isEvent(nodeOrEvent)) {
+    const node = nodeOrEvent.target as Element;
+    if (node) return localPointGeneric(node, nodeOrEvent);
+  }
+  return undefined;
```
williaster: @hshoff there are a few places in other packages (e.g., [`drag`](https://github.com/hshoff/vx/blob/master/packages/vx-drag/src/Drag.js#L23)) that expect to always get a value from this, I think we should probably return `{ x: 0, y: 0 }` by default.

packages/vx-event/src/getXAndYFromEvent.ts
```diff
@@ -0,0 +1,17 @@
+import { isTouchEvent } from './typeGuards';
+
+export default function getXAndYFromEvent(event?: MouseEvent | TouchEvent) {
+  if (!event) return { x: 0, y: 0 };
+
+  if (isTouchEvent(event)) {
+    return {
+      x: event.changedTouches[0].clientX,
```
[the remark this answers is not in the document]
williaster: not 100% sure, TS isn't complaining but I'll do a `length` check to be sure. good catch 🏆

packages/vx-event/src/localPoint.ts
hshoff: oh i see, you answered my questions below. hmm, not ideal. 

Maybe we could make a new `NullPoint` class and return that instead? `new NullPoint() => { x: null, y: null }`

then in Drag we can instanceof check the returned value and if it's `NullPoint` and handle the nulls there.

packages/vx-event/src/getXAndYFromEvent.ts
```diff
@@ -0,0 +1,22 @@
+import { EventType } from './types';
+import { isTouchEvent } from './typeGuards';
+
+const DEFAULT_POINT = { x: 0, y: 0 };
+
+export default function getXAndYFromEvent(event?: EventType) {
+  if (!event) return { ...DEFAULT_POINT };
```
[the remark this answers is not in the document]
williaster: yeah this is tricky. there are many instances across packages now that use whatever is returned by this function and assume numeric/valid `x/y` values exist.

that's why I went down the path of returning numeric values, but I agree that it doesn't seem right/is misleading, given an empty event. I like the idea of returning `null`s to signify invalid coordinates, it would still allow for type safety. I think this will require a change in `@vx/point` to support `x: number | null`, tho. wdyt?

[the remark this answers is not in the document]
williaster: I think no matter what, this means we have to update calls elsewhere to check for valid points. In that case I wonder if we could simply return `null` or `undefined` here instead of wrapping in a `Point` with `null` coordinates 🤔

williaster:
@hshoff I updated this to return `null` for non-event / invalid `inputs` as discussed offline.

PR #495, opened 2019-09-17
typescript(vx-gradient): re-write package in TypeScript

### LINKED ISSUE #494: Re-write `@vx/gradient` in TypeScript

williaster:
#### 🚀 Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/gradient` package to `TypeScript`. I also removed the `classnames` dep as it's not referenced anywhere.

The only new pattern introduced here is how we handle `{...restProps}`. As discussed offline we'll try to union `Types` for the element onto which we spread them to avoid breaking changes (the alternative was defining a new prop to put them all like `divProps: ...`, which would be breaking).

Closes #494.

**Testing**
- [x] CI
- [x]  Generates .d.ts files
  - [x] noting that `React.SVGProps` references don't get converted to `prop-types`. will avoid these so we can preserve `prop-types`
- [x] tested functionally with `/gallery`

image


packages/vx-gradient/src/gradients/LinearGradient.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+
+type LinearGradientElementProps = React.SVGProps<SVGLinearGradientElement>;
+type StopElementProps = React.SVGProps<SVGStopElement>;
+
+export type LinearGradientProps = {
+  /** Unique id for the gradient. Should be unique across all page elements. */
+  id: string;
+  /** Start color of gradient. */
+  from?: StopElementProps['stopColor'];
```
williaster: @milesj I noticed that none of these get converted to `prop-types`, is there any workaround for that? 

Most of these are just strings/numbers so I could remove the reliance on `React.SVGProps`

packages/vx-gradient/src/gradients/GradientOrangeRed.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import LinearGradient, { LinearGradientProps } from './LinearGradient';
+
+/**
+ * All props pass through to `<LinearGradient {...props} />`
+ */
+export default function GradientOrangeRed({
```
kristw: I am curious if we need one class per color scheme?

`GradientOrangeRed`, `GradientPinkBlue`, ...

Could we do 

```
<LinearGradient scheme="OrangeRed" />
```

or this if prefer to separate the class?

```
<PresetLinearGradient preset="OrangeRed" />
```

packages/vx-gradient/src/gradients/RadialGradient.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+
+type RadialGradientElementProps = React.SVGProps<SVGRadialGradientElement>;
+type StopElementProps = React.SVGProps<SVGStopElement>;
+
+export type RadialGradientProps = {
```
kristw: Could also pick from `LinearGradientProps` to avoid writing the field documentation twice.

```ts
export type RadialGradientProps = Pick<
  LinearGradientProps,
  | 'id'
  | 'from'
  | 'to'
  | 'fromOffset'
  | 'fromOpacity'
  | 'toOffset'
  | 'toOpacity'
  | 'rotate'
  | 'transform'
  | 'children'
> &
  RadialGradientElementProps;
```

packages/vx-gradient/src/gradients/LinearGradient.tsx
milesj: It doesn't generate prop types because these require TS type information. The Babel plugin only uses the raw source for information.

packages/vx-gradient/src/gradients/GradientOrangeRed.tsx
williaster: I agree that we don't need a class per scheme, I like the `<LinearGradient scheme="OrangeRed" />` proposal.

That said, I'd like to avoid major re-factors or breaking changes along with the TypeScript refactor so would like to punt on that for now.

PR #499, opened 2019-09-18
typescript(vx-drag): re-write package in TypeScript

### LINKED ISSUE #498: Re-write `@vx/drag` in TypeScript

williaster:
#### 🚀 Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/drag` package to `TypeScript`. 

Depends on #493 (for `@vx/event` types), and closes #498.

#### :boom: Breaking Changes
- Empty parent `<g>` wrapper around `Drag` `children` was replaced with a `React.Fragment` which removes a DOM element.

**Testing**
- [x] CI (depends on #493 for types)
- [x]  Generates .d.ts files
- [x] tested functionally with `/drag-i` and `/drag-ii`


packages/vx-drag/src/Drag.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const nextState = {
+      ...this.state,
+      isDragging: true,
+      dx: resetOnStart ? 0 : dx,
+      dy: resetOnStart ? 0 : dy,
+      x: resetOnStart ? point.x : -dx + point.x,
+      y: resetOnStart ? point.y : -dy + point.y,
+    };
+    if (onDragStart) onDragStart({ ...nextState, event });
+    this.setState(() => nextState);
```
kristw: Can this be `this.setState(nextState);`?

packages/vx-drag/src/util/raise.ts
```diff
@@ -0,0 +1,8 @@
+/** Given at an array of items, moves the item at the specified index to the end of the array. */
+export default function raise(items: unknown[], raiseIndex: number) {
```
kristw: Can preserve type with generic

```ts
export default function<T> raise(items: T[], raiseIndex: number) : T[] {
  // ...
}
```

or 

```ts
export default function<T> raise(items: T[], raiseIndex: number) {
  const array: T[] = [...items];
  // ...
}
```

packages/vx-drag/src/Drag.tsx
```diff
@@ -0,0 +1,121 @@
...
+    dy: 0,
+    isDragging: false,
+  };
+
+  handleDragStart = (event: React.MouseEvent) => {
+    const { onDragStart, resetOnStart } = this.props;
+    const { dx, dy } = this.state;
+    const point = localPoint(event);
+    const nextState = {
+      ...this.state,
```
kristw: This seems to override every field in `this.state` already. Is `...this.state` needed?

packages/vx-drag/src/Drag.tsx
williaster: Yeah good question, I'm not sure there's a good reason to use the functional variant here since we are not referencing any existing values.

packages/vx-drag/src/Drag.tsx
williaster: at first I was wondering the same, I think the goal is to provide a complete `state` to the callbacks tho.

packages/vx-drag/src/util/raise.ts
williaster: good call, wdyt about `function raise<T=unknown>(items: T[], ...)`?

packages/vx-drag/src/Drag.tsx
williaster: I refactored this to use functional `setState` with a callback for `onDrag*`. This should use freshest `state` and throttle `onDrag*` calls in sync with `setState` updates. Have to invoke `event.persist()`

williaster:
omg finally lol

packages/vx-drag/src/Drag.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  state = {
+    x: undefined,
+    y: undefined,
+    dx: 0,
+    dy: 0,
+    isDragging: false,
+  };
+
+  handleDragStart = (event: React.MouseEvent) => {
```
hshoff: these could also be TouchEvent

williaster: good catch, did a sweep 🧹

williaster:
Oi, type errors for vx-responsive + vx-text 🤔

williaster:
I added types to the `lodash` usage, not sure why they were failing all of a sudden 😢

hshoff:
RE: lodash dep https://github.com/hshoff/vx/pull/66

williaster:
ahhh, ic. reverted and just added `@types/lodash` to `vx-text` + `vx-responsive`. 

I think this is also failing the build in #512 .

PR #503, opened 2019-09-18
typescript(vx-pattern): re-write package in TypeScript

### LINKED ISSUE #502: Re-write `@vx/pattern` in TypeScript

williaster:
#### 🚀 Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/pattern` package to `TypeScript`.

Closes #502.

**Testing**
- [x] CI
- [x]  Generates .d.ts files
- [x] tested functionally with `/patterns`

#### :boom: Breaking Changes

- `PatternOrientation` is no longer the `default` export of `@vx/patterns/lib/constants` and is instead a named export. `PatternOrientation` is used as the export name if importing from the index: `import { PatternOrientation } from '@vx/pattern'`


packages/vx-pattern/src/constants/index.ts
```diff
@@ -0,0 +1,7 @@
+export type PatternOrientationType = 'horizontal' | 'vertical' | 'diagonal';
+
+export const PatternOrientation: { [key: string]: PatternOrientationType } = {
```
kristw: `[key: string]` => `[key in PatternOrientationType]`

packages/vx-pattern/src/patterns/Circles.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /** Width of the pattern element. */
+  width: number;
+  /** Height of the pattern element. */
+  height: number;
+  /** Radius of the pattern circles. */
+  radius?: number;
+  /** Fill applied to circles. */
+  fill?: string;
+  /** className applied to circles. */
+  className?: string;
```
kristw: How are the props ordered by?

packages/vx-pattern/src/patterns/Lines.tsx
```diff
@@ -1,84 +1,100 @@
...
+function pathForOrientation({
+  height,
+  orientation,
+}: {
+  height: number;
+  orientation: PatternOrientationType;
+}) {
+  let path;
+
+  switch (orientation) {
```
kristw: Maybe we can return within `switch` without defining variable `path`.

williaster: sgtm, I didn't add this logic but like the simplification.

packages/vx-pattern/src/patterns/Circles.tsx
williaster: there is not enforced ordering, I enabled several new rules in the `nimbus` upgrade but decided that was too big an undertaking / can't auto-fix.

packages/vx-pattern/src/constants/index.ts
```diff
@@ -0,0 +1,7 @@
+export type PatternOrientationType = 'horizontal' | 'vertical' | 'diagonal';
+
+export const PatternOrientation: { [key in PatternOrientationType]: PatternOrientationType } = {
```
kristw: change
`{ [key in PatternOrientationType]: PatternOrientationType }`
to
`{ [key in PatternOrientationType]: key }`

to enforce that `key === value`

packages/vx-pattern/src/patterns/Circles.tsx
```diff
@@ -0,0 +1,76 @@
...
+  radius = 2,
+  fill,
+  stroke,
+  strokeWidth,
+  strokeDasharray,
+  background,
+  complement = false,
+  className,
+}: PatternCirclesProps) {
+  let corners: number[][] | undefined;
```
kristw: can change `number[][]` to `[number, number][]` to be more explicit that there are always two numbers.

packages/vx-pattern/src/constants/index.ts
williaster: nice 👍 will update

packages/vx-pattern/src/patterns/Circles.tsx
williaster: some of the `d3` types are strict about this and others are not. will update 👍

packages/vx-pattern/test/PatternWaves.test.jsx
```diff
@@ -1,38 +0,0 @@
-import React from 'react';
-import { mount } from 'enzyme';
-
-import { PatternWaves } from '../src';
-
-describe('<PatternWaves />', () => {
-  beforeEach(() => {
-    global.console.error = jest.fn();
```
williaster: highlighting that `nimbus` mocks out `console.error` to throw, so we don't have to do this ourselves anymore.

PR #504, opened 2019-09-18
typescript(vx-tooltip): re-write package in TypeScript

### LINKED ISSUE #505: Re-write `@vx/tooltip` in TypeScript

Rudeg:
#### :rocket: Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the @vx-tooltip package to TypeScript.

`tooltip` component depends on `vx/bounds`, so CI might be failing because of this. Closes #505.

#### Testing
- [x] CI
- [x] Generates .d.ts files


packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,40 @@
...
+};
+
+type UpdateTooltipArgs = {
+  tooltipOpen?: Props['tooltipOpen'];
+  tooltipLeft?: Props['tooltipLeft'];
+  tooltipTop?: Props['tooltipTop'];
+  tooltipData?: Props['tooltipData'];
 };
 
 export default function withTooltip(
```
williaster: I think the `props` aren't quite right for this. For this to be compatible with any `BaseComponent` there will need to be a `Generic` here for the `BaseComponent`s own props. Something like this:

```jsx
function withTooltip<Props extends object = {}>(
  BaseComponent: (React.Component | React.PureComponent | React.FC)<ProvidedProps & Props>,
  containerProps: WithTooltipContainerProps = ...,
) {
  class WrappedComponent extends React.PureComponent<Props, WithTooltipState> {
     ...
  }
}
```

@milesj does this look right?

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,40 @@
...
-      height: 'inherit',
+      position: 'relative' as const,
+      width: 'inherit' as const,
+      height: 'inherit' as const,
     },
   },
 ) {
   class WrappedComponent extends React.PureComponent {
-    constructor(props) {
+    constructor(props: Props) {
```
williaster: Re comment above I think this will become `BaseProps`

packages/vx-tooltip/src/tooltips/TooltipWithBounds.tsx
```diff
@@ -1,44 +1,40 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import { withBoundingRects } from '@vx/bounds';
```
williaster: may need a `@ts-ignore` here as you mentioned because `@vx/bounds` currently has no types.

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,40 @@
...
 
-export const withTooltipPropTypes = {
-  tooltipOpen: PropTypes.bool,
-  tooltipLeft: PropTypes.number,
-  tooltipTop: PropTypes.number,
-  tooltipData: PropTypes.object,
-  updateTooltip: PropTypes.func,
-  showTooltip: PropTypes.func,
-  hideTooltip: PropTypes.func,
+type Props = {
```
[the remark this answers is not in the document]
Rudeg: the types of the state would be the same as part of the `WithTooltipProvidedProps`, so I defined the state types by using it. If you think it would be better to have it separately - I'll fix it :)

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,36 @@
...
+  updateTooltip?: (args: UpdateTooltipArgs) => void;
+  showTooltip?: (args: ShowTooltipArgs) => void;
+  hideTooltip?: () => void;
 };
+type WithTooltipState = Pick<
+  WithTooltipProvidedProps,
+  'tooltipOpen' | 'tooltipLeft' | 'tooltipTop' | 'tooltipData'
+>;
+type ShowTooltipArgs = Pick<WithTooltipProvidedProps, 'tooltipLeft' | 'tooltipTop' | 'tooltipData'>;
+type UpdateTooltipArgs = Pick<
```
kristw: same type?

```ts
type UpdateTooltipArgs = WithTooltipState;
```

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,36 @@
...
+  tooltipData?: Object;
+  updateTooltip?: (args: UpdateTooltipArgs) => void;
+  showTooltip?: (args: ShowTooltipArgs) => void;
+  hideTooltip?: () => void;
 };
+type WithTooltipState = Pick<
+  WithTooltipProvidedProps,
+  'tooltipOpen' | 'tooltipLeft' | 'tooltipTop' | 'tooltipData'
+>;
+type ShowTooltipArgs = Pick<WithTooltipProvidedProps, 'tooltipLeft' | 'tooltipTop' | 'tooltipData'>;
```
kristw:
```ts
type ShowTooltipArgs = Omit<WithTooltipState, 'tooltipOpen'>;
```

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,34 @@
...
+  'tooltipOpen' | 'tooltipLeft' | 'tooltipTop' | 'tooltipData'
+>;
+type ShowTooltipArgs = Omit<WithTooltipState, 'tooltipOpen'>;
+type UpdateTooltipArgs = WithTooltipState;
+type WithTooltipContainerProps = { style: React.CSSProperties };
 
-export default function withTooltip(
-  BaseComponent,
-  containerProps = {
+export default function withTooltip<Props extends object = {}>(
```
williaster: looks good, if the user specifies props on their component as `OwnProps & WithTooltipProvidedProps`, this should have everything on it 👍

packages/vx-tooltip/src/tooltips/Tooltip.tsx
```diff
@@ -1,16 +1,22 @@
...
 
-export default function Tooltip({ className, top, left, style, children, ...restProps }) {
+export default function Tooltip({
+  className,
+  top,
+  left,
+  style,
+  children,
+  ...restProps
+}: TooltipProps) {
```
williaster: missed this one earlier, but we'd like to preserve the ability to pass additional `restProps` here so we don't make breaking changes. the pattern I think we've aligned on is updating this line to `TooltipProps & JSX.IntrinsicElements.div` so that `TooltipProps` can still be imported separately but the component accepts any `div` props as well.

Rudeg: oh, you are right! It's a great point, thanks!

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -1,28 +1,34 @@
...
+  BaseComponent: React.ComponentType<Props>,
+  containerProps: WithTooltipContainerProps = {
     style: {
-      position: 'relative',
-      width: 'inherit',
-      height: 'inherit',
+      position: 'relative' as const,
+      width: 'inherit' as const,
+      height: 'inherit' as const,
     },
```
kristw: could do 
```ts
style: {
  position: 'relative',
  width: 'inherit',
  height: 'inherit'
} as const

williaster:
@Rudeg FYI I updated the base branch to `master`, and fixed a conflict.

PR #509, opened 2019-09-23
typescript(vx-bounds): re-write package in TypeScript

### LINKED ISSUE #508: Re-write `@vx/bounds` in TypeScript

Rudeg:
#### :rocket: Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the `@vx/bounds` package to TypeScript.

Closes #508.

#### Testing
- [x] CI
- [x] Generates .d.ts files


packages/vx-bounds/src/enhancers/withBoundingRects.tsx
```diff
@@ -39,15 +42,15 @@ export default function withBoundingRects(BaseComponent) {
     }
 
     componentDidMount() {
-      this.node = ReactDOM.findDOMNode(this);
+      this.node = ReactDOM.findDOMNode(this) as HTMLElement;
```
kristw: Should we use `ref` instead of doing `ReactDOM.findDOMNode()`?

williaster: there's been some discussion in the past about this. agree `findDOMNode` is gross 🤮 but it would be a breaking change unless we add an extra div wrapper, and that would also possibly break things for consumers. I think prob best to make that change in another PR.

williaster:
@Rudeg FYI I updated the base branch to master. I think this is good to go 👍

PR #512, opened 2019-09-23
typescript(vx-voronoi): re-write package in TypeScript

### LINKED ISSUE #511: Re-write `@vx/voronoi` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/voronoi` package in TypeScript. 

It also removes an unneeded `@vx/group` dependency. Closes #511.

#### :boom: Breaking Changes

**Tests**
- [x] CI
- [x] functional `/voronoi` demo

@hshoff @schillerk @milesj @kristw @rudeg


packages/vx-voronoi/src/components/VoronoiPolygon.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import cx from 'classnames';
+
+export type VoronoiPolygonProps = {
+  /** Override render function which is provided polygon and generated path. */
+  children?: ({ path, polygon }: { path: string; polygon: number[][] }) => React.ReactNode;
+  /** className to apply to path element. */
+  className?: string;
+  /** Array of coordinate arrays for the polygon (e.g., [[x,y], [x1,y1], ...]), used to generate polygon path. */
+  polygon?: number[][];
```
kristw: `[number, number][]`

packages/vx-voronoi/src/components/VoronoiPolygon.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import cx from 'classnames';
+
+export type VoronoiPolygonProps = {
+  /** Override render function which is provided polygon and generated path. */
+  children?: ({ path, polygon }: { path: string; polygon: number[][] }) => React.ReactNode;
```
kristw: `[number, number][]`

packages/vx-voronoi/src/components/VoronoiPolygon.tsx
williaster: looks like this is not possible when the type is optional, build fails with [this error](https://github.com/Microsoft/TypeScript/issues/12707#issuecomment-384878950)

williaster: made it required to avoid

williaster: I take that back, I think it comes from the [`@types/d3-voronoi` types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3-voronoi/index.d.ts#L177)

williaster:
Hrmm, not really sure how to fix this one. I think this error is coming from the [`@types/d3-voronoi` types](https://travis-ci.org/hshoff/vx/jobs/593725114#L342), and still throws even tho I ignore the call in the code 🤔

hshoff:
Looks like the tests are failing on [`packages/vx-voronoi/test/voronoi.test.ts(24,11)`](https://github.com/hshoff/vx/blob/8502ad5f622ca77ecbb70b9072957c53d6a56c2b/packages/vx-voronoi/test/voronoi.test.ts#L24).

```js
const [, [x1, y1]] = v.extent();
```

Might try changing this to:

```js
const [[x0,y1], [x1, y1]] = v.extent();
```

williaster:
old school syntax worked :neckbeard:  🎉

PR #515, opened 2019-09-24
typescript(vx-network): re-write package in TypeScript

### LINKED ISSUE #513: Re-write `@vx/network` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/network` package in TypeScript. 

It also removes an unneeded `d3-force` dependency. Closes #513.

#### :boom: Breaking Changes
- `<Nodes />` 
  - inner node wrapper `<g>` element `className` changed to singular (`vx-network-nodes` => `vx-network-node`)
  - outer wrapper `<g>` was replaced with a `React.Fragment`
- `<Links />` 
  - inner link wrapper `<g>` element `className` changed to singular (`vx-network-links` => `vx-network-link`)
  - outer wrapper `<g>` was replaced with a `React.Fragment`

Tests
- [x] CI
- [x] functional `/network` demo

@hshoff @schillerk @milesj @kristw @Rudeg


packages/vx-network/src/DefaultLink.tsx
```diff
@@ -1,12 +1,8 @@
...
-import PropTypes from 'prop-types';
+import { LinkProvidedProps } from './types';
 
-DefaultLink.propTypes = {
-  link: PropTypes.object,
-};
-
-export default function DefaultLink({ link }) {
-  return (
+export default function DefaultLink({ link }: LinkProvidedProps<any>) {
```
Rudeg: can it be `LinkProvidedProps<Link<DefaultNode>>`?

williaster: I tried this at first, but got TS errors in the `Links` component when setting `linkComponent = DefautlLink` along the lines of `Link<DefaultNode>` is not assignable to `Link` 😭 

wasn't sure of a better solution than this 👎

packages/vx-network/src/Nodes.tsx
```diff
@@ -0,0 +1,45 @@
...
+  x?: (d: Node) => number;
+  /** Returns the center y coordinate of a node. */
+  y?: (d: Node) => number;
+};
+
+export default function Nodes<Node>({
+  nodes = [],
+  nodeComponent = DefaultNode,
+  className,
+  x = (d: any) => (d && d.x) || 0,
```
kristw: `x = (d:any) => ...` => `(d?: {x: number}) => ...`
`y = (d:any) => ...` => `(d?: {y: number}) => ...`

williaster: this is similar to the [`defaultLink` above](https://github.com/hshoff/vx/pull/515#discussion_r328753856), where I get the error `Type 'Node' is not assignable to type '{ y: number; }'` (even if we default the `Node` type to `{ x, y }`, it's not guaranteed)

PR #522, opened 2019-09-27
typescript(vx-zoom): re-write package in TypeScript

### LINKED ISSUE #521: Re-write `@vx/zoom` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/zoom` package in TypeScript. 

Closes #521 , depends on #493 for `@vx/event` types.

Tests
- [x] CI (depends on #493 for types)
- [x] functional `/zoom` demo
- [x] generates `.d.ts` files

@hshoff @schillerk @milesj @kristw @Rudeg


packages/vx-zoom/src/Zoom.tsx
```diff
@@ -83,119 +159,119 @@ class Zoom extends React.Component {
...
 
-  handleWheel(event) {
-    const { passive } = this.props;
+  handleWheel = (event: React.WheelEvent | WheelEvent) => {
+    const { passive, wheelDelta } = this.props;
     if (!passive) event.preventDefault();
-    const { wheelDelta } = this.props;
     const point = localPoint(event);
-    const { scaleX, scaleY } = wheelDelta(event);
+    const { scaleX, scaleY } = wheelDelta!(event);
```
Rudeg: it can be just `wheelDelta`, cause you have a default props

packages/vx-zoom/src/Zoom.tsx
```diff
@@ -10,36 +9,113 @@ import {
...
+
+  containerRef: HTMLDivElement | null = null;
+
+  startPoint: Point | undefined = undefined;
+
+  startTranslate: Translate | undefined = undefined;
+
+  state = {
+    initialTransformMatrix: this.props.transformMatrix!,
+    transformMatrix: this.props.transformMatrix!,
```
Rudeg: it can be just 
`
initialTransformMatrix: this.props.transformMatrix,
transformMatrix: this.props.transformMatrix,
`

williaster: this actually needs the `!` because we don't have any `defaultProps` helpers so it doesn't know they are defined.

packages/vx-zoom/src/Zoom.tsx
williaster: this actually needs the `!` because we don't have any defaultProps helpers so it doesn't know they are defined.

PR #520, opened 2019-09-25
typescript(vx-heatmap): re-write package in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/heatmap` package in TypeScript. 

Closes #519. 

#### :boom: Breaking Changes
- This PR introduces `React.Fragment`s, which requires bumping the `peerDep` for `react` to `^16.3.0-0`

Tests
- [x] CI
- [x] functional `/heatmaps` demo
- [x] generates `.d.ts` files

@hshoff @schillerk @milesj @kristw @Rudeg

PR #527, opened 2019-09-30
typescript(vx-chord): re-write package in TypeScript

### LINKED ISSUE #525: Re-write `@vx/chord` in TypeScript

Rudeg:
#### :rocket: Enhancements

This PR builds off of #488 which added TypeScript build support, and converts the @vx-chord package to TypeScript.

Fixes #525.

#### Testing
- [x] CI
- [x] Generates .d.ts files


packages/vx-chord/src/Ribbon.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ...restProps
+}: RibbonProps & Omit<React.SVGProps<SVGPathElement>, keyof RibbonProps>) {
+  const ribbon = d3ribbon<any, Chord, ChordSubgroup>();
+  if (source) ribbon.source(source);
+  if (target) ribbon.target(target);
+  if (radius) ribbon.radius(radius);
+  if (startAngle) ribbon.startAngle(startAngle);
+  if (endAngle) ribbon.endAngle(endAngle);
+  const path = ribbon(chord) as any;
+  if (children) return children({ path });
```
williaster: will do a proper review later today but I believe the build is failing because of this line (and a similar one in `Chord.tsx`). I've had issues in other packages where this doesn't seem to satisfy the `React.ReactNode` type, and was able to fix it by wrapping it in fragments ([as suggested in this SO answer](https://stackoverflow.com/a/54908762/1933266)):

```jsx
return <>{children({ path })}</>;
```

That change requires us to bump the `react` peer dep to `^16.3.0` which is a breaking change, but we're moving that direction anyway due to deprecated lifecycle methods.

packages/vx-chord/src/Ribbon.tsx
```diff
@@ -0,0 +1,37 @@
...
+  className,
+  ...restProps
+}: RibbonProps & Omit<React.SVGProps<SVGPathElement>, keyof RibbonProps>) {
+  const ribbon = d3ribbon<any, Chord, ChordSubgroup>();
+  if (source) ribbon.source(source);
+  if (target) ribbon.target(target);
+  if (radius) ribbon.radius(radius);
+  if (startAngle) ribbon.startAngle(startAngle);
+  if (endAngle) ribbon.endAngle(endAngle);
+  const path = ribbon(chord) as any;
```
williaster: can this not be any? based on other `d3` `path` generators, it should return a `string` or `null`? 

[the types seem to suggest this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3-chord/index.d.ts#L269) as well.

williaster:
not totally sure what's up with the build, my most recent [`@vx/hierarchy` PR](https://github.com/hshoff/vx/pull/524) was giving similar errors even though [the base branch build passed](https://github.com/hshoff/vx/pull/488), and other branches based off of it have also passed (e.g., [`@vx/heatmap` PR](https://github.com/hshoff/vx/pull/520)).

looks like the types for jest are now missing, perhaps something broke with a minor version change 🤔@milesj do you have any thoughts?

williaster:
fyi if you rebase on `chris--typescript` the build problem should be fixed

williaster:
Hmm this error will be a bit tricky 🤔 any thoughts on a good fix?

image

Also, updated the base branch to `master`

Rudeg:
@williaster I use `[]` instead of the fragments and it should work fine for this case

packages/vx-chord/src/Ribbon.tsx
Rudeg: TypeScript not inferring the correct method overload in this case :( Not sure what is the best way to fix this

williaster:
as discussed offline, can we remove all of the `React.Fragment` `key` logic and update the test to something like the following? seems like something weird with Enzyme, I don't think the fragments need a key based on the actual usage.

```tsx
import React from 'react';
import { shallow } from 'enzyme';
import { chord as d3Chord } from 'd3-chord';

import { Ribbon } from '../src';

const matrix = [
  [11975, 5871, 8916, 2868],
  [1951, 10048, 2060, 6171],
  [8010, 16145, 8090, 8045],
  [1013, 990, 940, 6907],
];

const chords = d3Chord()(matrix);

describe('<Ribbon />', () => {
    expect(Ribbon).toBeDefined();
  });

  test('it should call children as a function with required args', () => {
    const children = jest.fn(() => 'test');
    shallow(<Ribbon chord={chords[0]} children={children} />);
    const args = children.mock.calls[0][0];
    expect(children.mock.calls).toHaveLength(1);
    expect(args.path).toBeDefined();
  });
});

```

PR #528, opened 2019-10-01
[root] disable no-useless-rename

hshoff:
Should fix the broken build on master https://travis-ci.org/hshoff/vx/builds/591675984

#### :house: Internal

- [eslint] disable `no-useless-name`


williaster:
thanks for the fix 🙌

PR #507, opened 2019-09-20
typescript(vx-shape): Re-write package in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces `Typescript` build config, and re-writes the `@vx/shape` package in `TypeScript`. Unfortunately this package is one of the biggest ones, so this PR is pretty sizable and larger than the other re-writes. 

A couple of notes:
- I added `@types/d3-path` and `@types/d3-shape` dependencies so that we can leverage/propagate existing types in our d3 usage.
- I removed the `@vx/point` dependency .
- I tried consolidating some of the duplicate code for the 12 link components, this was trickier than I thought so mostly share prop `Type`s.

#### :boom: Breaking Changes
- This PR introduces `React.Fragment`s, which requires bumping the `peerDep` for `react` to `^16.3.0-0`
- `@vx/shape` the `Arc` `centroid` prop was removed as it was not functional (it was called as if it was an `arc.centroid()` configuration parameter, but in reality the `.centroid` method is for returning the centroid of a datum.
-  `@vx/shape` the `Area` component is no longer wrapped in an empty `<g>` element
- `propTypes` across all components are likely stricter.
- `children(...)` render function overrides are wrapped in `React.Fragment`s  `<>{children(...)</>` to satisfy Types.
- `order` and `offset` props for `Stack`, `BarStack`, `BarStackHorizontal`, and `AreaStack` previously supported strings, arrays, or functions. Only the `string` prop was functional, and only the enumerated `string` presets are now allowed.
 
Closes #506

#### Testing
- [x] CI
  - [x] typescript
  - [x] eslint
  - [x] jest
- [x] Functional with `/gallery`
- [x] Generates `.d.ts` files

@hshoff @schillerk @milesj @techniq @kristw

Issue #410, opened 2019-01-13
Multiple threshold graphs on the same page distort threshold fill


isaacgr:
When adding multiple threshold graphs on the same page, the second graph (if using seperate data from the first) will have a distorted threshold fill. See attached image for issue. Code can be found here [https://github.com/isaacgr/TemperatureWebpage/blob/master/src/Graph.js](url).

I kind of think its related to the react component values I'm setting, but not sure how to address it. The bottom graph should have green above 30, and yellow below. If I make both graphs use the same data, or get rid of the top graph, then the threshold is not distorted.

![image](https://user-images.githubusercontent.com/20640588/51082010-df2dbe80-16cb-11e9-9493-18cc4a54f9ec.png)

techniq:
If you add a unique 'id' prop to each graph, this should fix this

techniq:
See https://github.com/hshoff/vx/issues/341

PR #533, opened 2019-10-02
typescript(vx-threshold): re-write package in TypeScript

### LINKED ISSUE #530: Re-write `@vx/threshold` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/threshold` package in TypeScript. It also exposes the `defined` prop from the `@vx/shape` `Area` component.

This depends on types for `@vx/shape` (#507) and `@vx/clip-path` (#497). Closes #530.

#### :boom: Breaking Changes

- Updates the `id` prop for `Threshold` to be required. We've had user issues not understand that this is needed for multiple charts on a page (e.g., #410) and stronger typing seems like a good time to fix this. 

(Alternatively we could try to more intelligently generate these but it doesn't seem too bad for a user to pass this.)

#### :memo: Documentation

- Tried to clarify the props for this semi-confusing component via type annotations 🤞 
- Updates the `/threshold` demo to include a description and use red/green color blind friendly colors.

**Tests**
- [x] functional `/threshold` demo
- [x] generates `.d.ts` files
- [ ] CI (depends on types for`@vx/shape` (#507) and `@vx/clip-path` (#497))

@hshoff @techniq @schillerk @milesj @kristw @Rudeg


packages/vx-threshold/src/Threshold.tsx
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+import cx from 'classnames';
+import { Area } from '@vx/shape';
+import { AreaProps } from '@vx/shape/lib/shapes/Area';
+import { ClipPath } from '@vx/clip-path';
+
+type NumberAccessor<Datum> = (datum: Datum, index: number, data: Datum[]) => number;
+
+type ThresholdProps<Datum> = {
```
Rudeg: could we `export` this props?

williaster: 🤦‍♂ yep thanks for catching

PR #551, opened 2019-10-31
typescript(vx-legend): re-write package in TypeScript

### LINKED ISSUE #546: Re-write `@vx/legend` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces `Typescript` build config, and re-writes the `@vx/legend` package in TypeScript.

This one was a bit gnarly vs some. Closes #546.

#### 💥 Breaking Changes

- This PR introduces `React.Fragments`, which requires bumping the `peerDep` for react to `^16.3.0-0`
- The following changes were made to the package directory structure which will affect deep import paths of components (importing from the `index` is unchanged):
  - `src/legends/*` => `src/*`

#### Tests
- [x] CI
- [x] functional `/legend` demo 
- [x] generates `.d.ts` files

#### Screen shots

**Prod**
image

**Local**
image

@hshoff  @kristw


packages/vx-demo/src/components/tiles/legends.js
```diff
@@ -40,6 +40,8 @@ const thresholdScale = scaleThreshold({
   range: ['#f2f0f7', '#dadaeb', '#bcbddc', '#9e9ac8', '#756bb1', '#54278f'],
 });
 
+global.s = thresholdScale;
```
williaster: testing, will remove

packages/vx-legend/src/legends/Threshold.tsx
```diff
@@ -0,0 +1,85 @@
...
+function defaultTransform<Datum extends BaseInput, Output extends BaseOutput>({
+  labelDelimiter,
+  labelLower,
+  labelUpper,
+}: Pick<
+  LegendThresholdProps<Datum, Output>,
+  'labelDelimiter' | 'labelLower' | 'labelUpper'
+>): LabelFormatterFactory<Datum, Output, ScaleThreshold<Datum, Output>> {
+  return ({ scale, labelFormat }) => (d, i) => {
+    let [x0, x1] = scale.invertExtent(scale.range()[i]);
```
williaster: this is messy, the problem is that the previous implementation used `scale.range()` and passed _that_ as the `domain` prop. that is no longer valid with types, and it's impossible to compute the right thresholds directly from the domain value `d`.

the domain of a `threshold` scale also is not guaranteed to be a `number` so we have extra checks for that here.

packages/vx-legend/src/index.ts
```diff
@@ -4,6 +4,6 @@ export { default as LegendLinear } from './legends/Linear';
 export { default as LegendOrdinal } from './legends/Ordinal';
 export { default as LegendThreshold } from './legends/Threshold';
 export { default as LegendSize } from './legends/Size';
-export { default as LegendItem } from './legends/LegendItem';
-export { default as LegendLabel } from './legends/LegendLabel';
-export { default as LegendShape } from './legends/LegendShape';
+export { default as LegendItem } from './legends/Legend/LegendItem';
```
williaster: these directory changes are technically **breaking** if consumers use deep imports, will update PR

packages/vx-legend/src/legends/Legend/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+  domain?: Datum[];
+  /** Width of the legend shape. */
+  shapeWidth?: string | number;
+  /** Height of the legend shape. */
+  shapeHeight?: string | number;
+  /** Margin of the legend shape. */
+  shapeMargin?: string | number;
+  /** Flex-box alignment of legend item labels. */
+  labelAlign?: string;
+  /** @TODO handle object type? */
```
williaster: `@TODO`

packages/vx-legend/src/types/index.ts
```diff
@@ -0,0 +1,71 @@
...
+export type BaseInput = string | number | Date;
+export type BaseOutput = string | number | Date;
+
+export type ScaleLinear<Input, Output> = d3Scale.ScaleLinear<Input, Output>;
+export type ScaleOrdinal<Input, Output> = d3Scale.ScaleOrdinal<Input, Output>;
+export type ScaleBand<Input> = d3Scale.ScaleBand<Input>;
+export type ScaleThreshold<Input extends BaseInput, Output> = d3Scale.ScaleThreshold<Input, Output>;
+export type ScaleQuantile<Output> = d3Scale.ScaleQuantile<Output>;
+
+// @TODO BaseInput only needed for `ScaleThreshold`
```
williaster: I'm also not sure how to get around this cleanly. only `ScaleThreshold` has restrictions on `Input`, but that vastly restricts uses of the others. @kristw do you have any advance `TS` ideas?

packages/vx-legend/src/legends/Quantile.tsx
```diff
@@ -0,0 +1,54 @@
...
+export default function LegendQuantile<Output extends BaseOutput>({
+  domain: inputDomain,
+  scale,
+  labelFormat = x => x,
+  labelTransform: inputLabelTransform,
+  labelDelimiter = '-',
+  ...restProps
+}: LegendQuantileProps<Output>) {
+  // transform range into input values because it may contain more elements
+  const domain = inputDomain || scale.range().map(output => scale.invertExtent(output)[0]);
```
williaster: this similarly just passed `scale.range()` as `domain` previously, this works correctly though. the note about the `range` I don't think is necessarily _always_ true.

packages/vx-legend/src/legends/Legend/LegendItem.tsx
```diff
@@ -1,12 +1,19 @@
...
-import PropTypes from 'prop-types';
 
-LegendItem.propTypes = {
-  flexDirection: PropTypes.string,
-  alignItems: PropTypes.string,
-  margin: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-  children: PropTypes.any,
-  display: PropTypes.string,
+export type LegendItemProps = {
+  flexDirection?:
```
williaster: can update to `FlexDirection`

packages/vx-legend/src/types/index.ts
williaster: might be able to use conditional types here

```ts
type ScaleType<Input, Output> = Input extends BaseInput
  ? ScaleThreshold<Input, Output> | NonThresholdType<Input, Output>
  : NonThresholdType<Input, Output>;
```

packages/vx-legend/src/legends/Threshold.tsx
williaster: after going through the d3 docs more, if `domain` has `n` items then the `range` of a valid scale should have `n+1`, so I think this is safe. I updated the logic to check that the item corresponding to index `i` is within the range, and added a comment. we can update in the future if need be.

packages/vx-legend/src/types/index.ts
williaster: I spent a lot of time on this and couldn't get the conditional typing to work 😱 I am TS-ignoring it for now and will sync with @kristw  and others when they're back from holiday.

williaster:
woot! finally 😅

PR #558, opened 2019-11-26
Fix debounced calls after unmount in responsive components

claisne:
#### :bug: Bug Fix

Debounce calls can still fire after unmount, so we need to cancel them.

### DIFF
--- a/packages/vx-responsive/src/components/ParentSize.tsx
+++ b/packages/vx-responsive/src/components/ParentSize.tsx
@@ -62,6 +62,7 @@ export default class ParentSize extends React.Component<
   componentWillUnmount() {
     if (this.animationFrameID) window.cancelAnimationFrame(this.animationFrameID);
     if (this.resizeObserver) this.resizeObserver.disconnect();
+    if (this.resize) this.resize.cancel();
   }
 
   resize = ({ width, height, top, left }: ParentSizeState) => {
--- a/packages/vx-responsive/src/components/ParentSize.tsx
+++ b/packages/vx-responsive/src/components/ParentSize.tsx
@@ -62,7 +62,7 @@ export default class ParentSize extends React.Component<
   componentWillUnmount() {
     if (this.animationFrameID) window.cancelAnimationFrame(this.animationFrameID);
     if (this.resizeObserver) this.resizeObserver.disconnect();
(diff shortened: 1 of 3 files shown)


packages/vx-responsive/src/enhancers/withScreenSize.tsx
```diff
@@ -37,6 +37,7 @@ export default function withScreenSize<Props extends WithScreenSizeProps = {}>(
 
     componentWillUnmount() {
       window.removeEventListener('resize', this.handleResize, false);
+      this.resize.cancel();
```
williaster: ditto re updating this to `this.handleResize.cancel()` since the unwrapped method wouldn't have the `.cancel()` method

packages/vx-responsive/src/enhancers/withParentSize.tsx
```diff
@@ -53,6 +53,7 @@ export default function withParentSize<Props extends WithParentSizeProps = {}>(
     componentWillUnmount() {
       if (this.animationFrameID) window.cancelAnimationFrame(this.animationFrameID);
       if (this.resizeObserver) this.resizeObserver.disconnect();
+      if (this.resize) this.resize.cancel();
```
williaster: ditto re removing the `if (...)` statement since it's defined in the constructor.

also I think these need to be updated to `this.debouncedResize`/`this.debouncedResize.cancel()` because the unwrapped method wouldn't have the `.cancel()` method (nice that we name this differently in every component 😱 )

packages/vx-responsive/src/components/ParentSize.tsx
```diff
@@ -62,6 +62,7 @@ export default class ParentSize extends React.Component<
   componentWillUnmount() {
     if (this.animationFrameID) window.cancelAnimationFrame(this.animationFrameID);
     if (this.resizeObserver) this.resizeObserver.disconnect();
+    if (this.resize) this.resize.cancel();
```
williaster: Is the `if (...)` statement needed here? since the debounced method is defined in the constructor I think this check is unneeded (versus `animationFrameID`/`resizeObserver` which can be `null` or `undefined`)

williaster:
@claisne thanks for making the updates, I think if you rebase or merge the latest master the build should be fixed.

PR #559, opened 2019-12-03
deps: upgrade nimbus + configs

williaster:
### :house: Internal

This bumps the `@airbnb/nimbus` + `@airbnb/config-*` packages to the latest to fix an error with the build (seen in #553 and #558). The [`nimbus` PR](https://github.com/airbnb/nimbus/pull/19) added a filter for non-babel CLI args because a patch change in `babel@7.7` (which bumped a CLI dependency) resulted in babel starting to throw for unrecognized arguments (like `--clean` or `--esm`).

Noting that this might not have happened / we wouldn't be as susceptible to breaking patch changes  if we commited the root `yarn.lock` file. any thoughts on that? guess it's still a risk for dependencies in packages by lib consumers which was the concern originally.

@hshoff 
cc @geekplux @claisne


geekplux:
Efficient action

PR #565, opened 2019-12-19
typescript(vx-grid): re-write package in TypeScript

### LINKED ISSUE #563: Re-write `@vx/grid` in TypeScript

williaster:
#### :rocket: Enhancements

This PR builds off #488 which introduces Typescript build config, and re-writes the `@vx/grid` package in TypeScript. 

Closes #563.

**Tests**
- [x] CI
- [x] functional via `/axis` demos
- [x] generates `.d.ts` files

_Note_: I did not update the demo to this, just testing functionality

cc @hshoff @schillerk @kristw @Rudeg


packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -34,9 +29,8 @@ export default function Columns({
   offset,
   tickValues,
   ...restProps
-}) {
-  let ticks = scale.ticks ? scale.ticks(numTicks) : scale.domain();
-  if (tickValues) ticks = tickValues;
+}: AllProps<ScaleInput>) {
+  const ticks = tickValues || scale.ticks ? scale.ticks(numTicks) : scale.domain();
```
schillerk: I don't have a ton of context, but seems like you're changing the logic here and again further down? Does the TS migration mean we never want ticks = tickValues? It is safe to call `scale.ticks` if we're not sure it exists?

williaster: I didn't really change the logic, just made it an immutable variable and simplified the expression. 

Same as before, if `tickValues` is set, it overrides everything. Otherwise we use `.ticks()` if it's defined and fallback to `.domain()` which is always defined.

packages/vx-grid/src/types.ts
```diff
@@ -0,0 +1,31 @@
+export interface Scale<Input = any, Output = any> {
+  (value: Input): Output;
+  ticks(count: number): Input[];
```
williaster: this type is wrong, will update to `ticks?: (count: number) => Input[];`

packages/vx-grid/src/grids/GridColumns.tsx
williaster: you were right, messed up the logic! fixed with some `()` 🙏

PR #567, opened 2019-12-20
fix(vx-brush): Fix BaseBrush, iterate on types + gallery

williaster:
#### :rocket: Enhancements

- Adds a bit of polish to the gallery brush example (changes the big:small chart ratio, changes colors from green to purple, uses our `LinearGradient` component instead of a manual `defs` version, hides bottom axes in smaller tile gallery vs page demo)
- Updates the brush demo to include the react code (it's not super obvious that you have to do this)

image

#### :bug: Bug Fix

- Previously the `shouldComponentUpdate` method in `BaseBrush` was written to almost always return `false`. This prevented the `BrushSelection` region to actually render and change dimensions as you drag making the `Brush` component unusable.
- While debugging I iterated on several of the types within the `vx-brush` package
  - there were lots of `Function` and `any`s which I tried to get rid of for the most part
  - I tried to leverage the `@vx/drag` types where possible

#### Testing

- [x] functional brush demo
- [x] CI
  - [x] I need to verify that the Drag types actually work

@hshoff @geekplux 
cc @schillerk @kristw @Rudeg

### DIFF
--- a/packages/vx-brush/src/BaseBrush.tsx
+++ b/packages/vx-brush/src/BaseBrush.tsx
@@ -122,8 +122,8 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     const marginLeft = inheritedMargin && inheritedMargin.left ? inheritedMargin.left : 0;
     const marginTop = inheritedMargin && inheritedMargin.top ? inheritedMargin.top : 0;
     const start = {
-      x: draw.x + draw.dx - left - marginLeft,
-      y: draw.y + draw.dy - top - marginTop,
+      x: (draw.x || 0) + draw.dx - left - marginLeft,
+      y: (draw.y || 0) + draw.dy - top - marginTop,
     };
     const end = { ...start };
 
@@ -148,11 +148,11 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
   handleDragMove = (draw: DragArgs) => {
     const { left, top, inheritedMargin } = this.props;
     if (!draw.isDragging) return;
-    const marginLeft = inheritedMargin && inheritedMargin.left ? inheritedMargin.left : 0;
-    const marginTop = inheritedMargin && inheritedMargin.top ? inheritedMargin.top : 0;
+    const marginLeft = (inheritedMargin && inheritedMargin.left) || 0;
+    const marginTop = (inheritedMargin && inheritedMargin.top) || 0;
     const end = {
-      x: draw.x + draw.dx - left - marginLeft,
-      y: draw.y + draw.dy - top - marginTop,
+      x: (draw.x || 0) + draw.dx - left - marginLeft,
+      y: (draw.y || 0) + draw.dy - top - marginTop,
     };
     this.updateBrush((prevBrush: BaseBrushState) => {
       const { start } = prevBrush;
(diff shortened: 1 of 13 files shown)


packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -347,34 +363,34 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
               onDoubleClick={() => this.reset()}
-              onClick={(event: React.MouseEvent<SVGRectElement, MouseEvent>) => {
+              onClick={(event: MouseHandlerEvent) => {
                 const duration = this.mouseUpTime - this.mouseDownTime;
                 if (onClick && duration < clickSensitivity) onClick(event);
               }}
-              onMouseDown={(event: React.MouseEvent<SVGRectElement, MouseEvent>) => {
-                this.mouseDownTime = new Date();
+              onMouseDown={(event: MouseHandlerEvent) => {
+                this.mouseDownTime = Date.now();
```
williaster: needs to be a number

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -280,16 +297,16 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
-      end: undefined,
+    this.updateBrush(() => ({
+      start: { x: 0, y: 0 },
+      end: { x: 0, y: 0 },
       extent: {
-        x0: undefined,
-        x1: undefined,
-        y0: undefined,
-        y1: undefined,
+        x0: 0,
```
williaster: these are the values the brush is initialized with, we might reconsider this to be `-1` in the future. non-numbers would have to be handled in a lot of places.

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -386,55 +402,53 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
-                />
+                corner && (
+                  <BrushCorner
+                    key={`corner-${cornerKey}`}
+                    type={cornerKey}
+                    brush={this.state}
+                    updateBrush={this.updateBrush}
+                    stageWidth={stageWidth}
+                    stageHeight={stageHeight}
+                    corner={corner}
```
williaster: I think corners previously did not work. I moved the coords into an object like `handle` does.

packages/vx-brush/src/Brush.tsx
```diff
@@ -29,8 +29,6 @@ export type BrushProps = {
 };
 
 class Brush extends React.Component<BrushProps> {
-  private BaseBrush: any = React.createRef();
```
williaster: this was not used and `BaseBrush` doesn't implement it so I removed it.

packages/vx-brush/src/BrushCorner.tsx
```diff
@@ -167,20 +151,19 @@ export default class BrushCorner extends React.Component<BrushCornerProps, Brush
                 fill="transparent"
                 width={stageWidth}
                 height={stageHeight}
-                style={{ cursor: style.cursor }}
+                style={{ cursor }}
```
williaster: need to update this to use `styleProp.cursor` if defined.

packages/vx-brush/src/BrushCorner.tsx
```diff
@@ -167,20 +151,19 @@ export default class BrushCorner extends React.Component<BrushCornerProps, Brush
...
                 onMouseMove={dragMove}
                 onMouseUp={dragEnd}
               />
             )}
             <rect
-              fill="transparent"
+              fill="violet"
```
williaster: testing, will remove

packages/vx-brush/src/BrushHandle.tsx
```diff
@@ -143,7 +139,7 @@ export default class BrushHandle extends React.Component<BrushHandleProps> {
               y={y}
               width={width}
               height={height}
-              fill="transparent"
+              fill="yellow"
```
williaster: testing will remove

williaster:
gonna land this since it fixes so much. can continue to iterate.

PR #574, opened 2020-01-02
Update Chord Diagram example to make more sense

jens-ox:
This PR updates the chord diagram.

1. The linear gradients are replaced by static colors to better show what is going on.
2. The coloring of the ribbons is fixed. Previously, the coloring function was used identically for the arcs and ribbons (using the index), but that doesn't work for ribbons, as the index of a ribbon is not related to the index of the arcs (resulting in a ribbon coloring that doesn't make sense). Now, the index of the target group is used to color the ribbons.
3. The code example of the chord graph is updated accordingly.

**Old**:
Notice how the coloring of the ribbons with identical source and target group is different than the coloring of that group. I suppose that this wasn't noticed because of the gradients :smile:
Selection_003

**New**:
Selection_004

### DIFF
--- a/packages/vx-demo/src/pages/chord.js
+++ b/packages/vx-demo/src/pages/chord.js
@@ -77,7 +77,7 @@ export default ({ width, height, centerSize = 20, events = false }) => {
                         fill={color(chord.target.index)}
                         fillOpacity={0.75}
                         onClick={() => {
-                          alert(\'\${JSON.stringify(chord)}\`);
+                          alert(\`\${JSON.stringify(chord)}\`);
                         }}
                       />
                     );
--- a/packages/vx-demo/src/pages/chord.js
+++ b/packages/vx-demo/src/pages/chord.js
@@ -19,8 +19,16 @@ import { Arc } from '@vx/shape';
 import { Group } from '@vx/group';
 import { Chord, Ribbon } from '@vx/chord';
 import { scaleOrdinal } from '@vx/scale';
-import { schemeDark2 } from 'd3-scale-chromatic';
+import { LinearGradient } from '@vx/gradient';
 
+const pink = '#ff2fab';
+const orange = '#ffc62e';
+const purple = '#dc04ff';
+const purple2 = '#7324ff';
+const red = '#d04376';
+const green = '#52f091';
+const blue = '#04a6ff';
+const lime = '#00ddc6';
 const bg = '#e4e3d8';
 
 const matrix = [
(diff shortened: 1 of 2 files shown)


jens-ox:
The color scale is dark2 from d3-scale-chromatic which is already used in vx-demo.

williaster:
Hey @jens-ox, thanks for checking out `vx` and making this demo more interpretable! 🙏 I agree that the color should deterministically come from the target (or source), not the arc index. 

My only concern is over the new color palette. We try to make the gallery quite vibrant and visually intriguing, and the dark d3 color palette feels less engaging to me. We also often create our own color palettes simply to promote diversity in color beyond what d3 provides, as d3 colors are used everywhere and therefore feel less unique. 

Would you be open to leveraging the previous colors? I would also love to keep the gradients because they are more visually interesting, I think we could make them interpretable by toning back the variation within each gradient (i.e., make the colors within a single gradient closer to each other, like same hue with varied brightness). I pasted some possible variations below but feel free to tweak them!

image

williaster:
also not sure what's up with the coverage check, don't worry about that.

jens-ox:
Thanks for getting back! I agree, the gradients were much more vibrant than dark2.

I tried the colors you proposed. As there are only 4 groups in the chart, I had to leave one color out:

Without orange           |  Without blue
:-------------------------:|:-------------------------:
Selection_007  |  Selection_006

In my opinion, this might be a little bit too colorful. I think the "old" gradients look better:
Selection_005

I switched the PR back to the old gradients. Also, the domain of the color scale was erroneous, fixed that.

packages/vx-demo/src/pages/chord.js
```diff
@@ -19,23 +19,15 @@ import { Arc } from '@vx/shape';
 import { Group } from '@vx/group';
 import { Chord, Ribbon } from '@vx/chord';
 import { scaleOrdinal } from '@vx/scale';
-import { LinearGradient } from '@vx/gradient';
+import { schemeDark2 } from 'd3-scale-chromatic';
```
williaster: could we remove the changes in the page source code as well? 

sorry this is more work than it should be, we're scheming ways to improve docs + examples in the near future after TS wraps up.

jens-ox: Sorry, forgot about that one. Updated the page source code.

williaster:
Thanks @jens-ox! Heads up that Harry is out for a few weeks and I don't have permissions to publish the gallery site, so this will likely not be visible immediately.

PR #575, opened 2020-01-02
Uninvert BarGroupHorizontal x scale

### LINKED ISSUE #573: BarGroupHorizontal expects an inverse x scale?
Hey folks! I find it odd that `BarGroupHorizontal` expects an inversed x scale (passing in `.range([xMax, 0])` as opposed to `.range([0, xMax])`) even though the chart itself isn't supposed to be rendered right to left.

To demonstrate with the demo in https://vx-demo.now.sh/bargrouphorizontal, if I use a bottom axis such as:

```jsx
<AxisBottom
  top={yMax}
  left={0}
  scale={xScale}
  stroke="#e5fd3d"
  tickStroke="#e5fd3d"
  tickLabelProps={() => ({
    fill: '#e5fd3d',
    fontSize: 11,
    textAnchor: 'middle',
    dy: '0.33em',
  })}
/>
```

The axis will be rendered backwards:
image

This seems unintuitive to me where you have to take the same scale and invert it just to accomodate the axis, and even potentially with markers as well.

This is caused by the line in:
https://github.com/hshoff/vx/blob/51f5cc3cfa12db6eb1703aceba1e204b19ca505d/packages/vx-shape/src/shapes/BarGroupHorizontal.tsx#L66

If you agree that the inversion based on width is unnecessary I'm happy to make a PR and fix it.

diagramatics:
Fixes #573.

#### :boom: Breaking Changes

- [shape] `BarGroupHorizontal` now takes an uninverted `xScale`.

### DIFF
--- a/packages/vx-demo/src/components/tiles/bargrouphorizontal.js
+++ b/packages/vx-demo/src/components/tiles/bargrouphorizontal.js
@@ -53,7 +53,7 @@ export default ({
   // scales
   y0Scale.rangeRound([0, yMax]);
   y1Scale.rangeRound([0, y0Scale.bandwidth()]);
-  xScale.rangeRound([xMax, 0]);
+  xScale.rangeRound([0, xMax]);
 
   return (
     <svg width={width} height={height}>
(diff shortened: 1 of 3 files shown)


williaster:
@diagramatics `0.0.193-alpha.1` is out with this fix 🎉

PR #576, opened 2020-01-04
Change Text's use of deprecated lifecycle

### LINKED ISSUE #341: @vx/threshold needs to generate unique clippath ids
Currently [Threshold](https://github.com/hshoff/vx/blob/master/packages/vx-threshold/src/Threshold.js#L26) is using the same `threshold-clip-below` and `threshold-clip-above` ClipPath ids which causes an issue when using more than 1 Threshold chart on a page:

image

We need to generate unique ids for each of these per instance.

### LINKED ISSUE #396: using strict with vx/text
`@vx/text` still has the soon to be deprecated lifecycle methods `componentWillMount` and `componentWillReceiveProps` which get flagged when running in strict mode.

I have created PR #395 which fits my needs but with little testing it is hard to say what this could break.

It would be hard to refactor to `getDerivedStateFromProps` with my limited knowledge due to the fact that this method is static and the existing code calls instance methods and it stores state in the instance variables `wordsWithComputedWidth` and `spaceWidth`.

What is the best way of refactoring this?

diagramatics:
#### :house: Internal

- [text] Remove deprecated lifecycles used in Text

Fixes #396. Closes #431. I thought I might as well take a crack at converting the deprecated lifecycles to something better. I was gonna use hooks but I have no idea if I can use it yet.

### DIFF
--- a/packages/vx-text/src/Text.tsx
+++ b/packages/vx-text/src/Text.tsx
@@ -85,16 +85,19 @@ class Text extends React.Component<TextProps, TextState> {
 
   private spaceWidth: number = 0;
 
-  // eslint-disable-next-line camelcase
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     this.updateWordsByLines(this.props, true);
   }
 
-  // eslint-disable-next-line camelcase
-  UNSAFE_componentWillReceiveProps(nextProps: TextProps) {
+  componentDidUpdate(prevProps: TextProps, prevState: TextState) {
+    // We calculated a new state, break out of the loop.
+    if (prevState.wordsByLines !== this.state.wordsByLines) {
+      return;
+    }
+
     const needCalculate =
-      this.props.children !== nextProps.children || this.props.style !== nextProps.style;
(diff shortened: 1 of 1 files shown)


williaster:
Nice, thanks for the PR @diagramatics! Code looks good to me but I will pull the branch to test it a bit. Meanwhile @techniq if you have thoughts let us know 👍 

We currently have not added hooks to enable compatibility with more react versions, but hope to do so soon after we land the TypeScript project.

techniq:
@diagramatics @williaster LGTM, but it's been a while since I wrote that to remember all the nuances (and have been using hooks / useEffect for too longer to remember all the nuances of the lifecycle methods 🤦‍♂).  I think if it tests out will with the demo we should be good.

I figure when we switch to hooks we would want to use [useLayoutEffect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect).

diagramatics:
Don't worry, I'm on the same boat here. I had to re-read the docs to figure out 😂

`useLayoutEffect` would be what we will reach for, yes.

PR #579, opened 2020-01-06
typescript(vx-demo): re-write in TypeScript

williaster:
#### :memo: Documentation

This is the final major PR for the [TypeScript project](https://github.com/hshoff/vx/projects/2), and re-writes the `vx-demo` (private) package which powers the [`vx` site](http://vx-demo.now.sh/) in TypeScript as a way to verify the usability of the rest of the types. It closes #571. 

I will update the following as examples are converted (**I will likely do many as separate PRs into this branch for review-ability**, merge them into this and this into master)

- [x] core components (`Page`, `Header`, `Nav`, etc.)
- [x] utils
- [x] `area` #595
- [x] `axis` #595
- [x] `bars` #593 
- [x] `bargroup` #593 
- [x] `bargrouphorizontal` #593 
- [x] `barstack` #593 
- [x] `barstackhorizontal` #593 
- [x] `brush` #598
- [x] `chord` #582 
- [x] `dendrogram` #596
- [x] `dots` #592 
- [x] `drag-i` #599
- [x] `drag-ii` #599
- [x] `geo-custom` #585
- [x] `geo-mercator`  #585
- [x] `gradients`  #583
- [x] `glyphs` #590
- [x] `heatmaps` #586
- [x] `legends` #587
- [x] `lineradial` #590
- [x] `lines` #581
- [x] `linkTypes` #594
- [x] `network`
  before/after

- [x] `pack`
  before/after

- [x] `patterns`
  before/after

- [x] `pies` #586
- [x] `polygons` #583
- [x] `radar` #590
- [x] `responsive`  #581
- [x] `stacked-areas` #593 
- [x] `statsplot` #584 
- [x] `streamgraph` #592 
- [x] `text` #597 
- [x] `threshold` #597
- [x] `tree` #589
- [x] `treemap` #589
- [x] `voronoi` #580
- [x] `zoom-i` #596

#### :boom: Breaking Changes

- The `@vx/mock-data` `Exoplanet` `radius` and `distance` values were updated from `string`s to `number`s to remove the need for consumers to coerce to numbers themselves

cc @hshoff @schillerk @kristw

PR #582, opened 2020-01-07
typescript(vx-demo): re-write Chord demo in TypeScript

williaster:
#### :memo: Documentation

This PR is part of #579 to re-write the `vx-demo` site in TypeScript. It re-writes the `chord` example in TypeScript and makes a couples `scaleOrdinal` fixes.

 **Before/After**
(note the color difference comes from updates made in #574 in master which are not live)

#### :bug: Bug Fix

This also makes some TypeScript generics fixes to `@vx/scale`s `scaleOrdinal` and `@vx/legend`s `LegendOrdinal` to allow for any `Input` that is coercible to a string (previously only `string` itself was allowed). This is consistent with the d3 docs + types.

#### Testing
- [x] functional
- [x] CI

cc @hshoff @kristw @schillerk


williaster:
going to merge this to leverage the scale type fixes in another PR

PR #592, opened 2020-01-09
typescript(vx-demo): rewrite Dots + Streamgraph examples in TypeScript

williaster:
### :memo: Documentation

This PR is part of #579 to re-write the `vx-demo` site in TypeScript. It re-writes the `Dots` + `Streamgraph` examples in TypeScript. 

### 🚀 Enhancements

Made `Stack` `Key` an optional generic in `@vx/shape`, previously it was `number | string`

 **Before/After**

#### Testing
- [x] functional
- [x] CI

cc @hshoff @kristw @schillerk

PR #593, opened 2020-01-10
typescript(vx-demo): rewrite Bars, BarGroup, BarGroupHorizontal, BarStack, BarStackHorizontal, and Area Stack in TypeScript

williaster:
### :memo: Documentation

This PR is part of #579 to re-write the `vx-demo` site in TypeScript. It re-writes the `Bars`, `BarGroup`, `BarGroupHorizontal`, `BarStack`,  `BarStackHorizontal`, and `AreaStack` examples in TypeScript. It's a doozy. 

### 🐛 Bug fixes

Lots of them 😱 

- **`@vx/shape`**
  - Fix the `Scale` `Input` and `Output` generics to be compatible with actual usage of `@vx/scale`s and allow scales that return `undefined` (e.g., `band` scales)
  - Fix the `x`, `y0`, and `y1` accessor props of `BarStack` and `y`, `x0`, and `x1` of accessor props `BarStackHorizontal` in to return `any`, as the corresponding `Scale`s accept `any` as input
  - Fix `restProps` types to `Omit` component props as well as props that are `Pick`ed from other types
  - Extend the changes in #592, and makes all `Stack` and `Group` component `Key`s optional generics, previously they were `number | string`
- Exports `SeriesPoint` for convenience

- **`@vx/grid`**
  - Fix the `Scale` type to have better compatible with actual usage of `@vx/scale`s and allow scales that return `undefined` (e.g., `band` scales)
  - Fix `restProps` types to `Omit` component props as well as props that are `Pick`ed from other types
  - Fix an issue where `restProps` for underlying `<Line />`s were not propagated to `Grid*` components
  - Fix the `Grid` component to accept different generics for the `XScaleInput` and `YScaleInput` (not just a single generic used for both, since they may be different)

- **`@vx/axis`**
  - Update return type of `Scale`s that return `this` to `any` because we can't fully capture the correct type

 **Before/After** 

#### Testing
- [x] functional
- [ ] CI

cc @hshoff @kristw @schillerk


williaster:
going to merge this to re-use the fixes in other PRs

PR #594, opened 2020-01-10
typescript(vx-demo): rewrite LinkTypes example in TypeScript

williaster:
### :memo: Documentation

This PR is part of #579 to re-write the `vx-demo` site in TypeScript. It re-writes the `LinkTypes` example in TypeScript. 

 **Before/After**

#### Testing
- [x] functional
- [x] CI

cc @hshoff @kristw @schillerk


williaster:
noting that we have to resolve `handlebars` to `4.5.3` because of a [breaking patch change](https://github.com/wycats/handlebars.js/blob/master/release-notes.md#v460---january-8th-2020) they released (again, we have no lock file so are subject to these things 😢 )

for reference we depend on this via `_project_#@airbnb#config-jest#jest#jest-cli#@jest#core#@jest#reporters#istanbul-reports#handlebars`

gonna merge to use the fix in other PRs

PR #599, opened 2020-01-11
typescript(vx-demo): rewrite Drag-i and Drag-ii demos in TypeScript

williaster:
#### :memo: Documentation

This PR is part of #579 to re-write the `vx-demo` site in TypeScript. It's THE LAST example re-write and converts the `Drag-i` + `Drag-ii` examples to TypeScript. 

#### 🐛 Bug fix

- Makes a fix in `@vx/event`s `localPoint` method where the `event.target` `node`’s `getScreenCTM` method instead of the `node`s owner `svg`s `getScreenCTM`

 **Before/After**

#### Testing
- [x] functional
  - [x] there is an unresolved issue with `Drag-i` locally where `dx`/`dy` are `0` every other frame which gives a jumping effect. trying to fix this but even reverting to the pre #499 re-write does not fix it 🤔 EDIT: fixed as noted in the bug section above
- [ ] CI

cc @hshoff @kristw @schillerk


packages/vx-demo/src/components/tiles/Drag-ii.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  if (width < 10) return null;
+  return (
+    <div className="DragII" style={{ touchAction: 'none' }}>
+      <svg width={width} height={height}>
+        <LinearGradient
+          id="stroke"
+          from="#ff614e"
+          to="#ffdc64"
+          gradientUnits="userSpaceOnUse" // need for gradients to render on straight lines
```
williaster: noting this interesting fix, noticed in another place I've used gradients

Issue #591, opened 2020-01-09
Update source code links in readmes to TS files


williaster:
Need to make the changes in #588 for all packages

dennisja:
Is anybody working on this or should I take it?

williaster:
@dennisja if you'd like to work on it I'm happy to review a PR! 🙌

dennisja:
I will start working on it today. Thanks.

dennisja:
Hello @williaster 
I have looked through all the Readmes and they are up to date. The only readme remaining with a reference to a `.js` is the vx-scale readme as seen [here](https://github.com/hshoff/vx/blame/master/packages/vx-scale/Readme.md#L33). However, this references a valid [`js` file](https://github.com/hshoff/vx/blob/master/packages/vx-demo/src/components/charts/SimpleBar.js). Should I go ahead and migrate that file to typescript and make that link reference the typescript version?

williaster:
Hi @dennisja ! Good to know we don't have a lot of references to specific files, really appreciate you digging into that more!

If you'd like to convert `vx-demo/src/components/charts/SimpleBar.js` to TS and update the readme, that'd be great!

PR #606, opened 2020-01-17
Migrate SimpleBar example to typescript

dennisja:
Fixes https://github.com/hshoff/vx/issues/591

#### :rocket: Enhancements

- Migrate `vx-demo/src/components/charts/SimpleBar.js` to typescript

#### :memo: Documentation

- Update references to `vx-demo/src/components/charts/SimpleBar.js`


packages/vx-demo/src/components/charts/SimpleBar.tsx
```diff
@@ -1,19 +1,22 @@
-import React from 'react';
+import React, { FC } from 'react';
 import { Bar } from '@vx/shape';
 import { Group } from '@vx/group';
 import { PatternLines } from '@vx/pattern';
 import { letterFrequency } from '@vx/mock-data';
 import { scaleBand, scaleLinear } from '@vx/scale';
 import { max } from 'd3-array';
+import { MarginShape } from '../../types/index';
```
williaster: nit – you can just write `... from '../../types` without the `index`

packages/vx-demo/src/components/charts/SimpleBar.tsx
```diff
@@ -1,19 +1,22 @@
...
-function round(value, precision) {
+function round(value: number, precision?: number): number {
   const multiplier = 10 ** (precision || 0);
   return Math.round(value * multiplier) / multiplier;
 }
 
-export default ({ width, height, margin }) => {
+type SimpleBarProps = { width: number; height: number; margin: MarginShape };
+
+const SimpleBar: FC<SimpleBarProps> = ({ width, height, margin }) => {
```
williaster: also a nit – personally I find `React.FC` more readable than `FC` alone, without the `React` context `FC` might be a little more ambiguous.

packages/vx-demo/src/components/charts/SimpleBar.tsx
```diff
@@ -1,19 +1,22 @@
-import React from 'react';
+import React, { FC } from 'react';
 import { Bar } from '@vx/shape';
 import { Group } from '@vx/group';
 import { PatternLines } from '@vx/pattern';
 import { letterFrequency } from '@vx/mock-data';
```
williaster: just noting that to fix the build, you'll probably need to update this to 
`import letterFrequency, { LetterFrequency } from '@vx/mock-data/lib/mocks/letterFrequency';`

which will allow you to use the `LetterFrequency` type in the accessor functions like `const x = (d: LetterFrequency) => d.letter;`

packages/vx-demo/src/components/codeblocks/SimpleBarCode.js
```diff
@@ -4,7 +4,7 @@ import Codeblock from './Codeblock.tsx';
 export default () => {
```
williaster: I think you can actually delete this file, it doesn't seem to be reference anywhere as far as I can tell (most of the gallery examples have actual source code, and then a string version of that code, but since this is not a gallery example I don't think it's in use)

dennisja: I have deleted it

packages/vx-demo/src/components/charts/SimpleBar.tsx
dennisja: Actually this wasn't the reason as to why the build was failing but I was able to find out why and fixed it.

PR #614, opened 2020-01-27
feat(vx-brush): adds resetOnEnd property

gillesdemey:
This will reset the brush after dragging, we've overlay-ed the `Brush` element on top of a chart so we can have zoom interaction without an additional smaller brushable chart.

brush-zoom-reset

#### :boom: Breaking Changes

- none

#### :rocket: Enhancements

- add a `resetOnEnd` property to the `BaseBrush` and `Brush` component.

#### :memo: Documentation

- No documentation yet for the `Brush` component :)

#### :bug: Bug Fix

- none

#### :house: Internal

- none


gillesdemey:
Not sure why the test suite is failing, but at first glance seems unrelated to the `@vx/brush` component 🤔

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -33,6 +33,7 @@ export type BaseBrushProps = {
   onClick?: (event: MouseHandlerEvent) => void;
   clickSensitivity: number;
   disableDraggingSelection: boolean;
+  resetOnEnd: boolean;
```
williaster: one nit – can we make this optional? `resetOnEnd?`

gillesdemey: Yep absolutely, should have thought of that 😅

williaster:
@gillesdemey could you try rebasing this on the latest master to see if it fixes the build? there have been PRs opened since this that have passed CI.

williaster:
#616 which duplicates this on the latest `master` does seem to work for me.

gillesdemey:
Looks like CI is green 🎉

williaster:
amazing, thanks again for the add, should be out soon in `0.0.194`

Issue #577, opened 2020-01-06
Controlled Brush


hellosmithy:
I have a use case for a brush that can be initialized with some state, more like a controlled component rather than an uncontrolled component (as it currently maintains it's own state).

For example on initial render there would already be a brushed area that the user can then drag to resize, rather than always starting from an un-brushed state.

I think this could be achieved by optionally passing in brush state rather than the brush managing it's own state. Is there appetite to support this use-case? I'm happy to submit a PR if so.

williaster:
Hey @hellosmithy 👋 thanks for the suggestion, I think there's definitely a use case for this! 

If not too complex to implement, it seems like we could have a dual controlled/uncontrolled component where you can optionally pass in the brush state as you say, but if you don't it will manage it on its own as it does now. 

Happy to review a PR or discuss implementation details more! 🙌

hellosmithy:
> we could have a dual controlled/uncontrolled component where you can optionally pass in the brush state

Yea that's my thoughts too. OK - I'll put together a PR then for you to review.

manufont:
I was looking for a way to control the Brush component by giving it a `brushPosition` props. Here's my solution :
```
import React, { useEffect, useState } from "react";
import { Brush } from "@visx/brush";

const positionToKey = ({ start, end }) => `${start.x},${start.y},${end.x},${end.y}`;

const ControlledBrush = ({ brushPosition, domainToPosition, ...brushProps }) => {
  const [innerPosition, setInnerPosition] = useState(brushPosition);
  const [key, setKey] = useState(positionToKey(brushPosition));

  const { onChange } = brushProps;

  useEffect(() => {
    if (positionToKey(innerPosition) !== positionToKey(brushPosition)) {
      setKey(positionToKey(brushPosition));
    }
  }, [innerPosition, brushPosition]);

  return (
    <Brush
      key={key}
      {...brushProps}
      initialBrushPosition={brushPosition}
      onChange={(domain) => {
        if (domain !== null) {
          setInnerPosition(domainToPosition(domain));
        }
        return onChange && onChange(domain);
      }}
    />
  );
};

export default ControlledBrush;
```

The `key` props change only when an external change to `brushPosition` is detected. This make it compatible with brush, move & resize interactions.

You have to specify the `domainToPosition` props witch is a function used to convert domain (`{x0, x1, y0, y1}`) into position (`{start: {x, y}, end:{x, y}}`). There's probably a better way to handle this part.

PR #618, opened 2020-02-01
new(vx-brush): handle mouseup outside of Brush extent

maryschmidt:
#### :boom: Breaking Changes

- None

#### :rocket: Enhancements

- Brush now listens to when a mouseup event happens on the window to disable brush drag
- Brush accepts an optional prop to default the scrubber state to an initial value

#### :memo: Documentation

-

#### :bug: Bug Fix

-

#### :house: Internal

-

### DIFF
--- a/packages/vx-brush/src/BaseBrush.tsx
+++ b/packages/vx-brush/src/BaseBrush.tsx
@@ -6,7 +6,7 @@ import Drag, { HandlerArgs as DragArgs } from '@vx/drag/lib/Drag';
 import BrushHandle from './BrushHandle';
 import BrushCorner from './BrushCorner';
 import BrushSelection from './BrushSelection';
-import { MarginShape, Point, BrushShape, ResizeTriggerAreas } from './types';
+import { MarginShape, Point, BrushShape, ResizeTriggerAreas, ScrubberShape } from './types';
 
 const BRUSH_OVERLAY_STYLES = { cursor: 'crosshair' };
 
@@ -16,6 +16,7 @@ type MouseHandlerEvent =
 
 export type BaseBrushProps = {
   brushDirection?: 'horizontal' | 'vertical' | 'both';
+  scrubberInitialState?: ScrubberShape;
   width: number;
   height: number;
   left: number;
@@ -46,27 +47,45 @@ export type UpdateBrush =
   | ((prevState: Readonly<BaseBrushState>, props: Readonly<BaseBrushProps>) => BaseBrushState);
 
 export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrushState> {
+  private constructor(props: BaseBrushProps) {
+    super(props);
+    let scrubberState;
+    if (this.props.scrubberInitialState) {
+      scrubberState = {
+        start: this.props.scrubberInitialState.start,
+        end: this.props.scrubberInitialState.end,
+        extent: this.getExtent(
+          this.props.scrubberInitialState.start,
+          this.props.scrubberInitialState.end,
+        ),
+      };
+    } else {
+      scrubberState = {
+        start: { x: 0, y: 0 },
+        end: { x: 0, y: 0 },
+        extent: {
+          x0: -1,
+          x1: -1,
+          y0: -1,
+          y1: -1,
+        },
+      };
+    }
+    this.state = {
+      ...scrubberState,
+      bounds: {
+        x0: 0,
+        x1: this.props.width,
+        y0: 0,
+        y1: this.props.height,
+      },
+      isBrushing: false,
+      activeHandle: null,
+    };
+  }
+
   private mouseUpTime: number = 0;
   private mouseDownTime: number = 0;
-
-  state = {
-    start: { x: 0, y: 0 },
-    end: { x: 0, y: 0 },
-    extent: {
-      x0: -1,
-      x1: -1,
-      y0: -1,
-      y1: -1,
-    },
-    bounds: {
-      x0: 0,
-      x1: this.props.width,
-      y0: 0,
-      y1: this.props.height,
-    },
-    isBrushing: false,
-    activeHandle: null,
-  };
 
   static defaultProps = {
     brushDirection: 'both',
@@ -88,15 +107,16 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     disableDraggingSelection: false,
     clickSensitivity: 200,
     resetOnEnd: false,
+    scrubberInitialState: null,
   };
-      
+
   componentDidMount = () => {
-    window.addEventListener("mouseup", this.handleDragEnd);
-  }
-  
+    window.addEventListener('mouseup', this.handleDragEnd);
+  };
+
   componentWillUnmount = () => {
-    window.removeEventListener("mouseup", this.handleDragEnd);
-  }
+    window.removeEventListener('mouseup', this.handleDragEnd);
+  };
 
   componentDidUpdate(prevProps: BaseBrushProps) {
     if (this.props.width !== prevProps.width || this.props.height !== prevProps.height) {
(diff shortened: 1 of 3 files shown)


packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -89,6 +89,14 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     clickSensitivity: 200,
     resetOnEnd: false,
   };
+      
+  componentDidMount = () => {
+    window.addEventListener("mouseup", this.handleDragEnd);
```
williaster: this will result in `handleDragEnd` being invoked (and subsequently `this.props.onChange/onBrushEnd` invocation and a `state` update) for `mouseup` events that have nothing to do with the `Brush` component. 

it's probably easiest to add a condition check in `this.handleDragEnd` that returns early / does nothing if `this.state.isBrushing` is `false`

williaster:
One other thought is that we need to prevent duplicate invocations of `onDragEnd`/`onChange` with the addition of this feature.

maryschmidt:
Sounds good, I'll update the PR

williaster:
this one won't make `0.0.194`

maryschmidt:
Sorry for the delay on this. Things have been absolutely crazy lately... 

After some thought, I believe we can address the vast majority of use cases for #577 by allowing users to provide a starting range for the selected area controller (I'm calling it `scrubber`). My reasoning is that `Brush` should own `BrushState` calculations + state. I certainly don't want to reproduce that logic in my own UI components... But would I like to be able to deep-link users to regions within a Brush + accompanying chart? Why yes, yes I would 😀 

Thoughts?

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -88,6 +107,15 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     disableDraggingSelection: false,
     clickSensitivity: 200,
     resetOnEnd: false,
+    scrubberInitialState: null,
+  };
+
+  componentDidMount = () => {
```
williaster, line 113: can we remove the arrow functions here? (lifecycle events should be normal functions)

packages/vx-brush/src/types.ts
```diff
@@ -19,13 +19,16 @@ export interface MarginShape {
...
 
-export interface BrushShape {
-  start: Point;
-  end: Point;
+export interface BrushShape extends ScrubberShape {
   extent: Bounds;
   bounds: Bounds;
 }
 
+export interface ScrubberShape {
```
williaster, line 27: let's not introduce new naming semantics for the `Brush`, maybe we can be very literal and boring and simply call this `BrushStartEnd`. otherwise everyone will be confused why it's called brush sometimes and scrubber sometimes.

packages/vx-brush/src/Brush.tsx
```diff
@@ -20,6 +20,7 @@ export type BrushProps = {
   onClick: BaseBrushProps['onClick'];
   margin: MarginShape;
   brushDirection: 'vertical' | 'horizontal' | 'both';
+  scrubberInitialState?: ScrubberShape;
```
williaster, line 23: can we call this `initialBrushPosition` throughout?

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -88,6 +107,15 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+    window.addEventListener('mouseup', this.handleDragEnd);
+  };
+
+  componentWillUnmount = () => {
+    window.removeEventListener('mouseup', this.handleDragEnd);
   };
 
   componentDidUpdate(prevProps: BaseBrushProps) {
```
williaster, line 121: I think ideally we would update `state` if the `initialBrushState` updates, but don't want to add a ton of checks in this method. I guess for now users could force re-render with a new component `key` 🤔 not sure if @hshoff has thoughts here.

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -88,6 +107,15 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
     clickSensitivity: 200,
     resetOnEnd: false,
+    scrubberInitialState: null,
+  };
+
+  componentDidMount = () => {
+    window.addEventListener('mouseup', this.handleDragEnd);
+  };
+
+  componentWillUnmount = () => {
```
williaster, line 117: ditto remove arrow function

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -16,6 +16,7 @@ type MouseHandlerEvent =
 
 export type BaseBrushProps = {
   brushDirection?: 'horizontal' | 'vertical' | 'both';
+  scrubberInitialState?: ScrubberShape;
```
williaster, line 19: ditto `initialBrushPosition`

packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -46,28 +47,46 @@ export type UpdateBrush =
   | ((prevState: Readonly<BaseBrushState>, props: Readonly<BaseBrushProps>) => BaseBrushState);
 
 export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrushState> {
+  private constructor(props: BaseBrushProps) {
```
williaster, line 50: I think you can remove the constructor and simplify this whole thing to this

```tsx
  state = {	
    start: { x: 0, y: 0 },	
    end: { x: 0, y: 0 },	
    ...this.props.initialBrushPosition,
    extent: this.props.initialBrushPosition ? 
      this.getExtent(
          this.props.scrubberInitialState.start,
          this.props.scrubberInitialState.end,
      ) : {	
      x0: -1,	
      x1: -1,	
      y0: -1,	
      y1: -1,	
    },	
    bounds: {	
      x0: 0,	
      x1: this.props.width,	
      y0: 0,	
      y1: this.props.height,	
    },	
    isBrushing: false,	
    activeHandle: null,	
  };
```

packages/vx-brush/src/types.ts
maryschmidt: Good idea! Thanks

packages/vx-brush/src/BaseBrush.tsx
maryschmidt: This is a tough one... I've been thinking about this as well.

For now, I think pairing `key` with `initialBrushState` makes the most sense in this context. - It avoids the hazards of getting into "semi-controlled component" territory where users would _re_-default initial values but the component would otherwise manage its own logic.
- The behavior would be consistent with conventions around setting default values on other input elements (dropdowns, date pickers, HTML inputs, etc.) 

The downside I see is this wouldn't support animating between changes. My own immediate use case for this is for linking to different regions of a chart, so that would definitely be nice to have, but like you said, I would like to avoid adding checks everywhere...

packages/vx-brush/src/BaseBrush.tsx
maryschmidt: We need the constructor in order to get around `this.getExtent` being used before its initialization, barring pasting the function about the state initialization, but that's a little unconventional. Thoughts?

williaster: ah, I see. 

I think the `constructor` is fine, but could we update to the more succinct / immutable syntax I suggested here (assigning to `this.state` in the `constructor` instead)?

maryschmidt:
Updated! Thanks for the feedback 😄

hshoff:
Awesome, thanks for the contribution @maryschmidt!

minikomi:
Was just looking into how to set `initialBrushState` for the brush -- hope this can be merged soon (small minor release?)

williaster:
sorry for the delay in the release, I'm testing this now and it doesn't really work :( will have a fix out soon / then release.

williaster:
out in `0.0.196`

PR #621, opened 2020-02-03
fix(vx-responsive): fix enhancer parent/screen size checks

### LINKED ISSUE #619: withParentSize no longer delays rendering until size is known
Previously, `withParentSize` would only render its children when the size was known. However, since commit https://github.com/hshoff/vx/commit/b05a17fbc92a698b036c64886b8679ee2475be0e, the children are initially rendered without size. This breaks components that expect `parentWidth` and `parentHeight` to be available.

The problem is that the state is initialized with `undefined`, while the render method checks for `null`.

williaster:
#### :bug: Bug Fix

Fixes #619 where the TS re-write introduced a regression wherein the state is initialized with `undefined`, while the render method checks for `null`s

@hshoff @kristw 
cc @dson

### DIFF
--- a/packages/vx-responsive/src/enhancers/withParentSize.tsx
+++ b/packages/vx-responsive/src/enhancers/withParentSize.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import debounce from 'lodash/debounce';
 import ResizeObserver from 'resize-observer-polyfill';
 
+const CONTAINER_STYLES = { width: '100%', height: '100%' };
+
 export type WithParentSizeProps = {
   debounceTime?: number;
 };
@@ -58,6 +60,10 @@ export default function withParentSize<BaseComponentProps extends WithParentSize
       if (this.resizeObserver) this.resizeObserver.disconnect();
     }
 
+    setRef = (ref: HTMLDivElement) => {
+      this.container = ref;
+    };
+
     resize = ({ width, height }: { width: number; height: number }) => {
       this.setState({
         parentWidth: width,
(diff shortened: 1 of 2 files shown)


packages/vx-responsive/src/enhancers/withParentSize.tsx
```diff
@@ -68,13 +74,8 @@ export default function withParentSize<BaseComponentProps extends WithParentSize
...
       return (
-        <div
-          style={{ width: '100%', height: '100%' }}
-          ref={ref => {
-            this.container = ref;
-          }}
-        >
-          {parentWidth !== null && parentHeight !== null && (
+        <div style={CONTAINER_STYLES} ref={this.setRef}>
+          {parentWidth != null && parentHeight != null && (
```
kristw: Do we want to check for `0` as well?

packages/vx-responsive/src/enhancers/withScreenSize.tsx
```diff
@@ -54,7 +54,7 @@ export default function withScreenSize<BaseComponentProps extends WithScreenSize
 
     render() {
       const { screenWidth, screenHeight } = this.state;
-      if (!screenWidth && !screenHeight) return null;
+      if (screenWidth == null || screenHeight == null) return null;
```
kristw: Do we want to check for `0` as well?

williaster:
hmm 🤔  if it is zero and a consumer wants to handle that case for some reason they won't be able to. we were previously inconsistent between the two HOCs, `parentSize` **would** render with `0` and `screenSize` **would not** (but only if both were `0`). 

I think it's okay for us to pass zeros and let the consumer handle that case for now.

williaster:
Also noting that rendering with `0`s mirrors the behavior of `<ParentSize />` which does not do any check.

Issue #609, opened 2020-01-23
withTooltip enhancer can't be used on components that render within an svg


ptmx:
The `withTooltip` enhancer from `@vx/tooltip` currently adds a container div that wraps the component: https://github.com/hshoff/vx/blob/f8e6c6d20dd146de8fbf3879aa00806619ad0ffc/packages/vx-tooltip/src/enhancers/withTooltip.tsx#L75-L87

Since HTML elements can't be appended to SVG elements, this means that the `withTooltip` enhancer can't be used on components that will be rendered within an SVG, such as an `Axis`.

I made a [minimal example](https://codesandbox.io/s/vx-with-tooltip-gchfo) of a reusable `AxisBottomWithTooltip` component that adds tooltips to axis tick labels, but I have a similar use case in a charting library I'm building for a project.

In the example above, I modified `withTooltip` to render the wrapped component directly, without inserting the container `div`, so the axis renders correctly with the tooltip even though it's rendered within an SVG element.  

Using the `withTooltip` enhancer from `@vx/tooltip` instead, [nothing is rendered](https://codesandbox.io/s/vx-with-tooltip-wd4mb) due to the container `div`. 

The downside of removing the container div is that it would be a breaking change for any existing code which relies on the container, so users would have to migrate by adding such a container around their enhanced component instead.  From my point of view that seems like the more flexible approach, but I might be missing some context on why it's helpful for the container `div` to be added by `withTooltip`.

williaster:
Hey @ptmx 👋 thanks for checking out `vx` and raising this issue. I think you point is definitely fair and there could be a use for using `withTooltip` with components that are children of an `svg`. I would point out that an alternative implementation of your axis tooltip using the current `withTooltip` is to simply [pass the `showTooltip` + `hideTooltip` from the parent `svg` down to the child `Axis` component](https://codesandbox.io/s/vx-with-tooltip-zdlkc?fontsize=14&hidenavigation=1&theme=dark).

Still, your proposal could be more flexible. I wonder if rather than making a breaking change and removing the `div` container altogether if we could instead add a hook to render the container element, which would allow a consumer to render **no container** if they wanted (and render the current `div` with necessary `position: 'relative'` style) by default. wdyt of that approach?

ptmx:
[...]

PR #631, opened 2020-02-22
Add useTooltip hook

ptmx:
#### :rocket: Enhancements

- Add a `useTooltip` hook to manage tooltip state for functional components, as discussed in https://github.com/hshoff/vx/issues/609

#### :boom: Breaking Changes

- This PR introduces `useState`, which requires bumping the `peerDep` for react to `^16.8.0-0`

#### :memo: Documentation

- Update `vx-tooltip` readme to include the new `useTooltip` hook, and recommend a state management approach depending on use case (specifically: recommending `useTooltip` for functional components, or `withTooltip` for class components)
- Rewrite the `BarStack` demo to use `useTooltip` instead of `withTooltip`, so there is an example of each approach to managing tooltip state in the gallery (`BarStack` for `useTooltip`, `BarStackHorizontal` for `withTooltip`)

#### :house: Internal

- Rewrote `withTooltip` in terms of `useTooltip`, to avoid maintaining multiple different implementations

### DIFF
--- a/packages/vx-demo/src/pages/BarStack.tsx
+++ b/packages/vx-demo/src/pages/BarStack.tsx
@@ -19,8 +19,7 @@ import { AxisBottom } from '@vx/axis';
 import cityTemperature, { CityTemperature } from '@vx/mock-data/lib/mocks/cityTemperature';
 import { scaleBand, scaleLinear, scaleOrdinal } from '@vx/scale';
 import { timeParse, timeFormat } from 'd3-time-format';
-import { withTooltip, Tooltip } from '@vx/tooltip';
-import { WithTooltipProvidedProps } from '@vx/tooltip/lib/enhancers/withTooltip';
+import { useTooltip, Tooltip } from '@vx/tooltip';
 import { LegendOrdinal } from '@vx/legend';
 import { ShowProvidedProps } from '../../types';
 
@@ -76,7 +75,7 @@ const colorScale = scaleOrdinal<CityName, string>({
 
 let tooltipTimeout: number;
 
-export default withTooltip<ShowProvidedProps, TooltipData>(
+export default 
   ({
     width,
     height,
@@ -87,13 +86,16 @@ export default withTooltip<ShowProvidedProps, TooltipData>(
       bottom: 0,
       left: 0,
     },
-    tooltipOpen,
-    tooltipLeft,
-    tooltipTop,
-    tooltipData,
-    hideTooltip,
-    showTooltip,
-  }: ShowProvidedProps & WithTooltipProvidedProps<TooltipData>) => {
+  }: ShowProvidedProps) => {
+    const {
+      tooltipOpen,
+      tooltipLeft,
+      tooltipTop,
+      tooltipData,
+      hideTooltip,
+      showTooltip
+    } = useTooltip<TooltipData>();
+
     if (width < 10) return null;
     // bounds
     const xMax = width;
@@ -208,8 +210,7 @@ export default withTooltip<ShowProvidedProps, TooltipData>(
         )}
       </div>
     );
-  },
-);
+  }
 `}
     </Show>
   );
--- a/packages/vx-tooltip/Readme.md
+++ b/packages/vx-tooltip/Readme.md
@@ -4,10 +4,127 @@
 npm install --save @vx/tooltip
 ```
 
-The `@vx/tooltip` package provides utilities for making it easy to add `Tooltip`s to a visualization and includes higher-order component (HOC) enhancers and Tooltip components.
+The `@vx/tooltip` package provides utilities for making it easy to add `Tooltip`s to a visualization and includes hooks, higher-order component (HOC) enhancers and Tooltip components.
 
-### Example:
-``` js
+### Hooks and Enhancers
+
+This package provides two ways to add tooltip state logic to your chart components:
+
+- a hook: `useTooltip()` 
+- a higher order component (HOC): `withTooltip()`
+
+The `useTooltip` hook is the recommended way to add tooltip state logic to your components, but can only be used in functional components. The `withTooltip` HOC can be used with both functional and class components, and is the recommended way to add tooltip state logic to class components.
+
+Both `useTooltip` and `withTooltip` expose the same values and functions for use in your component:
+
+| Name          | Type   | Description                                                                                                                                           |
+| :------------ | :----- | :---------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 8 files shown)


packages/vx-tooltip/package.json
```diff
@@ -35,7 +35,7 @@
     "prop-types": "^15.5.10"
   },
   "peerDependencies": {
-    "react": "^15.0.0-0 || ^16.0.0-0"
+    "react": "^16.8.0-0"
```
ktmud: Why not bump it to “>= 16.8.0 <= 16.12.0”? Sure later versions are also supported, right?

packages/vx-tooltip/Readme.md
```diff
@@ -4,10 +4,127 @@
...
+
+const ChartWithTooltip = () => {
+  const {
+    tooltipData,
+    tooltipLeft,
+    tooltipTop,
+    tooltipOpen,
+    showTooltip,
+    hideTooltip,
+  } = useTooltip();
```
ktmud: Is it possible to create a “Tooltip” class and return a `tooltip` instance with the hook? `const [tooltip, updateTooltip] = useTooltip()` feels a more consistent to how people would use other hooks. It’s also more scalable in case we want to add more API to the tooltip.

packages/vx-tooltip/Readme.md
```diff
@@ -4,10 +4,127 @@
...
+The `@vx/tooltip` package provides utilities for making it easy to add `Tooltip`s to a visualization and includes hooks, higher-order component (HOC) enhancers and Tooltip components.
 
-### Example:
-``` js
+### Hooks and Enhancers
+
+This package provides two ways to add tooltip state logic to your chart components:
+
+- a hook: `useTooltip()` 
+- a higher order component (HOC): `withTooltip()`
```
ktmud: If `withTooltip` is to be removed (Or deprecated`), docs about HOC should probably be removed, too.

williaster: I don't think there's any reason to deprecate the HOC, I think this would be an unnecessary breaking change. 

They are complementary approaches and there are valid uses for both for now.

packages/vx-tooltip/package.json
williaster: This will match any minor versions above and including `16.8`, you can try it [here](https://semver.npmjs.com/).

packages/vx-tooltip/Readme.md
williaster: I think returning  an `object` of the same `prop`s that are passed via `withTooltip` makes sense. 

I've seen several examples of hooks that use this pattern (e.g., `react-three-fiber`'s [`useThree` hook](https://github.com/react-spring/react-three-fiber#usethree-sharedcanvascontext)), and it would avoid a breaking change for `withTooltip`.

packages/vx-tooltip/src/enhancers/withTooltip.tsx
```diff
@@ -33,61 +20,11 @@ export default function withTooltip<BaseComponentProps = {}, TooltipData = {}>(
...
-          updateTooltip={this.updateTooltip}
-          showTooltip={this.showTooltip}
-          hideTooltip={this.hideTooltip}
-          {...this.state}
-          {...this.props}
-        />,
-        containerProps,
-      );
-    }
+    return renderContainer(<BaseComponent {...props} {...tooltipProps} />, containerProps);
```
williaster: just noting that previously we spread the `WrappedComponent` `props` _after_ `tooltipProps`, so this might result in a change in behavior if a consumer did pass an overlapping `prop`. I'm not really sure this is a big concern, but would prevent doing things like overriding `tooltipOpen` 🤷‍♂  any thoughts on this?

packages/vx-tooltip/src/hooks/useTooltip.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+type UseTooltipState<TooltipData> = Pick<
+  UseTooltipParams<TooltipData>,
+  'tooltipOpen' | 'tooltipLeft' | 'tooltipTop' | 'tooltipData'
+>;
+type ShowTooltipArgs<TooltipData> = Omit<UseTooltipState<TooltipData>, 'tooltipOpen'>;
+type UpdateTooltipArgs<TooltipData> = UseTooltipState<TooltipData>;
+
+export default function useTooltip<TooltipData = {}>(): UseTooltipParams<TooltipData> {
+  const [{ tooltipOpen, tooltipLeft, tooltipTop, tooltipData }, setTooltipState] = useState<
```
williaster: Since they're mostly updated together, I agree that a single `useState` with _all_ state is better than storing each variable separately in different `hook`s 👍

packages/vx-tooltip/package.json
ktmud: Ah, I must’ve confused caret and tilde. Thank you!

packages/vx-tooltip/src/enhancers/withTooltip.tsx
ptmx: Ah, good catch!  I agree it's not super likely that any consumers are relying on that behavior now, but in the absence of a reason to change it I think it makes sense to stay consistent with the existing behavior.  Will update shortly.

packages/vx-demo/src/pages/BarStack.tsx
```diff
@@ -87,13 +86,16 @@ export default withTooltip<ShowProvidedProps, TooltipData>(
...
     },
-    tooltipOpen,
-    tooltipLeft,
-    tooltipTop,
-    tooltipData,
-    hideTooltip,
-    showTooltip,
-  }: ShowProvidedProps & WithTooltipProvidedProps<TooltipData>) => {
+  }: ShowProvidedProps) => {
+    const {
```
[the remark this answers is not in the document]
ptmx: This stuck out to me as something to come back to as well.  It might be doable with `raw-loader` or [`raw.macro`](https://www.npmjs.com/package/raw.macro) but I didn't have time to dig into it more deeply.  Now this has me curious so I'll look into it later in the week if I have a chance 🤔

packages/vx-tooltip/src/enhancers/withTooltip.tsx
williaster: sounds good, thanks!

packages/vx-demo/src/pages/BarStack.tsx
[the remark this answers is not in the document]
williaster: sounds good! Created #632 to track this, I'll update status there if I start work on it. `raw.macro` looks interesting, linked to how `react-window` does it in the issue as well 👍

williaster:
this is now published under [`0.0.195`](https://github.com/hshoff/vx/releases/tag/v0.0.195)

PR #634, opened 2020-02-27
fix: BrushSelection offset in brush demo

### LINKED ISSUE #620: (vx-demo/vx-brush) Brush drag has offset when group has margin.
The brush positioning seems to be affected by the margin of the group surrounding it, which leads to the area being drawn with an offset from where your cursor is. I noticed this in my own code, thought that I must be doing something wrong, but the example (which is quite recent) has the same issue: https://github.com/hshoff/vx/blob/master/packages/vx-demo/src/components/tiles/Brush.tsx

I am not sure if this is a bug (if it is at all) with the Brush package or if Drag is not working as intended (as the values come from there).

ufOwXXrGjC

geekplux:
#### :boom: Breaking Changes

-

#### :rocket: Enhancements

-

#### :memo: Documentation

-

#### :bug: Bug Fix

- Fix brush selection offset issue, close #620. Update source code in demo page.

#### :house: Internal

-


williaster:
Thanks for the fix @geekplux ! 🙌 

Pulled it locally to verify the fix 👍 and I fixed a conflict with master (we removed the copy/paste-ing of the example code in #637 🎉 )

geekplux:
thank you, let me check if there anything I can help further

Issue #632, opened 2020-02-24
[internal] load and render example source code programmatically instead of duplicating source as a string


williaster:
Currently developers must copy/paste example source code into a string version to render alongside the example (e.g., [Axis](https://github.com/hshoff/vx/blob/master/packages/vx-demo/src/pages/Axis.tsx)). 

This leads to stale source code strings and is generally cumbersome for the developer. We should be able to use a webpack loader (e.g., [`react-window`](https://github.com/bvaughn/react-window/blob/51a24c4972c9ca5eee4df5ea3ec25bc4328e40c5/website/webpack/codemirror-loader.js) or [`raw.macro`](https://www.npmjs.com/package/raw.macro)) to automate this.

PR #637, opened 2020-03-01
refactor(vx-demo): Remove duplication of demo example source

ptmx:
#### :house: Internal

- Removed duplication of gallery example source by importing the source directly with `raw-loader` (closes https://github.com/hshoff/vx/issues/632).  This involved relaxing a couple `eslint` rules, but at least the changes to `import/no-unresolved` could be more narrowly scoped to only ignore the `!!raw-loader` imports specifically while still applying to all other imports.


williaster:
Wowow, amazing @ptmx! will review ASAP and test locally 👍

williaster:
just confirming that these all look similar, example for horizontal stacked bar below. will file a new issue to improve the styling for TS instead of JS 😉 

**before** 
image


**after**
image

Issue #638, opened 2020-03-01
Travis build doesn't run lint step


ptmx:
It looks like the intention of the Travis config is to run tests and lint in parallel:

https://github.com/hshoff/vx/blob/6bdca40d9b7150b307328f1c07543b06bb6c5fcd/.travis.yml#L14-L19

However, the actual build only runs the default test script, i.e., it runs `yarn test` [since `yarn.lock` is present](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#yarn-is-supported).  `yarn lint` doesn't get executed at all.

For example, see the logs from the latest build on `master`: https://travis-ci.org/hshoff/vx/builds/655661819.  The build runs `yarn test` instead of `yarn run jest -w 4`  as specified in the config file, and omits the `yarn lint` step altogether.

One option is to override the default `script`:

```
script:
  - yarn run jest -w 4
  - yarn run lint
```

Although this wouldn't run those steps in parallel.  Alternatively I'm sure there's a way to accomplish this with the existing build matrix too, although from skimming the docs for a bit it's not immediately obvious to me how the `script` and `jobs` keys interact.

williaster:
Thanks for flagging this @ptmx and digging into the root-cause (I didn't know about the default `yarn test` logic), this explains some strange observations on travis. I'll dig into docs more to see if there's a solid solution for running it all in parallel 👍

ptmx:
@williaster I ended up digging into it a bit more and opening https://github.com/hshoff/vx/pull/640 to fix this.  I found that the `matrix` and `jobs` keys conflicted with each other, but once I consolidated those two configs then Travis correctly ran the two jobs in parallel instead of the default `yarn test` 👍

williaster:
🙉  sorry I missed that, I converged on the same thing with #641, but I'll close that in favor of yours.

One other bit of info I found after talking to some folks was that the parallel mode can actually be _slower_ because it has to completely setup and yarn install the project for each parallel job. In that case we might actually want to use a single `test` step with the `script: yarn run jest -w 4 && yarn run lint`

I'll try this to see if it's faster for vx in #641 just to compare the times, and circle back 👍

ptmx:
> 🙉 sorry I missed that, I converged on the same thing with #641, but I'll close that in favor of yours.
> 
[...]

PR #640, opened 2020-03-01
ci(travis): Move fast_finish to jobs key

ptmx:
#### :house: Internal

- Consolidated the `matrix` and `jobs` configuration in `.travis.yml`, since those two keys override each other.  This was causing Travis to run the default test script (`yarn test`) instead of running both the test and lint steps as desired (closes https://github.com/hshoff/vx/issues/638).

~~Note: the CI build for this this will (correctly) fail, since the linter will throw errors until https://github.com/hshoff/vx/pull/639 is merged.~~ I've rebased over the latest master and force pushed, so I'd expect CI to pass now 🤞

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ install:
   - yarn install --ignore-engines
 before_script:
   - yarn run build
-matrix:
-  fast_finish: true
 jobs:
+  fast_finish: true
   include:
     - name: Test
       script: yarn run jest -w 4


williaster:
Below are the speed results from my experiments in #641 for **parallel** linting + testing vs a single step that runs them **sequentially**. Overall the sequential job runs in about the amount of time as a **single** lint or test job (likely because it has to completely setup and yarn install the project for each parallel job as mentioned in #638). 

So we don't get a speed gain by running them in parallel, but practically it doesn't make a huge difference since they're done in `8-9 min`. All else equal, it seems like there **is** a benefit of keeping the tasks separate so you can identify which failed in the UI. Given that this is a strict improvement I think we should just merge it 🚀 

**Running test + lint in _parallel_**
image

**Running test + lint _sequentially_**
image

ptmx:
> Below are the speed results from my experiments in #641 for **parallel** linting + testing vs a single step that runs them **sequentially**. Overall the sequential job runs in about the amount of time as a **single** lint or test job (likely because it has to completely setup and yarn install the project for each parallel job as mentioned in #638).
> 
> So we don't get a speed gain by running them in parallel, but practically it doesn't make a huge difference since they're done in `8-9 min`. All else equal, it seems like there **is** a benefit of keeping the tasks separate so you can identify which failed in the UI. Given that this is a strict improvement I think we should just merge it 🚀
> 
> **Running test + lint in _parallel_** [link](https://travis-ci.org/hshoff/vx/builds/657524085?utm_medium=notification&utm_source=github_status)
> image
> 
> **Running test + lint _sequentially_** [link](https://travis-ci.org/hshoff/vx/builds/657529416?utm_source=github_status&utm_medium=notification)
> image

Right on, thanks for benchmarking it!  Yeah, unfortunately the bulk of the time is taken by installing and building, so the parallelization doesn't speed things up much.  Agreed that it does make it clearer at a glance whether linting or testing caused the build to fail, so it doesn't hurt to keep them split up.

PR #641, opened 2020-03-02
internal(travis): run jest + lint in parallel

### LINKED ISSUE #638: Travis build doesn't run lint step

williaster:
#### :house: Internal

This is an attempt to fix #638, might need some iteration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ install:
   - yarn install --ignore-engines
 before_script:
   - yarn run build
-matrix:
-  fast_finish: true
 jobs:
+  fast_finish: true
   include:
     - name: Test
       script: yarn run jest -w 4
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ jobs:
   fast_finish: true
   include:
     - name: Test
-      script: yarn run jest -w 4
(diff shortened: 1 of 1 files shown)


williaster:
closing in favor of #640

PR #645, opened 2020-03-07
Fix typo within readme

hamlim:
Additionally, should the section header above this fix be `Rebuild one package` (one instead of on)?

#### :memo: Documentation

- Fix typo in root readme


williaster:
Thanks for the fix @hamlim 🙌 

I made the `one` fix also 👍

PR #642, opened 2020-03-05
new(demo): Add proof-of-concept code sandbox example

williaster:
#### :house: Internal

This PR closes #628 for creating a proof-of-concept code sandbox example. It:
- refactors the `Axis` demo example to be a standalone app which enables linking out to code sandbox
- adds a link out to to code sandbox in the `<Show />` component so it can be reused as other examples are refactored. Example code sandbox [link for the example below](https://codesandbox.io/s/github/hshoff/vx/tree/chris--poc-code-sandbox/packages/vx-demo/src/docs-v2/examples/vx-axis).
- converts `gallery.js` => `Gallery.tsx` which I missed in the TS refactor

**before**

**after** code sandbox [link](https://codesandbox.io/s/github/hshoff/vx/tree/chris--poc-code-sandbox/packages/vx-demo/src/docs-v2/examples/vx-axis)

#### :bug: Bug Fix
- loads the prism css styles so codeblocks are actually styled (see images above)

- Fixes an issue where `Drag-i` didn't update items when the parent size changed; this results in incorrectly-place items during an initial gallery rendering where width is near `0`

**before**

**after**

- Fixes an issue where `Drag-ii` did not show default data (was introduced in the demo TS re-write)
**before**

**after**

@hshoff @kristw

Issue #624, opened 2020-02-14
[docs] link out to codesandbox from examples


williaster:
We've gotten a lot of requests to link out to codesandbox from demo examples. 

This should be pretty straight forward using their github API https://codesandbox.io/docs/importing#import-from-github 

This will require refactoring each demo example to be a standalone package.

geekplux:
Sounds good and useful for promotion, and I think tutorials(#11 ) or guides are needed if we want more users, we could provide these docs embedding codesandbox https://codesandbox.io/docs/embedding#example-embeds

williaster:
Enumerating the examples to be ported / tracking status here cc @pedroapfilho

- [x] Axis #642 
- [ ] Areas
- [ ] BarGroup
- [ ] BarGroupHorizontal
- [ ] Bars
- [ ] BarStack
- [ ] BarStackHorizontal
- [ ] Brush
- [ ] Chord
- [ ] Dendrograms
- [ ] Dots
- [ ] Drag-i
- [ ] Drag-ii
- [ ] Geo-Custom
- [ ] Geo-Mercator
- [ ] Glyphs
- [ ] Gradients
- [ ] Heatmap
- [ ] Legends
- [ ] LineRadial
- [ ] Lines
- [ ] LinkTypes
- [ ] Network
- [ ] Pack
- [ ] Patterns
- [ ] Pies
- [ ] Polygons
- [ ] Radar
- [ ] Responsive
- [ ] Stacked-Areas
- [ ] Statsplot
- [ ] Streamgraph
- [ ] Threshold
- [ ] Treemap
- [ ] Trees
- [ ] Voronoi
- [ ] Zoom-i

PR #648, opened 2020-03-11
new(demo): link `Streamgraph` + `LineRadial` examples to codesandbox

williaster:
#### :house: Internal

This PR is part of #624 and converts the `Streamgraph` + `LineRadial` examples to link out to code sandbox as demo-ed in #642. For both demos I've updated to include `react-spring` hooks animation as we've gotten many requests for this and I had them partially implemented elsewhere.

**Before**
image

image

**After**
[Example codesandbox link](https://codesandbox.io/s/github/hshoff/vx/tree/chris--sandbox-streamgraph-radial/packages/vx-demo/src/docs-v2/examples/vx-shape-line-radial) (update to master branch upon merge)

lineradial-animated

[Example codesandbox link](https://codesandbox.io/s/github/hshoff/vx/tree/chris--sandbox-streamgraph-radial/packages/vx-demo/src/docs-v2/examples/vx-streamgraph) (update to master branch upon merge)
streamgraph-animated

@kristw @hshoff 
cc @pedroapfilho


packages/vx-demo/src/docs-v2/examples/vx-shape-line-radial/Example.tsx
```diff
@@ -0,0 +1,160 @@
...
+const background = '#744cca';
+const darkbackground = '#603FA8';
+const springConfig = {
+  tension: 100,
+  friction: 150,
+};
+
+// utils
+function extent<Datum>(data: Datum[], value: (d: Datum) => number) {
+  return [Math.min(...data.map(value)), Math.max(...data.map(value))];
```
kristw: could extract `const values = data.map(value)` to reduce mapping to once.

packages/vx-demo/src/docs-v2/examples/vx-streamgraph/useForceUpdate.ts
```diff
@@ -0,0 +1,6 @@
+import { useState } from 'react';
```
kristw: Should `react` version requirement be `^16.8` to support hooks?

williaster: ah yes, good catch. there's another dep issue I'm not sure how to handle quiet yet where after all of the examples are refactored, all of the deps should be in the independent sandbox `package.json`s but that doesn't work with `yarn` workspaces because technically they are nested:

```
vx-root/
package.json
...
  packages/
    vx-axis/
      package.json
      ...
    ...
    vx-demo/
      package.json     <-- all deps currently in workspace
      sandboxes/
        sandbox/
          package.json <-- slowly getting moved here, nested inside a workspace
```

will prob figure that out when we get there.

Issue #650, opened 2020-03-11
Remove default styling for Tooltip


jgonera:
Would you be open to remove these default styles for `Tooltip`?

https://github.com/hshoff/vx/blob/ac76afad960c4715a1f339b395e5b1e6a4e7de9c/packages/vx-tooltip/src/tooltips/Tooltip.tsx#L25-L32

It seems like most of VX doesn't assume much styling, but for `Tooltip` I need to override all of these with `!important` in my CSS modules. I know one could just replace `Tooltip` with a `div`, but `TooltipWithBounds` is harder to replace.

geekplux:
Inline `style` should work, but what you mentioned is a problem exactly, using css files to add style might be better

williaster:
This component is quite simple so I'm curious about the benefits of this versus creating your own component? if you are using `withTooltip`, you should be able to specify your own Tooltip component there. 

Alternatively we could update the component such that these current `styles` are set by default:

```
export default function Tooltip({
  className,
  top,
  left,
  style = defaultStyles,
  children,
  ...restProps
}: TooltipProps & JSX.IntrinsicElements['div']) {
  return (
    <div
      className={cx('vx-tooltip-portal', className)}
      style={style}
      {...restProps}
    >
      {children}
    </div>
  );
}
```

I think in this case we should export the default styles so if you want to e.g., override 1 style, you don't have to define all of them and could do `<Tooltip style={{ ...defaultStyles, color: 'magenta' }} ... />`.

I don't have a ton of bandwidth right now and am focusing on improved docs, but would happily review a PR!

dennisja:
I can make a PR to do this, What do you guys think  about using CSS for the default styles?

hshoff:
Historically, vx has avoided shipping CSS. I would prefer the defaultStyles export because it is lower friction.

As @williaster highlights, extension is easy:
```jsx
<Tooltip style={{ ...defaultStyles, color: 'magenta' }} ... />
```

But in terms of API how do folks feel about opting out of the default inline styles?

Option A) Pass empty object
```jsx
<Tooltip style={{}} ... />
```

Option B) Prop flag
```jsx
<Tooltip unstyled />
```

For option b, the code becomes:

```jsx
export default function Tooltip({
  className,
  top,
  left,
  unstyled = false,
  style = defaultStyles,
  children,
  ...restProps
}: TooltipProps & JSX.IntrinsicElements['div']) {
  return (
    <div
[...]

PR #666, opened 2020-04-25
Remove tooltip default styles

dennisja:
#### :boom: Breaking Changes

-

#### :rocket: Enhancements

- Enable users to use or remove default tooltip styles

#### :memo: Documentation

- Update tooltip docs

#### :bug: Bug Fix

-

#### :house: Internal

-

Fixes https://github.com/hshoff/vx/issues/650

### DIFF
--- a/packages/vx-tooltip/package.json
+++ b/packages/vx-tooltip/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "@types/classnames": "^2.2.9",
     "@types/react": "*",
-    "@vx/bounds": "^0.0.195",
+    "@vx/bounds": "0.0.195",
     "classnames": "^2.2.5",
     "prop-types": "^15.5.10"
   },
--- a/packages/vx-tooltip/package.json
+++ b/packages/vx-tooltip/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "@types/classnames": "^2.2.9",
     "@types/react": "*",
-    "@vx/bounds": "0.0.195",
+    "@vx/bounds": "^0.0.195",
     "classnames": "^2.2.5",
     "prop-types": "^15.5.10"
   },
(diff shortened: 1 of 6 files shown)


williaster:
@dennisja thanks for the PR including tests! 🙌 

It differs slightly from the solution @hshoff discussed/proposed in #650, instead of adding `withDefaultStyles`, can we add `unstyled` (with a default to `= false`), and also update the default of `style` to be `= defaultStyles`? (along with the export of `defaultStyles`) 

this would support the current behavior + `styles` by default, but allow you override all of (or a subset of) `defaultStyles`, or apply no `styles` whatsoever.

dennisja:
I have changed it to behave like how @hshoff suggested but there is an issue with the implementation. If a user accidentally uses the tooltip like 

```jsx
<Tooltip unstyled style={{...someStylesAUserWantsToApply}} />
```
The styles will not be applied

packages/vx-tooltip/package.json
```diff
@@ -30,7 +30,7 @@
   "dependencies": {
     "@types/classnames": "^2.2.9",
     "@types/react": "*",
-    "@vx/bounds": "0.0.195",
+    "@vx/bounds": "^0.0.195",
```
williaster, line 33: can you actually revert this? version bumps are managed by `lerna` upon release.

williaster:
@dennisja thanks for this contribution! I think that the case you call out is okay, `unstyled` seems like it should be self-explanatory and we can revisit if we get negative feedback about this 👍 

will try and get this out in the `0.0.196` release today.

williaster:
out in `0.0.196`

hellosmithy:
Just a note that this is a breaking change for anyone who was passing in a style prop to Tooltip. Might be worth updating the CHANGELOG to reflect that.

hshoff:
We’ll update the changelog. 

For now, work around is to import defaultStyles and extend it through style prop. 

```jsx
<Tooltip style={{ ...defaultStyle, style }} />
```

williaster:
Updated the [changelog](https://github.com/hshoff/vx/blob/master/CHANGELOG.md#v00196) and included a before/after 👍

PR #667, opened 2020-05-02
fix(vx-brush): Fix initialBrushPosition

williaster:
#### :bug: Bug Fix

This fixes some problems with the implementation of `initialBrushPosition` from #618 
- the type `BrushStartEnd` expected both `x` and `y` of `start`/`end` to be defined, which doesn't work if you have a `vertical-` or `horizontal-`only brush. Fixed to handle partials.
- the initial brush worked if you dragged in a new position, or extended a side, but not if you dragged the initial brush window. to fix this I updated some of the initial `state.start/end` logic which is used by `BrushSelection` upon moving a selection.

**Before**
broken-brush-move

**After**
fixed-brush-move

Also updates the demo to include an `initialBrushPosition` which should be useful for peeps
image

#### :house: Internal
Jest was failing locally to collect coverage due to a missing react component function name. I added one to fix locally, and see if it fixes anything on CI.

@hshoff @kristw 
cc @maryschmidt

### DIFF
--- a/packages/vx-brush/src/BaseBrush.tsx
+++ b/packages/vx-brush/src/BaseBrush.tsx
@@ -6,7 +6,7 @@ import Drag, { HandlerArgs as DragArgs } from '@vx/drag/lib/Drag';
 import BrushHandle from './BrushHandle';
 import BrushCorner from './BrushCorner';
 import BrushSelection from './BrushSelection';
-import { MarginShape, Point, BrushShape, BrushStartEnd, ResizeTriggerAreas } from './types';
+import { MarginShape, Point, BrushShape, ResizeTriggerAreas, PartialBrushStartEnd } from './types';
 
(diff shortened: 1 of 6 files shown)


packages/vx-brush/src/BaseBrush.tsx
```diff
@@ -185,36 +177,33 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
   };
 
   handleDragEnd = () => {
-    const { isBrushing } = this.state;
-    if (isBrushing) {
```
williaster, line 189: noting to document: 
by the time this function is called `state` has been updated `isBrushing` usually has been updated to `false` and so this feature didn't work as intended.

williaster:
[CI passed](https://travis-ci.org/github/hshoff/vx/builds/682185139) even tho it's not posting here

Issue #680, opened 2020-05-08
docs: fix doc-gen


williaster:
Automated doc-gen has been broken since the TS re-write, so the current `/docs` page cannot leverage the improved type annotations.

Additionally, the current flow requires manual generation and check-in of static html pages. Ideally we can make this all automated.

williaster:
Enumerating the packages to be updated here to track progress

- [ ] `@vx/annotation` 
- [ ] `@vx/axis` #681 
- [ ] `@vx/bounds` 
- [ ] `@vx/brush` 
- [ ] `@vx/chord` 
- [ ] `@vx/clip-path` 
- [ ] `@vx/curve` 
- [ ] `@vx/drag` 
- [ ] `@vx/event` 
- [ ] `@vx/geo` 
- [ ] `@vx/glyph` 
- [ ] `@vx/gradient` 
- [ ] `@vx/grid` 
- [ ] `@vx/group` 
- [ ] `@vx/heatmap` 
- [ ] `@vx/hierarchy` 
- [ ] `@vx/legend` 
- [ ] `@vx/marker` 
- [ ] `@vx/mock-data` 
- [ ] `@vx/network` 
- [ ] `@vx/pattern` 
- [ ] `@vx/point` 
- [ ] `@vx/responsive` 
- [ ] `@vx/scale` 
- [ ] `@vx/shape` 
- [ ] `@vx/stats`
- [ ] `@vx/text` 
- [ ] `@vx/threshold` 
- [ ] `@vx/tooltip` 
- [ ] `@vx/voronoi` 
- [ ] `@vx/zoom`

williaster:
closed by #681 🔥

williaster:
actually, opening this back up until I also get rid of the legacy scripts / static docgen.

PR #681, opened 2020-05-08
docs(react-docgen-typescript-loader): add proof-of-concept Axis doc page

williaster:
#### :memo: Documentation

This PR is the first of several to address #680, to fix auto-generation of package documentation to include our `TypeScript` annotations, and make the process more efficient. In this PR I've made several updates and created a proof-of-concept for the `Axis` docs. In subsequent PRs we can update the docs for other packages.

For documentation/others, the current flow for auto-doc generation is as follows:
- For a given package, the `doc` script is invoked, which runs a `bash` `node` script that reads in the files for the package and generates an `api.md` file
- in `package/docs/*`, the `api.md` file is concatenated with `description.md` and `install.md` files to create `docs.md`
- `docs.md` is used to create a static `.html` page that is rendered by the `next js` app under e.g., `/static/docs/vx-responsive.html`

This is decently complex/lots of pieces, requires a lot of manual check-in work which means docs can become stale, and the script has been broken since the `ts` rewrite. In this PR I update all of this to be in the `next js` app as `react` components. 

Specifically this PR makes the following changes:
- [one time]
  - updates the `next` `webpack` config to include the `react-docgen-typescript-loader` which annotates components with any defined types
  - creates a new `ApiTable` to render prop types + descriptions + default values (mirroring the old `api.md`)
  - creates a new `DocPage` component to render the docs for a given package
  - re-writes the `/docs` page to 
  - group packages functionally to make the `vx` ecosystem more understandable 
  - adds a brief overview of each package to make them more understandable
- [to be done for each package to be updated]
  - created a new `pages/docs/Axis` page using `DocPage` `ApiTable`
  - updates the static `.html` links in `/docs` to point to this new doc page
  - removes `docs` script, and old `vx-axis/docs/*.md` files

Overall the result is similar to before, but hopefully will be more manageable. If you have design feedback, or suggestions for package categorization labels lmk 🙏 

**Before**
image

image

**After**
image

image

@hshoff @kristw


package.json
```diff
@@ -38,12 +37,6 @@
     "prepare-release": "git checkout master && git pull --rebase origin master && yarn run docs && lerna updated",
     "release": "yarn run prepare-release && lerna publish --exact"
   },
-  "lint-staged": {
```
williaster, line 41: this was becoming annoying/very slow 🐌

package.json
```diff
@@ -62,11 +55,10 @@
     "fs-jetpack": "^1.3.0",
     "husky": "^3.0.0",
     "lerna": "^3.15.0",
-    "lint-staged": "^9.2.5",
     "marked": "^0.7.0",
     "raf": "^3.4.0",
     "react": "^15.0.0-0 || ^16.0.0-0",
-    "react-docgen": "^2.21.0",
+    "react-docgen": "^5.3.0",
```
williaster, line 61: I think we can remove this once all packages are converted

packages/vx-demo/src/components/ApiTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+        .prop {
+          border-top: 1px solid #eaeaea;
+          padding: 1em 0.5em;
+          line-height: 1.2em;
+          vertical-align: middle;
+        }
+        .title > :not(:last-child) {
+          margin-right: 12px;
+        }
+        .description > :global(p) {
```
williaster, line 88: `:global` is needed because `styled-jsx` can't figure out that `Markdown` creates a `p` so the style doesn't apply

packages/vx-demo/src/components/DocPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+          flex-direction: row;
+        }
+        .doc-nav {
+          margin-top: 40px;
+          margin-right: 5em;
+          width: 165px;
+        }
+        @media (max-width: 600px) {
+          .doc-container {
+            flex-direction: column-reverse;
```
williaster, line 32: this mostly mirrors the current layout/responsive behavior. noting that `sticky` is broken in the current docs/I couldn't get it to work here either so removed it

packages/vx-demo/src/components/Footer.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 export default () => (
   <div className="footer">
     <div>
-      <img src="static/favicon.ico" alt="vx" />
+      <img src="/static/favicon.ico" alt="vx" />
```
williaster, line 6: all of these 'static => `/static` changes are needed for correct URLs within the nested `next` pages like `pages/docs/Axis`

packages/vx-demo/src/components/PackageList.tsx
```diff
@@ -0,0 +1,344 @@
+import React from 'react';
+import { VxPackage } from '../types';
+
+export default function PackageList({
```
williaster, line 4: tried to use the same component for `/docs` and for the package side nav on the `/docs/axis` page

packages/vx-demo/src/components/PackageList.tsx
```diff
@@ -0,0 +1,344 @@
...
+              <a
+                href="/static/docs/vx-annotation.html"
+                className={emphasizePackage === 'annotation' && 'emphasize'}
+              >
+                @vx/annotation
+              </a>
+              {!compact && <p>Annotate elements of a chart</p>}
+            </li>
+            <li>
+              <a href="/docs/axis" className={emphasizePackage === 'axis' && 'emphasize'}>
```
williaster, line 30: noting the only one that goes to a new `/docs` page versus `.html`

packages/vx-demo/src/pages/docs/Axis.tsx
```diff
@@ -0,0 +1,56 @@
...
+          </a>
+        </li>
+        <li>
+          <a href={`#AxisTop`}>
+            <code>&lt;AxisTop /&gt;</code>
+          </a>
+        </li>
+      </ul>
+
+      <ApiTable anchorId="Axis" docgenInfo={Axis.__docgenInfo} />
```
williaster, line 48: gotta @ts-ignore these, not sure there's a clean way to avoid that 🤔 

also noting that we'll have to do this for each component we want to annotate, but I think that's okay, it's not a lot of work

williaster:
cleaned up the styles some
image

packages/vx-demo/src/pages/docs/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "no-underscore-dangle": "off"
```
williaster, line 3: will have to ignore any `Component.__docgenInfo` without this

PR #698, opened 2020-05-13
docs(vx-brush): update to docgen

williaster:
#### :memo: Documentation

Part of #680, updates `@vx/brush` docs to the new non-static docgen, and adds types to the `Brush` component to make this way more useful.

**After**
image

**Before**
image

@hshoff @kristw


packages/vx-brush/package.json
```diff
@@ -37,7 +37,6 @@
     "@vx/group": "0.0.196",
     "@vx/shape": "0.0.196",
     "classnames": "^2.2.5",
-    "prop-types": "^15.6.1",
-    "recompose": "^0.23.1"
```
hshoff, line 41: 🥳

PR #706, opened 2020-05-14
docs(vx-gradient): update to docgen

williaster:
#### :memo: Documentation

Part of #680, updates `@vx/gradient` docs to the new non-static docgen.

**After**
image
image

**Before**
image

@hshoff @kristw


packages/vx-gradient/src/gradients/GradientDarkgreenGreen.tsx
```diff
@@ -4,7 +4,7 @@ import LinearGradient, { LinearGradientProps } from './LinearGradient';
 /**
  * All props pass through to `<LinearGradient {...props} />`
  */
-export default function GradientDarkGreen({
```
hshoff, line 7: 😄

williaster: docgen barfs (not really, it just doesn't parse props correctly) if the component name + file name don't match 😅

Issue #691, opened 2020-05-12
TooltipWithBounds types do not have the new unstyled props


tbassetto:
The `<TooltipWithBounds />` component [redefines `TooltipProps`](https://github.com/hshoff/vx/blob/da9f09df06ca16e534d95027812e0a74f7488a14/packages/vx-tooltip/src/tooltips/TooltipWithBounds.tsx#L21) instead of inheriting it from [`<Toolip/>`](https://github.com/hshoff/vx/blob/da9f09df06ca16e534d95027812e0a74f7488a14/packages/vx-tooltip/src/tooltips/Tooltip.tsx#L4) and thus TypeScript is complaining that it doesn't recognize `unstyled`.

williaster:
Thanks for flagging this @tbassetto. I'm happy to review a PR if you/others are up for it. 

I think to fix it properly we'd need to not just inherit the `types` from `Tooltip`, but remove the `styles` set by `TooltipWithBounds` and instead pass `top` and `left` to `Tooltip` so it can handle the merging of `styles` or not applying them if `unstyled` is set.

PR #721, opened 2020-05-17
Fix tooltip with bounds

dennisja:
#### :boom: Breaking Changes

-

#### :rocket: Enhancements

- Add unstyled prop to TooltipWithbounds

#### :memo: Documentation

- Update Tooltips documentation

#### :bug: Bug Fix

- Make Typescript happy when the `unstyled` prop is passed to `TooltipWithBounds`

#### :house: Internal

-
Fixes https://github.com/hshoff/vx/issues/691

### DIFF
--- a/packages/vx-tooltip/Readme.md
+++ b/packages/vx-tooltip/Readme.md
@@ -86,7 +86,7 @@ component (i.e., ...restProps):
 | offsetRight | number | 10      | Vertical offset of the tooltip from the passed `top` value, functions as a vertical padding.                                                                                     |
 | style       | object | --      | Sets / overrides any styles on the tooltip container (including top and left)                                                                                                    |
 | children    | node   | --      | Sets the children of the tooltip, i.e., the actual content                                                                                                                       |
-| unstyled  | bool             | true    | Whether the tooltip should render with default styles or not                       |
+| unstyled  | bool             | true    | Whether the tooltip should use styles from the style prop or not              |
 
 Note that this component is positioned using a `transform`, so overriding `left` and `top` via
 styles may have no effect.
--- a/packages/vx-tooltip/Readme.md
+++ b/packages/vx-tooltip/Readme.md
@@ -69,7 +69,7 @@ accepts the following props, and will spread any additional props on the tooltip
 | className | string           | --      | Adds a class (in addition to `vx-tooltip-portal`) to the tooltip container    |
 | style     | object           | --      | Sets / overrides any styles on the tooltip container (including top and left) |
 | children  | node             | --      | Sets the children of the tooltip, i.e., the actual content                    |
-| unstyled  | bool             | true    | Whether the tooltip use styles from the style prop or not                     |
+| unstyled  | bool             | true    | Whether the tooltip should use styles from the style prop or not              |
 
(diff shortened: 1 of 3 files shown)


packages/vx-tooltip/src/tooltips/TooltipWithBounds.tsx
```diff
@@ -64,7 +57,12 @@ function TooltipWithBounds({
 
   return (
     <Tooltip
-      style={{ top: 0, transform: `translate(${left}px, ${top}px)`, ...style }}
+      style={{
+        top: 0,
+        transform: `translate(${left}px, ${top}px)`,
+        ...style,
```
williaster, line 63: I think for API consistency between the two components this should probably be `...(!unstyled && style)`

packages/vx-tooltip/Readme.md
```diff
@@ -86,6 +86,7 @@ component (i.e., ...restProps):
 | offsetRight | number | 10      | Vertical offset of the tooltip from the passed `top` value, functions as a vertical padding.                                                                                     |
 | style       | object | --      | Sets / overrides any styles on the tooltip container (including top and left)                                                                                                    |
 | children    | node   | --      | Sets the children of the tooltip, i.e., the actual content                                                                                                                       |
+| unstyled  | bool             | true    | Whether the tooltip should render with default styles or not                       |
```
williaster, line 89: ditto I think this should probably match above.

dennisja:
@williaster What do you think about adding a warning when a user passes both the `unstyled` and `style` props to any of the `Tooltip` components?

williaster:
@dennisja sorry for the delay. We don't have a precedent anywhere in `vx` for issuing warnings like this. I think we should see if users find it confusing/open issues about it, and if so we can worry about it then.

packages/vx-tooltip/test/TooltipWithBounds.test.tsx
```diff
@@ -1,7 +1,25 @@
-import { TooltipWithBounds } from '../src';
+import React from 'react';
+import { mount } from 'enzyme';
```
williaster, line 2: ideally we would use `shallow` over `mount`, is that possible?

dennisja: For some reason using `shallow` fails. I first suspected the fact that `TooltipWithBounds` is wrapped by a higher-order component that uses life cycle methods to be the reason but even when I use `shallow` and enable life cycle methods it still fails. 
I have seen an issue pointing out that it might be a problem with HOCs but I'm tired now. I Will read through it tomorrow and see whether I can switch mount to shallow. Its been long ever since I used Enzyme

dennisja: @williaster I found a workaround to use `shallow` over `mount`. Let me know what you think about it

williaster: this is perfect! yeah usually you have to use `.dive()` to traverse as needed to whatever selection you are looking for. Thanks for making the change! 🙏

williaster:
Great work @dennisja , thanks for the bug fix + enhancement!

PR #725, opened 2020-05-20
fix getting coords from touch event for y typo

vodoleev-vladislav:
#### :bug: Bug Fix
Zoom component's zoom.dragMove worked incorrectly on touchmove event, moving only alongside 135 degree angle. Fixing the typo fixes this problem.


hshoff:
Thanks for the bug fix!

PR #732, opened 2020-05-28
docs(Gallery): add package filters

williaster:
#### :memo: Documentation

This PR adds the ability to filter the `Gallery` by a specific `vx` package. 

I went back and forth on auto-generating the `example <> vx dep` map with a script but think it's probably easier to just manually import them all, no worse than an `index` file but let me know if you have thoughts. 

I also wish there was an easy way to re-use the `PackageList` component we use in `docs/` but the items are all `<Links />`s (I updated them from `<a />`s to `<Link />`s here which fixed an issue I'd seen `np-progress` loader), and we need `<button />`s with `onClick` handlers. An alternative would be to push the `vxPackageFilter` state to the url and try to use `<Link />`s 🤔  The way that the mapping is built now also ensures that we only show packages for which there are examples (not all packages have examples right now) ... this is both good and bad since someone could interpret the list as comprehensive.

Overall I think it's a good add, tho.

gallery-package-filters

**Responsive view**

@hshoff @kristw

### DIFF
--- a/packages/vx-demo/src/components/PackageList.tsx
+++ b/packages/vx-demo/src/components/PackageList.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import Link from 'next/link';
 import cx from 'classnames';
 import { VxPackage } from '../types';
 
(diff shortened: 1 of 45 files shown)


packages/vx-demo/src/components/PackageList.tsx
```diff
@@ -19,43 +20,67 @@ export default function PackageList({
           <HeaderElement>Chart primitives</HeaderElement>
           <ul>
             <li className={cx(emphasizePackage === 'annotation' && 'emphasize')}>
-              <a href="/docs/annotation">@vx/annotation</a>
+              <Link href="/docs/annotation">
```
williaster, line 23: all these are the same, just wrapped `<a />`s in `<Link />`s (without the `<a />`s, `next` was complaining about direct string children being deprecated and suggested adding `<a />`s as a child of `<Link />`)

PR #736, opened 2020-06-02
[grid] add gridLineProps to style grids per tick value

jithureddy:
#### :rocket: Enhancements

- Add gridLineProps function for GridRow & GridColumn. This will allow styling the grid line based on the tick value

### DIFF
--- a/packages/vx-demo/src/pages/docs/Grid.tsx
+++ b/packages/vx-demo/src/pages/docs/Grid.tsx
@@ -7,10 +7,11 @@ import DocPage from '../../components/DocPage';
 import AxisTile from '../../components/Gallery/AxisTile';
 import BarStackTile from '../../components/Gallery/BarStackTile';
 import ThresholdTile from '../../components/Gallery/ThresholdTile';
+import GridTile from '../../components/Gallery/GridTile';
 
 const components = [GridRows, GridColumns, Grid];
 
-const examples = [AxisTile, BarStackTile, ThresholdTile];
+const examples = [AxisTile, BarStackTile, ThresholdTile, GridTile];
 
 export default () => (
   <DocPage components={components} examples={examples} readme={GridReadme} vxPackage="grid" />
--- a/packages/vx-grid/src/grids/Grid.tsx
+++ b/packages/vx-grid/src/grids/Grid.tsx
@@ -1,8 +1,11 @@
 import React from 'react';
 import cx from 'classnames';
 import { Group } from '@vx/group';
-import GridRows, { AllGridRowsProps } from './GridRows';
-import GridColumns, { AllGridColumnProps } from './GridColumns';
+import GridRows, { AllGridRowsProps, GridRowLineProps } from './GridRows';
+import GridColumns, {
+  AllGridColumnProps,
+  GridColumnLineProps,
+} from './GridColumns';
 import { Scale, CommonGridProps } from '../types';
 
 type CommonPropsToOmit =
@@ -39,6 +42,10 @@ export type GridProps<XScaleInput, YScaleInput> = Omit<
     rowTickValues?: CommonGridProps['tickValues'];
     /** Exact values to be used for GridColumns lines, passed to xScale. Use this if you need precise control over GridColumns values.  */
     columnTickValues?: CommonGridProps['tickValues'];
+    /** A function that returns props for grid row line */
+    gridRowLineProps?: GridRowLineProps<YScaleInput>;
+    /** A function that returns props for grid column line */
+    gridColumnLineProps?: GridColumnLineProps<XScaleInput>;
   };
 
 export default function Grid<XScaleInput, YScaleInput>({
@@ -60,6 +67,8 @@ export default function Grid<XScaleInput, YScaleInput>({
   yOffset,
   rowTickValues,
   columnTickValues,
+  gridRowLineProps,
+  gridColumnLineProps,
   ...restProps
 }: GridProps<XScaleInput, YScaleInput>) {
   return (
@@ -75,6 +84,7 @@ export default function Grid<XScaleInput, YScaleInput>({
         lineStyle={rowLineStyle}
         offset={yOffset}
         tickValues={rowTickValues}
+        gridRowLineProps={gridRowLineProps}
         {...restProps}
       />
       <GridColumns<XScaleInput>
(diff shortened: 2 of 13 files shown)


packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -5,11 +5,18 @@ import { Group } from '@vx/group';
...
+  tick: ScaleInput,
+  index: number
+) => React.CSSProperties;
+
 export type GridColumnProps<ScaleInput> = CommonGridProps & {
   /** `@vx/scale` or `d3-scale` object used to map from ScaleInput to x-coordinates. */
   scale: Scale<ScaleInput, number>;
   /** Total height of the each grid column line. */
   height: number;
+  /**A function that returns props for grid column line */
```
williaster, line 18: nit: missing space after `/**` and missing period.

packages/vx-demo/src/pages/docs/Grid.tsx
```diff
@@ -7,10 +7,11 @@ import DocPage from '../../components/DocPage';
 import AxisTile from '../../components/Gallery/AxisTile';
 import BarStackTile from '../../components/Gallery/BarStackTile';
 import ThresholdTile from '../../components/Gallery/ThresholdTile';
+import GridTile from '../../components/Gallery/GridTile';
 
 const components = [GridRows, GridColumns, Grid];
 
-const examples = [AxisTile, BarStackTile, ThresholdTile];
+const examples = [AxisTile, BarStackTile, ThresholdTile, GridTile];
```
williaster, line 14: Instead of adding a whole new tile for this, could we instead demo the functionality in an existing example like `/axis`?. posting a screenshot of the change would be ideal, too! ✨

packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -47,13 +58,18 @@ export default function GridColumns<ScaleInput>({
           x,
           y: height,
         });
+
+        const columnLineProps = gridColumnLineProps(d, i);
+        const lineStroke = columnLineProps.stroke || stroke;
```
williaster, line 63: currently you don't use any other line props that might be specified by `gridLineProps()`.

what if instead of extracting `lineStroke` and `lineWidth`, you keep them as they were and then simply spread the output of `gridLineProps` which could override `stroke/strokeWidth` or not: `<Line stroke={stroke} ... {...gridColumnLineProps(d, i, ticks)} />`

packages/vx-grid/src/grids/Grid.tsx
```diff
@@ -39,6 +42,10 @@ export type GridProps<XScaleInput, YScaleInput> = Omit<
     rowTickValues?: CommonGridProps['tickValues'];
     /** Exact values to be used for GridColumns lines, passed to xScale. Use this if you need precise control over GridColumns values.  */
     columnTickValues?: CommonGridProps['tickValues'];
+    /** A function that returns props for grid row line */
```
williaster, line 45: nit – no period

packages/vx-grid/src/grids/Grid.tsx
```diff
@@ -39,6 +42,10 @@ export type GridProps<XScaleInput, YScaleInput> = Omit<
...
+    gridRowLineProps?: GridRowLineProps<YScaleInput>;
+    /** A function that returns props for grid column line */
```
williaster, line 47: nit – no period

packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -5,11 +5,18 @@ import { Group } from '@vx/group';
 import { Point } from '@vx/point';
 import { Scale, CommonGridProps } from '../types';
 
+export type GridColumnLineProps<ScaleInput> = (
+  tick: ScaleInput,
+  index: number
```
williaster, line 10: it could be useful to pass `ticks` as the third argument here so you can more easily know the exact number of ticks (say if you want to style the last one). we do this for [`tickFormat`](https://github.com/airbnb/visx/blob/master/packages/visx-axis/src/types.ts#L19) in `@visx/axis` and it's useful because `numTicks` is approximate / can be difficult to know final number of ticks ahead of time.

williaster: also, I'm wondering if we can simply add this to `CommonGridProps` as `gridLineProps`, and then in `Grid.tsx` we update the name to be specific to column / row (`gridColumnLineProps/gridRowLineProps`). This more closely matches what we do for other props in `Grid` and I think there's no reason it needs to be specific on the `GridRow/Column` components. It will require passing the `ScaleInput` generic to `CommonGridProps`, tho.

packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -31,6 +38,10 @@ export default function GridColumns<ScaleInput>({
   lineStyle,
   offset,
   tickValues,
+  gridColumnLineProps = (/** tickValue, index */) => ({
```
williaster, line 41: I think if you update to the spread logic I suggest below, you don't need a default for this prop (could use optional chaining on invocation `gridLineProps?.()`)

packages/vx-grid/src/grids/GridRows.tsx
```diff
@@ -31,6 +38,10 @@ export default function GridRows<ScaleInput>({
   lineStyle,
   offset,
   tickValues,
+  gridRowLineProps = (/** tickValue, index */) => ({
```
williaster, line 41: ditto don't need a default

packages/vx-grid/src/grids/GridRows.tsx
```diff
@@ -47,13 +58,17 @@ export default function GridRows<ScaleInput>({
           x: width,
           y,
         });
+        const rowLineProps = gridRowLineProps(d, i);
```
williaster, line 61: ditto spread suggestion / leaving `stroke/strokeWidth` as they were

williaster (commented):
Hey @jithureddy thanks for the contribution! Sorry you caught us in the middle of a repo rename so there are some conflicting files. Git should know which files moved tho so a rebase should work.

I think this will be a great feature, and left a handful of suggestions.

kachkaev:
I was solving this problem locally via a workaround and was glad to discover this PR @jithureddy! Would be interested in resolving git conflicts?

mindtraveller:
Have you got any difficulties in understanding the scope of the changes while reviewing this PR? This might simplify the review: https://viezly.com/change_requests/4140

williaster:
we might want to close this one and start a new PR as we went through a large rename (`vx => visx`) since this PR.

PR #740, opened 2020-06-03
[demo] bump deps, updates + fixes

hshoff:
#### :house: Internal

- deps(root): bump yarn.lock, add @types/webpack
- fix(vx-demo/package.json): lock next.js version
- fix(vx-demo): prettier config updated so lots of minor style updates
- fix(vx-demo/next.config.js): fix invalid webpack config error
- fix(vx-demo/pages): routes are case sensitive (`/Docs` => `/docs`)
- new(vx-demo/gallery): gallery filter persists on query param `?pkg` instead of local state


.gitignore
```diff
@@ -20,7 +20,6 @@ coverage/
 dist/
 esm/
 lib/
-public/
```
williaster, line 23: nice, have been wanting to fix the warning about `static` being deprecated

packages/vx-demo/src/components/Gallery/index.tsx
```diff
@@ -88,10 +91,14 @@ const tiles = [
 ];
 
 export default function Gallery() {
-  const [activePackageFilter, setActivePackageFilter] = useState<VxPackage | null>(null);
-  const filteredTiles = activePackageFilter
-    ? tiles.filter(Tile =>
-        exampleToVxDependencyLookup[Tile.packageJson.name]?.has(activePackageFilter),
+  const router = useRouter();
+  const { pkg } = router.query;
```
williaster, line 95: 🍾

packages/vx-demo/src/next.config.js
```diff
@@ -1,43 +1,38 @@
 /* eslint-disable no-undef */
-const withCss = require('@zeit/next-css');
```
williaster, line 2: 👍 I think initially I needed this because the `sandbox`s need css to work on codesandbox, but I think I moved these to the `index` (which does `ReactDOM` stuff) so they aren't referenced in `vx-demo` anymore

packages/vx-demo/src/pages/_app.tsx
```diff
@@ -0,0 +1,8 @@
+import { AppProps } from "next/app";
```
williaster, line 1: did you mean to add this file?

yarn.lock
```diff
@@ -3,143 +3,157 @@
...
+    "@beemo/driver-jest" "^1.2.1"
     airbnb-js-shims "^2.2.1"
     dom-storage "^2.1.0"
     jest "^25.1.0"
 
 "@airbnb/config-prettier@^2.0.4":
-  version "2.1.1"
-  resolved "https://registry.yarnpkg.com/@airbnb/config-prettier/-/config-prettier-2.1.1.tgz#cf18b45555f92309b470634da176ddc9ebb53e81"
-  integrity sha512-lt/q40KUWK3GdQEARGbJG5aCoPaFmlLCIY/9C/0dIYsCT/MmwGBQ6c5ez7JZ5dvWaGZb5Z0eNUnz44+LzHjq/Q==
+  version "2.1.3"
```
williaster, line 65: I'm surprised that a minor version change resulted in the `'` => `"` changes 🤔 do you think that's right?

also just saw that there are `3.0` versions of all of these, should probably upgrade sometime 🔜 .

hshoff: this is actually my fault. I had the Deno prettier config set as my default formatter on save 🤦

PR #750, opened 2020-06-17
fix(vx-mock-data/cityTemperature): use better date format

williaster:
#### :rocket: Enhancements

This updates the date format of the `@vx/mock-data` `cityTemperature` dataset from
`YYYYMMDD` => `YYYY-MM-DD`

The old format requires parsing with something like `d3-time-format`, whereas the new one can be parsed with `new Date(d)`. I think this is a win since it requires one few steps / dependencies for someone to use it.

**Testing**
- [x] CI
- [x] Updated demo examples which use this dataset, verified each one still works.

@hshoff @kristw


williaster:
#751 should fix the CI on this

PR #754, opened 2020-06-18
new(vx-responsive): add leading option to resize debounce

### LINKED ISSUE #753: withParentSize: add "leading" option to debounced resize handler
Hello 👋,

I noticed that the `withParentSize` HOC introduces an artificial 300ms delay when my component renders on first page load. When setting `debounceTime` manually to something lower like `10` this delay disappears.

While debouncing is obviously a great way to prevent a high frequency re-renders, I do feel user experience is negatively impacted if the chart takes 300ms longer to render than it normally would.

A solution would be to enable the `leading` option for the debounce call. That way the very first invocation will always trigger immediately.

See: https://lodash.com/docs/4.17.15#debounce

I'll gladly submit a PR for this if you're open to the idea

VX ❤️

Pringels:
#### :rocket: Enhancements

- add leading calls to debounced resize function in all @vx/responsive HOCs and components. This prevents the initial render from blocking, increasing perceived page load speed.

Fixes #753 

The gifs below represent a page refresh. I set the debounceTime to 1000ms for demonstration purposes. The **Before** image shows a 1 second delay before the chart gets rendered on reload. The **After** image immediately renders the component on reload.

**Before:**
vx-responsive-before
**After:**
vx-responsive-after

### DIFF
--- a/packages/vx-responsive/src/components/ParentSize.tsx
+++ b/packages/vx-responsive/src/components/ParentSize.tsx
@@ -66,7 +66,7 @@ export default class ParentSize extends React.Component<
 
   resize = debounce(({ width, height, top, left }: ParentSizeState) => {
     this.setState(() => ({ width, height, top, left }));
-  }, this.props.debounceTime);
+  }, this.props.debounceTime, { leading: true });
 
   setTarget = (ref: HTMLDivElement | null) => {
     this.target = ref;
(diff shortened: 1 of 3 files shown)


williaster:
@Pringels after discussing more with @hshoff, and looking at the rendering implications (this will add an additional render which is likely fine), we think this probably should be configurable via `props`, could you make that change?

You can test the difference in behavior in [this sandbox](https://codepen.io/dcorb/pen/KVxGqN)
image

Pringels:
@williaster I added the additionial key to `props`. I set the default value to `true`, as I suspect this would be the desired behaviour in most cases, but if someone really needs to fine-tune performance they can always opt out.

williaster:
Awesome, thanks @Pringels 🙏 This should be out in the `0.0.198` release 🔜 .

williaster:
(marking this as breaking for the release since it will add new behavior, still below `1.0.0` so will be `0.0.198` as mentioned above)

PR #755, opened 2020-06-18
new(vx-tooltip): add Portal + demo

williaster:
#### :memo: Documentation

This PR
- adds a new `Portal` component to `@vx/tooltip`
- adds ability to set a default tooltip state in `useTooltip`
- sets `left: 0` by default on `TooltipWithBounds` which was needed for correct positioning within a `Portal`
- adds a new gallery demo `/tooltip`, which demos the boundary detection + `Portal` usage
  - required adding `react-use-measure` dependency in `vx-demo`)
  - added to `Gallery`, and to `@vx/tooltip` examples

Testing
- tested example on mobile screen size / with touch events
- tooltip position is correct in demo, in gallery, and in `@vx/tooltip` docs; note: portal does require additional calculation complexity, but it's not too bad. Hopefully the demo will be good for showing people how to do this.

**Without Portal**

**With Portal**

**Gif**
tooltip-portal

**In docs**
image

**In gallery**
image

@hshoff @kristw @techniq


packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+interface TooltipData {
+  text: string;
+  containerX: number;
+  containerY: number;
+  pageX: number;
+  pageY: number;
+}
+
+let tooltipTimeout: number;
```
williaster, line 19: oops, I can remove this now

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -0,0 +1,248 @@
...
+            ? 'I detect my container boundary'
+            : 'I will get clipped by my container',
+        },
+      });
+    },
+    [showTooltip, ownBounds, detectBounds, renderInPortal],
+  );
+
+  const TooltipComponent = detectBounds ? TooltipWithBounds : Tooltip;
+  const PortalWrapper = renderInPortal ? Portal : React.Fragment;
```
williaster, line 87: maybe `TooltipWrapper` instead

packages/vx-tooltip/src/tooltips/TooltipWithBounds.tsx
```diff
@@ -1,23 +1,8 @@
 import React from 'react';
-import { withBoundingRects } from '@vx/bounds';
+import { withBoundingRects, WithBoundingRectsProps } from '@vx/bounds';
 
 import Tooltip, { TooltipProps, defaultStyles } from './Tooltip';
 
-type RectShape = {
```
williaster, line 6: these were written before the `@vx/bounds` types were added

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -0,0 +1,244 @@
...
+  backgroundColor: 'rgba(53,71,125,0.8)',
+  color: 'white',
+  width: 152,
+  height: 72,
+  padding: 12,
+};
+
+export default function Example({ width, height, showControls = true }: TooltipProps) {
+  // bounds of the container are needed to convert page coordinates to container coordinates
+  const [ref, ownBounds] = useMeasure({ scroll: true });
```
ktmud, line 32: Do end users have to import `useMeasure` themselves and copy all these boilerplate code in order to use the portal mode?

Any plan to integrate `renderInPortal` as an option for `useTooltip`?

williaster: currently, yes, users must use `useMeasure`, but if you have thoughts on how to abstract this I haven't given it a lot of thought. a couple points:

- `vx` is low-level and the components are meant to be very generic. So I think some work on the user's end to use this is okay as we are _just_ adding the functionality, but hopefully we can make it a little easier.
  - also noting that I am adding this as part of the `XYChart` package, which will abstract everything for the developer

- It's a little tricky to abstract all of the coordinate logic because there are different use cases. For example, _this_ usage results in the `Tootlip` exactly following the mouse cursor which uses `pageX/Y`. But if you want the `Tootlip` to snap to data points you can't use `pageX`, you would have to
1) find the closest `datum` 
2) scale the `x` and `y` of the `datum`, which gives some `px` values in `svg` coordinates 
3) convert the `svg` coordinates  to `page` coordinates for use in the tooltip

- `useTooltip` is very generic and only holds `Tooltip` state; it doesn't do any rendering so it can't handle the `Portal` wrapping on its own. `Tooltip` / `TooltipWithBounds` might be a better candidate as it actually does rendering (see next point)

- perhaps we could create something like `const [coordinateRef, Tooltip] = useTooltipInPortal()` where you could pass `ref={coordinateRef}` to whatever container that will match the coordinates you pass to `<Tooltip left={..} top={..}>{content}</Tooltip>` (this would support following the cursor position, or adding extra logic like snapping to data points). `Tooltip` could then handle conversion of those coordinates to page coordinates using the `bounds` of `coordinateRef` (using `useMeasure` behind the scenes)
 




However, I am adding this for use in I am adding thi

ktmud: Thanks for the additional context.

I don't have a clear idea who are the intended users for the different levels of abstractions, but it simply feels weird to me that we are adding a `Portal` component to `vx/tooltip` but still left a good amount of heavy lifting related to portal outside of the package itself.

`coordinateRef` sounds cool. Haven't dug into it too much, but I'm wondering whether it's possible to do everything in just one hook. Something like this:

```tsx
function useTooltip({
  hideDelay = 300,
  xOffset = 0,
  yOffset = 0,
  portalContainer = document.body,
  getCoordinate,
}) {
  const [measureRef, bounds, forceRefresh] = useMeasure({ scroll: true });
  const tooltipElement = useRef(null);
  const tooltipTimeout = useRef(null);

  const Tooltip = {
    showTooltip(datum, e) {
      const elem = tooltipElement.current;
      if (!elem) return;
      const pos = getCoordinate(datum, e);
      elem.style.left = `${bounds.x + xOffset + pos.x}px`;
      elem.style.right = `${bounds.y + yOffset + pos.y}px`;
    }
  }

  useLayoutEffect(() => {
    const elem = tooltipElement.current;
    if (!portalContainer || !elem) {
        return null;
    }
    const timeout = tooltipTimeout.current;
    portalContainer.appendChild(elem);
    return () => {
      if (timeout) clearTimeout(timeout);
      tooltipTimeout.current = setTimeout(() => {
        if (portalContainer.contains(elem)) {
          portalContainer.removeChild(elem);
        }
      }, hideDelay);
    }
  }, [portalContainer, hideDelay])

  return [measureRef, Tooltip];
}


const [coordinateRef, { showTooltip, hideToolTip }] = useTooltip({
  hideDelay: 300,
  xOffset: 10,
  yOffset: 20,
  getCoordinate: (datum, e) => {
    return {
      x: xScale(datum.x),
      y: yScale(datum.y)
    }
  }
});

<svg ref={coordinateRef}>
  {bars.map(bar => (
    <rect
      onMouseLeave={hideTooltip}
      onMouseMove={showTooltip.bind(null, bar)}
    />
 ))}
</svg>;

```

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -0,0 +1,244 @@
+import React, { useState, useCallback } from 'react';
+import useMeasure from 'react-use-measure';
+import { Tooltip, TooltipWithBounds, useTooltip, defaultStyles, Portal } from '@vx/tooltip/src';
```
williaster, line 3: reminder to update `src` => `lib`

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
williaster: Hmm, I think _not_ having the dev handle rendering of the `Tooltip` at all and having `showTooltip` do this feels a little _too_ magical ,and doesn't fit the declarative nature of `react` as nicely. Also not sure if specifying `portalContainer` will work with SSR since `document` doesn't exist 🤔 

I also don't think we should couple the current `useTooltip` hook (which manages tooltip **_state_**, not rendering) with the `Portal` **_rendering_** logic (which shouldn't care about state) because a consumer might want to use the two separately (e.g., in XYChart I want to get `tooltipData` from `Context`, not the hook).

Working on a separate PR to add the hook I outlined above 🍾

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
williaster: this is fixed in #756 which set to merge into this branch

packages/vx-tooltip/src/Portal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children: NonNullable<React.ReactNode>;
+};
+
+/** Render within a portal using a declarative component API. */
+export default class Portal extends React.PureComponent<PortalProps> {
+  private node?: HTMLDivElement;
+
+  componentWillUnmount() {
+    if (this.node && document.body) {
+      document.body.removeChild(this.node);
```
kristw, line 17: can also do `this.node.remove()`

packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -0,0 +1,244 @@
...
+              className="crosshair horizontal"
+              style={{ transform: `translateY(${tooltipData?.containerY ?? 0}px)` }}
+            />
+            <div
+              className="crosshair vertical"
+              style={{ transform: `translateX(${tooltipData?.containerX ?? 0}px)` }}
+            />
+            <TooltipWrapper>
+              <TooltipComponent
+                key={Math.random()} // needed for bounds to update correctly
```
hshoff, line 116: > Keys should be stable, predictable, and unique. Unstable keys (like those produced by Math.random()) will cause many component instances and DOM nodes to be unnecessarily recreated, which can cause performance degradation and lost state in child components.
https://reactjs.org/docs/reconciliation.html#trade-offs

Does it have to be recreated all the time?

williaster: yeah I think unfortunately for now, we do this in most of our demos. This forces a re-mount (😱) to force-update the bounds the component has, which basically change on every mouse move. Without this, if you start the tooltip on the left of the chart and move it to the right edge, it still thinks it has room because it's using the initial bounds. 

This really can't be that great for perf because of the remount and because [calling `.getBoundingClientRect()` forces layout/reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#getting-box-metrics). I tried to switch to `use-measure` for the boundary detection, but it only uses `scroll` listeners + `ResizeObserver`, so it doesn't re-compute on mouse move.

williaster: left [a comment on `react-use-measure`](https://github.com/react-spring/react-use-measure/issues/14#issuecomment-649138515) about it cc @techniq

techniq: `IntersectionObserver` seems promising.  If `react-use-measure` doesn't add something, we might be able to only update the `key` when the observer fires... to force an update.

I initially thought we could pre-compute the bounds based on the parent and tooltip size, but sinse the tooltip size can change size based on the contents, this wouldn't work.

williaster: > I initially thought we could pre-compute the bounds based on the parent and tooltip size, but since the tooltip size can change size based on the contents, this wouldn't work.

ah this is interesting, currently we use the `bounds.left/top` attributes of the tooltip, rather than `bounds.width/height` (which should be more stable). I took a shot at updating to use the latter and not updating `key`, but hit issues in the portal where `width/height` of the parent can be `0` inside a portal 😭

williaster:
gonna merge this to wrap it up, happy to address any concerns in a followup PR

PR #756, opened 2020-06-19
new(vx-tooltip): add useTooltipInPortal

williaster:
#### :memo: Documentation

This PR is into #755 and 
- adds a new `userTooltipInPortal` hook to addresses the issue of `Portal` not being easy to use by lib consumers. Notes: 
  - this adds a new [`react-use-measure`](https://github.com/react-spring/react-use-measure) dependency for the robust bounds logic
  - `react-use-measure` relies on `ResizeObserver`. rather than adding the polyfill here like we do in `@vx/responsive`, I'm opting to just expose the `useMeasure` polyfill injection system. So, users can add it themselves to `window` or pass it in directly to the hook. I've noted this in the `Readme`.
- updates the `tooltip` demo added in #755 to use this hook instead of `Portal` + `react-use-measure`
- updates the `vx-tooltip` Readme to include `useTooltipInPortal` and `Portal`
- updates the `barstack` example to use `useTooltipInPortal` as a validation point within `svg`s
- adds `offsetLeft/Top` to `TooltipProps`, making `TooltipProps === TooltipWithBoundsProps`. This simplifies type checking in the new `useTooltipInPortal`, and is a breaking change for `Tooltip` as it will add `10px` of padding to the `Tooltip` left/top.

Using `userTooltipInPortal`
image

TODO
- [x] remove `react-use-measure` dep in `vx-demo`

@hshoff @kristw @ktmud @techniq

### DIFF
--- a/packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
+++ b/packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
@@ -1,6 +1,11 @@
 import React, { useState, useCallback } from 'react';
-import useMeasure from 'react-use-measure';
-import { Tooltip, TooltipWithBounds, useTooltip, defaultStyles, Portal } from '@vx/tooltip/src';
+import {
+  Tooltip,
+  TooltipWithBounds,
+  useTooltip,
+  useTooltipInPortal,
+  defaultStyles,
+} from '@vx/tooltip';
 
 export type TooltipProps = {
   width: number;
@@ -8,13 +13,7 @@ export type TooltipProps = {
   showControls?: boolean;
 };
 
-interface TooltipData {
-  text: string;
-  containerX: number;
-  containerY: number;
-  pageX: number;
-  pageY: number;
-}
+type TooltipData = string;
 
 const positionIndicatorSize = 8;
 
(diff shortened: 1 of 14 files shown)


packages/vx-demo/src/sandboxes/vx-tooltip/Example.tsx
```diff
@@ -96,36 +88,30 @@ export default function Example({ width, height, showControls = true }: TooltipP
...
             />
             <div
               className="crosshair vertical"
-              style={{ transform: `translateX(${tooltipData?.containerX ?? 0}px)` }}
+              style={{ transform: `translateX(${tooltipLeft}px)` }}
             />
-            <TooltipWrapper>
-              <TooltipComponent
-                key={Math.random()} // needed for bounds to update correctly
-                left={(tooltipLeft ?? 0) + (detectBounds ? 0 : 10)}
```
williaster, line 117: noting that we got to remove this conditional `10` by making the `Tooltip` APIs consistent.

techniq:
LGTM.

I recently started using `react-use-measure` to calculate some offsets when using `position: sticky` for table headers.  I also created a `Measure` component within our application to simplify usage, as well as handle scrolling offsets (`viewport` prop).

Should we consider using `react-use-measure` within `ParentSize` as well (I had considered using this `Measure` component in place of `ParentSize` within our components but hadn't yet.

I know we just added support for a `leading` debounce trigger and currently `react-use-measure` does not have this, but it [looks](https://github.com/react-spring/react-use-measure/blob/master/src/web/index.ts#L2) like they use [debounce](https://github.com/component/debounce#debouncefn-wait--immediate--false-) and it has support for firing on the leading edge (`immediate`)

```tsx
import React from 'react';
import useMeasure, { Options, RectReadOnly } from 'react-use-measure';

type MeasureProps = Options & {
  children: (bounds: RectReadOnly) => React.ReactNode;
  viewport?: boolean;
};

function Measure(props: MeasureProps) {
  const { children, viewport, ...options } = props;

  let [ref, bounds] = useMeasure({ debounce: 300, ...options });

  if (viewport) {
    // See note about scrolling position: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect
    bounds = {
      ...bounds,
      left: bounds.left + window.scrollX,
      top: bounds.top + window.scrollY,
      x: bounds.x + window.scrollX,
      y: bounds.y + window.scrollY
    };
  }

  return <div ref={ref}>{children(bounds)}</div>;
}

export default Measure;
```

williaster:
yeah I think updating `ParentSize` to use `react-use-measure` could be a good call, it would enable polyfill injection + fix scroll bounding box issues which we don't currently correct for 👍

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Hook that handles rendering of a Tooltip or TooltipWithBounds in a Portal.
+ * Handles conversion of container coordinates to page coordinates using the container bounds.
+ */
+export default function useTooltipInPortal({
+  scroll,
+  detectBounds = true,
+}: UseTooltipPortalParams | undefined = {}): UseTooltipInPortal {
+  const [containerRef, containerBounds] = useMeasure({ scroll });
```
kristw, line 38: Only `scroll` is passed to `useMeasure` but the type above also defines `debounce`, `detectBounds` and `polyfill`

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export type UseTooltipInPortal = {
+  containerRef: (element: HTMLElement | SVGElement | null) => void;
+  containerBounds: RectReadOnly;
+  TooltipInPortal: React.FC<TooltipProps>;
+};
+
+export type UseTooltipPortalParams = {
+  /** Whether TooltipWithBounds should be used to auto-detect its container boundaries and update its position accordingly. */
+  detectBounds?: boolean;
```
kristw, line 16: Isn't this the same with the `detectBounds` in `UseMeasureOptions`?

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
```diff
@@ -0,0 +1,57 @@
...
+} & UseMeasureOptions;
+
+export type UseMeasureOptions = {
```
kristw, line 19: Perhaps this should be defined above `UseTooltipPortalParams`

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      const portalLeft = containerLeft + (containerBounds.left || 0) + window.scrollX;
+      const portalTop = containerTop + (containerBounds.top || 0) + window.scrollY;
+
+      return (
+        <Portal>
+          <TooltipComponent left={portalLeft} top={portalTop} {...tooltipProps} />
+        </Portal>
+      );
+    },
+    [detectBounds, containerBounds.left, containerBounds.top],
```
kristw, line 53: This could also use `useCallback` instead of `useMemo`.

> useCallback(fn, deps) is equivalent to useMemo(() => fn, deps).

https://reactjs.org/docs/hooks-reference.html#usecallback

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
williaster: 🙏 good catch. I added all the `useMeasure` hooks to the types at the end, but missed implementing it 🤦

packages/vx-tooltip/src/hooks/useTooltipInPortal.tsx
williaster: yeah I think the two are pretty interchangeable for this, `useCallback` is more clean/readable.

williaster: Ah, I remember why I did this now, `useCallback` has the wrong type to be considered a `React.FC`, I think anything goes in `useMemo`

williaster:
gonna merge this into the parent branch

PR #770, opened 2020-07-27
feat: add meta scale types and update vx/stats to use types from vx/scale

kristw:
#### :boom: Breaking Changes

None

#### :rocket: Enhancements

* Export a few more types from `@vx/scale`: 

```
TimeScaleType

ContinuousScaleType
DiscretizingScaleType
DiscreteScaleType

ContinuousDomainScaleType
```

#### :memo: Documentation

a bit more comment

#### :bug: Bug Fix

None

#### :house: Internal

Update `vx/stats` to use types from `vx/scale`

### DIFF
--- a/packages/vx-stats/src/BoxPlot.tsx
+++ b/packages/vx-stats/src/BoxPlot.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import classnames from 'classnames';
 import { Group } from '@vx/group';
-import { PickD3Scale } from '@vx/scale';
+import { PickD3Scale, ContinuousDomainScaleType } from '@vx/scale';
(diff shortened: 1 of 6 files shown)


packages/vx-stats/src/BoxPlot.tsx
```diff
@@ -16,7 +17,7 @@ type ScaleInput = number;
 
 export type BoxPlotProps = SharedProps & {
   /** Scale for converting ScaleInput values to pixel offsets. */
-  valueScale: GenericScale<ScaleInput>;
+  valueScale: PickD3Scale<ContinuousDomainScaleType, number>;
```
williaster, line 20: should we either use `ScaleInput` here or just remove `ScaleInput` altogether / opt for `number`?

PR #773, opened 2020-07-30
feat: update vx/axis types 

kristw:
## :boom: Breaking Changes

- N/A

### vx/axis

- Consistent and compatible typings across `vx/scale` and `vx/axis`.
- More fields passed to child render props of `Axis`
- Also export types

### vx/scale

- A few new types: `AnyD3Scale`, `InferD3ScaleOutput`, `InferD3ScaleDiscreteInput`, `InferD3ScaleThresholdInput` and `ScaleInput`.
- A few new utilities functions: `getTicks`, `coerceNumber` and `toString`.

## :memo: Documentation

- More field documentation 

### vx/axis

- The `SharedAxisProps` type is now based on generic type `Scale` instead of `ScaleInput`. We can infer `ScaleInput` from `Scale` but not the other way around.
- Refactor `Axis`. Separate the logic that were used when a child renderer prop (`children`) was not provided into its own component `AxisRenderer`. This `AxisRenderer` will receive the exact same props with `children` so that enforces us to pass everything necessary.   
- Refactor `AxisRight`, `AxisLeft`, `AxisBottom`, `AxisTop` to only list modified/used fields and pass the rest of `props` through via `...`.
- Merge type `AxisOrientation` and constant `Orientation`. Also refer to it as `Orientation` consistently. (Replace all `ORIENT` and `AxisOrientation` with `Orientation`)
- Rename `center.ts` to `getTickPosition.ts`, `toNumberOrUndefined.ts` to `coerceNumber.ts`, `labelTransform.ts` to `getLabelTransform.ts`

## 🏋️  Tests

- Added unit tests and verified that the demo still look the same.


packages/vx-axis/src/types.ts
```diff
@@ -1,30 +1,37 @@
...
+import Orientation from './constants/orientation';
 
-export type AxisOrientation = 'top' | 'right' | 'bottom' | 'left';
+// In order to plot values on an axis, output of the scale must be number.
+// Some scales return undefined.
+export type AxisScaleOutput = number | NumberLike | undefined;
+
+/** A catch-all type for scales that are compatible with axis */
+export type AxisScale<Output extends AxisScaleOutput = AxisScaleOutput> =
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kristw, line 11: This has to be `any`. `unknown` does not work for this case.

packages/vx-axis/src/types.ts
```diff
@@ -1,30 +1,37 @@
...
 export type TickRendererProps = Partial<TextProps> & {
   x: number;
   y: number;
   formattedValue: FormattedValue;
 };
 
-export type SharedAxisProps<ScaleInput> = {
-  /** The class name applied to the outermost axis group element. */
-  axisClassName?: string;
+interface CommonProps<Scale extends AxisScale> {
```
kristw, line 26: This is a new interface that is the shared props between `Axis` and `AxisRenderer`

packages/vx-axis/src/types.ts
```diff
@@ -49,75 +54,59 @@ export type SharedAxisProps<ScaleInput> = {
...
-}
-
-export interface Point {
+interface Point {
   x: number;
   y: number;
 }
 
-export type ChildRenderProps<ScaleInput> = {
+export type AxisRendererProps<Scale extends AxisScale> = CommonProps<Scale> & {
```
kristw, line 76: Instead of accepting `ScaleInput` generic, accept `Scale` with the constraint that it must be a scale compatible for axis (compatible meaning scale output is `AxisScaleOutput = number | NumberLike | undefined` regardless of input type) then infer `ScaleInput` from `Scale`

packages/vx-scale/src/types/Scale.ts
```diff
@@ -62,3 +62,33 @@ export type D3Scale<
...
+ * unless other generic types (`Output`, `DiscreteInput` and `ThresholdInput`)
+ * are also included and passed to `D3Scale`.
+ * Otherwise it may not match some scales (band, point, threshold) correctly and cause TS errors.
+ *
+ * Example error messages:
+ * * "Type 'StringLike' is not assignable to type 'string'"
+ * * "Type 'number' is not assignable to type 'DefaultThresholdInput'"
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AnyD3Scale = D3Scale<any, any, any>;
```
kristw, line 79: This has to be `any`. Cannot use `unknown`.

packages/vx-axis/src/axis/AxisBottom.tsx
```diff
@@ -1,31 +1,12 @@
...
-import ORIENT from '../constants/orientation';
-import { SharedAxisProps } from '../types';
+import Orientation from '../constants/orientation';
+import { SharedAxisProps, AxisScale } from '../types';
 
-export type AxisBottomProps<ScaleInput> = SharedAxisProps<ScaleInput>;
-
-export default function AxisBottom<ScaleInput>({
-  children,
+export default function AxisBottom<Scale extends AxisScale>({
```
kristw, line 7: only destructure fields that are modified or used.

packages/vx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
```
kristw, line 1: Extracted from `Axis.tsx`

packages/vx-demo/src/sandboxes/vx-barstack/Example.tsx
```diff
@@ -161,7 +161,7 @@ export default function Example({
             }
           </BarStack>
         </Group>
-        <AxisBottom<string>
+        <AxisBottom
```
kristw, line 164: generic type can be inferred from the passed scale

packages/vx-axis/src/axis/Axis.tsx
```diff
@@ -1,216 +1,86 @@
...
 
-  const isLeft = orientation === ORIENT.left;
-  const isTop = orientation === ORIENT.top;
-  const axisIsHorizontal = isTop || orientation === ORIENT.bottom;
+  const tickPosition = getTickPosition(scale);
   const tickSign = isLeft || isTop ? -1 : 1;
 
-  const position = center(scale.copy());
-
-  const axisFromPoint = new Point({
```
kristw, line 82: I extracted the logic for flipping `x` and `y` into util function `createPoint`.

packages/vx-axis/src/utils/getTickPosition.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Create a function that returns a tick position for the given tick value
+ */
+export default function getTickPosition<Scale extends AxisScale>(scale: Scale) {
+  // Broaden type before using 'xxx' in s as typeguard.
+  const s = scale as AxisScale;
+
+  // For point or band scales,
+  // have to add offset to make the tick centered.
```
hshoff, line 12: Is there a use case for making this configurable?  `(start | center | end)`

kristw: Adding additional alignment would be nice addition. Will mod it.
(The original file was called `center`.)

PR #775, opened 2020-08-03
feat(grid): update scale types in vx/grid

kristw:
#### :boom: Breaking Changes

Similar to `@vx/axis` in #773 , `@vx/grid` components now accept D3 Scale as generic type instead of `ScaleInput`. Developers should not expect to specify this generic type as it can be inferred from the passed scale.

### DIFF
--- a/packages/vx-demo/src/sandboxes/vx-area/Example.tsx
+++ b/packages/vx-demo/src/sandboxes/vx-area/Example.tsx
@@ -109,7 +109,7 @@ export default withTooltip<AreaProps, TooltipData>(
           />
           <LinearGradient id="area-background-gradient" from={background} to={background2} />
           <LinearGradient id="area-gradient" from={accentColor} to={accentColor} toOpacity={0.1} />
-          <GridRows<number>
+          <GridRows
             scale={stockValueScale}
             width={xMax}
             strokeDasharray="3,3"
@@ -117,7 +117,7 @@ export default withTooltip<AreaProps, TooltipData>(
             strokeOpacity={0.3}
             pointerEvents="none"
           />
-          <GridColumns<Date>
+          <GridColumns
             scale={dateScale}
             height={yMax}
             strokeDasharray="3,3"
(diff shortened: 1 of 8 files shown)


packages/vx-demo/src/sandboxes/vx-area/Example.tsx
```diff
@@ -109,15 +109,15 @@ export default withTooltip<AreaProps, TooltipData>(
           />
           <LinearGradient id="area-background-gradient" from={background} to={background2} />
           <LinearGradient id="area-gradient" from={accentColor} to={accentColor} toOpacity={0.1} />
-          <GridRows<number>
```
kristw, line 112: The generic type can be inferred from the input scale and this is no longer needed.

packages/vx-grid/src/grids/GridRows.tsx
```diff
@@ -3,22 +3,25 @@ import cx from 'classnames';
...
-import { Scale, CommonGridProps } from '../types';
+import { getTicks, ScaleInput } from '@vx/scale';
+import { CommonGridProps, GridScale } from '../types';
 
-export type GridRowsProps<ScaleInput> = CommonGridProps & {
+export type GridRowsProps<Scale extends GridScale> = CommonGridProps & {
   /** `@vx/scale` or `d3-scale` object used to map from ScaleInput to y-coordinates. */
-  scale: Scale<ScaleInput, number>;
+  scale: Scale;
+  /** Exact values used to generate grid coordinates (y- for Rows, x- for Columns) lines using `scale`. Overrides `numTicks` if specified. */
```
williaster, line 12: should we remove the `(y- for Rows, x- for Columns)` piece here? also I think a leftover typo from me, could likely also remove the word `coordinates` here

packages/vx-grid/src/grids/GridColumns.tsx
```diff
@@ -3,22 +3,25 @@ import cx from 'classnames';
...
-import { Scale, CommonGridProps } from '../types';
+import { getTicks, ScaleInput } from '@vx/scale';
+import { CommonGridProps, GridScale } from '../types';
 
-export type GridColumnProps<ScaleInput> = CommonGridProps & {
+export type GridColumnProps<Scale extends GridScale> = CommonGridProps & {
   /** `@vx/scale` or `d3-scale` object used to map from ScaleInput to x-coordinates. */
-  scale: Scale<ScaleInput, number>;
+  scale: Scale;
+  /** Exact values used to generate grid coordinates (y- for Rows, x- for Columns) lines using `scale`. Overrides `numTicks` if specified. */
```
williaster, line 12: ditto re removing `y- ... and x-` from comment + removing `coordinates`

PR #778, opened 2020-08-12
new(vx-shape): add SplitLinePath

williaster:
#### :rocket: Enhancements

Digging this one up from over a year ago before I lose the branch:
- adds a `SplitLinePath` component to `@vx/shape` that allows you to create a line path split into multiple smaller line paths that can be styled independently
- adds a demo example

image
image

@hshoff @kristw


packages/vx-demo/src/sandboxes/exampleToVxDependencyLookup.ts
```diff
@@ -43,33 +44,34 @@ import { VxPackage } from '../types';
 const examples = [
   areaPackageJson,
   axisPackageJson,
-  bargroupPackageJson,
   bargroupHorizontalPackageJson,
+  bargroupPackageJson,
```
williaster, line 48: 💬 abc

packages/vx-shape/src/shapes/SplitLinePath/getLineSegments.ts
```diff
@@ -0,0 +1,69 @@
...
+      return coords;
+    });
+
+    return lineSegments;
+  } catch (e) {
+    return [];
+  }
+}
+
+export default memoize(
```
williaster, line 65: this is generally the approach that [we use in `@vx/text`](https://github.com/hshoff/vx/blob/master/packages/vx-text/src/util/getStringWidth.ts#L30)

williaster:
I can refactor it to `shapes/SplitLinePath`, I created `SplitLinePath/` to co-locate the `getLineSegments.ts` util with the component it's used with. But that's easy to move to `util/` which already exists in the package

packages/vx-shape/src/shapes/SplitLinePath/getLineSegments.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // create a single path element if not done already
+    if (!pathElement) {
+      const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+      svg.style.width = '0';
+      svg.style.height = '0';
+      svg.style.position = 'absolute';
+      svg.style.top = '-100%';
+      svg.style.left = '-100%';
+      pathElement = document.createElementNS('http://www.w3.org/2000/svg', 'path');
```
kristw, line 30: can extract the namespace url to constant

kristw: Perhaps add

```ts
// not visible
svg.style.opacity = 0;
// won't capture mouse events
svg.style.pointerEvents = 'none';
``` 

for extra precaution

packages/vx-shape/src/shapes/SplitLinePath/getLineSegments.ts
```diff
@@ -0,0 +1,69 @@
...
+    const pieceSize = totalPathLength / totalPieces;
+
+    let cumulativeSize = 0;
+
+    const lineSegments = segments.map(segment => {
+      const segmentPointCount = segment.length;
+      const coords: { x: number; y: number }[] = [];
+
+      for (let i = 0; i < segmentPointCount + sampleRate; i += sampleRate) {
+        const distance = cumulativeSize * pieceSize + i * pieceSize;
```
kristw, line 49: nit. one fewer multiplication

```ts
const distance = (cumulativeSize + i) * pieceSize
```

packages/vx-shape/src/shapes/SplitLinePath/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { CurveFactory } from 'd3-shape';
+import LinePath from '../LinePath';
+import getLineSegments from './getLineSegments';
+
+interface Point {
+  x: number;
+  y: number;
+}
+
+type SplitLinePathProps<Datum> = {
```
kristw, line 11: could remove `curve`, `defined`, `x` and `y` and use

```ts
type SplitLinePathProps<Datum> = {
  ...
} & LinePathConfig<Datum>;
```

packages/vx-shape/src/shapes/SplitLinePath/getLineSegments.ts
williaster: ^was wondering about this. we don't do this in `@vx/text` and I forget if there was a reason. will search for old PRs

PR #779, opened 2020-08-13
new(vx-axis): add AnimatedAxis

williaster:
#### :rocket: Enhancements

This PR adds a new `AnimatedAxis` component to `@vx/axis` and the needed `react-spring` dependency. The surface area is pretty minimal because only `ticks` are animated so we can fully re-use `Axis/AxisBottom/etc` and override a new `ticksComponent` prop which defaults to the old behavior.

I'm a little conflicted because generally `vx` is not opinionated on animation libraries and we don't want to bloat bundle sizes. However, the incentive for this comes from the `XYChart` POC in #745 where we would like to support animation if users opt in for it. So, we could either
a) simply add the needed `Axis` hooks (`ticksComponent`) to enable this cleanly, and keep `AnimatedAxis` in `XYChart`. or
b) add it to `@vx/axis` to let users opt into using it. It shouldn't affect bundle sizes if consumers use tree shaking or deep imports

To make b) a bit cleaner we could consider
- avoid exporting `AnimatedAxis` from the `index` and require a deep import
- add `react-spring` as a peer-dep if we want consumers to be able to use an existing version they have

#### Updated `/axis` demo

@hshoff @kristw @techniq


packages/vx-axis/package.json
```diff
@@ -36,10 +36,8 @@
     "@vx/scale": "0.0.198",
     "@vx/text": "0.0.198",
     "classnames": "^2.2.5",
-    "prop-types": "^15.6.0"
-  },
-  "devDependencies": {
-    "@vx/scale": "0.0.198"
```
williaster, line 42: this is now a `dep` so don't need to duplicate it

packages/vx-axis/src/axis/AnimatedTicksRenderer/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const transitionConfig = useTickTransitionConfig({ horizontal, scale });
+  const animatedTicks = useTransition(ticks, tick => `${tick.value}-${horizontal}`, {
+    unique: true,
+    ...transitionConfig,
+  });
+
+  return animatedTicks.map(({ item, key, props }, index) => {
+    // @ts-ignore react-spring types don't handle fromX, etc.
+    const { fromX, toX, fromY, toY, opacity } = props;
+    const tickLabelProps = allTickLabelProps[index] ?? allTickLabelProps[0] ?? {};
```
williaster, line 30: I couldn't come up with a great solution for this. when tick values update, there may not be a `tickLabelProps` for that index.

ultimately the weirdness comes from the `Axis` component trying to find the max tick label font size to correctly offset the axis label (will comment on this). maybe we should deprecate that logic and clean this up?

packages/vx-axis/src/axis/Axis.tsx
```diff
@@ -46,20 +46,21 @@ export default function Axis<Scale extends AxisScale>({
...
-    });
+  const ticks = filteredTickValues.map(({ value, index }) => {
+    const scaledValue = coerceNumber(tickPosition(value));
+
+    return {
+      value,
+      index,
+      from: createPoint({ x: scaledValue, y: 0 }, horizontal),
+      to: createPoint({ x: scaledValue, y: tickLength * tickSign }, horizontal),
+      formattedValue: format(value, index, filteredTickValues),
```
williaster, line 61: have been wanting to add this for a while (will update the PR description): we now pass all ticks as a third argument (common d3 pattern) so that users can determine if a tick is first/last, etc. with `index` only you couldn't ever determine this because you may not know the exact # ticks unless you specify `tickValues`

packages/vx-axis/src/axis/AxisBottom.tsx
```diff
@@ -4,16 +4,19 @@ import Axis from './Axis';
 import Orientation from '../constants/orientation';
 import { SharedAxisProps, AxisScale } from '../types';
 
-export default function AxisBottom<Scale extends AxisScale>({
-  axisClassName,
-  labelOffset = 8,
-  tickLabelProps = (/** tickValue, index */) => ({
+export const bottomTickLabelProps = (/** tickValue, index */) =>
```
williaster, line 7: exporting these so that they could be used outside the component. tick label props are the primary differences between axis configurations

packages/vx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -30,51 +29,36 @@ export default function AxisRenderer<Scale extends AxisScale>({
...
   tickLength,
   tickStroke = '#222',
   tickTransform,
   ticks,
+  ticksComponent = TicksRenderer,
 }: AxisRendererProps<Scale>) {
-  let tickLabelFontSize = 10; // track the max tick label size to compute label offset
-
+  // compute the max tick label size to compute label offset
+  const allTickLabelProps = ticks.map(({ value, index }) => tickLabelProps(value, index));
```
williaster, line 44: this is the jank logic I alluded to above. The label and ticks were previously coupled with this, so I updated the logic to pass an array of `tickLabelProps` to `ticksComponent`. 

Might be worth a breaking change to remove this?

packages/vx-demo/src/sandboxes/vx-axis/Example.tsx
```diff
@@ -87,76 +110,80 @@ export default function Example({ width: outerWidth = 800, height: outerHeight =
...
+                xScale={scale}
+                yScale={yScale}
+                stroke={gridColor}
+                width={width}
+                height={scaleHeight}
+                numTicksRows={2}
+                numTicksColumns={numTickColumns}
+              />
+              <AnimatedAxis
+                orientation={Orientation.bottom}
```
williaster, line 155: this is all just indented and we use `AnimatedAxis` + specify `orientation`

hshoff:
I'd like to propose option **c):** create a new package `@vx/react-spring` that includes `react-spring` as a peerDep and can be a home for things that depend on `react-spring`?

williaster:
🤯 I dig it! Noting that I think a lot of the `XYChart` elements might end up living there like `AnimatedBars`, `AnimatedStackedBars`, etc. (likely animated equivalents of many things in `@vx/shape`). 

Organizing that over time might be tricky, but I think it'll be a good forcing function to keep things modular / might converge on good patterns over time.

williaster:
doc page for the new package
image

williaster:
@hshoff 🎉 
> Coverage increased (+2.2%) to 46.97%

I have an `AnimatedGrid` PR stacked on this, will add some more legit ones there

PR #782, opened 2020-08-18
new(vx-xychart): add package boilerplate

williaster:
#### :rocket: Enhancements

This adds the boilerplate code for a new `@vx/xychart` package as part of the [XYChart project](https://github.com/hshoff/vx/projects/3). 

The package is marked as `private` so shouldn't be publishable, and I created an empty demo page which isn't linked to so its partial implementation shouldn't affect the demo site.

image

@hshoff @kristw @techniq


packages/vx-demo/src/sandboxes/vx-xychart/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from 'react-dom';
+import ParentSize from '@vx/responsive/lib/components/ParentSize';
```
kristw, line 3: Can this import from just `@vx/responsive`?

packages/vx-xychart/Readme.md
```diff
@@ -0,0 +1,3 @@
+# @vx/xychart
```
kristw, line 1: nit: usually the `README.md` file name is ALL_CAPS?

packages/vx-demo/src/sandboxes/vx-xychart/index.tsx
williaster: yeah, but this is in the sandbox template / all of them are that way so would rather make a sweep to fix them all in the future.

PR #783, opened 2020-08-19
feat(vx-marker): add markers

hshoff:
Markers are pretty handy. You define them in [`<defs>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) and attach them to paths or shapes through [`marker-start`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/marker-start), [`marker-end`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/marker-end) and/or [`marker-mid`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/marker-mid) attributes.

This PR was inspired by Keynote when I was editing a presentation and saw the line endpoints editor:

#### :boom: Breaking Changes

- [marker] remove old [`<Marker />`](https://github.com/hshoff/vx/blob/v0.0.198/packages/vx-marker/src/markers/Marker.tsx#L72-L98) implementation of a Line and some Text

#### :rocket: Enhancements

- [marker] add new [`<Marker />`](https://github.com/hshoff/vx/compare/harry-svg-marker?expand=1#diff-7a829322360dc21940b608c760087e29R32-R42) that matches actual SVG [`<marker>`](https://github.com/hshoff/vx/blob/v0.0.198/packages/vx-marker/src/markers/Marker.tsx#L72-L98)
- [marker] add `<MarkerArrow />`, `<MarkerCross />`, `<MarkerX />`, `<MarkerCircle />`, `<MarkerLine />`

#### :memo: Documentation

- [demo] updated curves demo to use new `<Marker>`

#### :weight_lifting: Tests

- [marker] added tests

#### :camera_flash: Screenshots

image
image
image


packages/vx-marker/Readme.md
```diff
@@ -4,24 +4,8 @@
...
-  from={markerFrom}
-  to={markerTo}
-  stroke={'white'}
-  label={'deploy'}
-  labelStroke={'none'}
-  labelDx={6}
-  labelDy={15}
-/>
-```
+Markers are elements attached to shapes.
```
williaster, line 7: should this be "shapes attached to elements"?

packages/vx-marker/src/markers/Circle.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Marker
+      id={id}
+      markerWidth={size}
+      markerHeight={size}
+      refX={0}
+      refY={mid}
+      orient="auto-start-reverse"
+      markerUnits="strokeWidth"
+      fill="steelblue"
```
williaster, line 28: do you think the default color should be more neutral? or to be consistent with other `Marker`s we could remove this and depend on `restProps`?

packages/vx-marker/src/markers/Marker.tsx
```diff
@@ -1,100 +1,44 @@
...
-  /** x,y coordinates of the marker line end point. */
-  to?: {
-    x?: number;
-    y?: number;
-  };
-  /** stroke color of the marker line. */
-  stroke?: string;
-  /** strokeWidth of the marker line. */
+export interface Props {
+  /** Unique id for the gradient. Should be unique across all page elements. */
```
williaster, line 4: `gradient` => `marker`

packages/vx-marker/src/markers/Marker.tsx
```diff
@@ -1,100 +1,44 @@
...
-    y?: number;
-  };
-  /** x,y coordinates of the marker line end point. */
-  to?: {
-    x?: number;
-    y?: number;
-  };
-  /** stroke color of the marker line. */
-  stroke?: string;
-  /** strokeWidth of the marker line. */
```
williaster, line 23: should we add back the prop annotation for `strokeWidth`?

packages/vx-marker/src/markers/X.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Cross from './Cross';
+import { MarkerProps } from './Marker';
+
+export interface MarkerXProps {
+  id: string;
+  size?: number;
+  strokeWidth?: number;
```
williaster, line 8: could remove this since it is defined in `MarkerProps`

packages/vx-marker/src/markers/Line.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import Marker, { MarkerProps } from './Marker';
+
+export interface MarkerLineProps {
+  id: string;
+  size?: number;
+  stroke?: string;
```
williaster, line 7: do we need to defined `stroke` and `fill` since they are part of `MarkerProps`? 

could still keep the logic for pulling them out of `restProps` to ensure that `fill` is defined and `stroke="none"` in the component.

rayshan:
This is an awesome feature! What's the replacement for the old marker? Make my own line / text?

PR #784, opened 2020-08-19
test: increase code coverage

hshoff:
This takes vx from 26% to **44%** code coverage. More to come in future PRs...

#### :house: Internal

- [jest] ignore vx-demo, vx-vx code coverage
- [annotation] 100% coverage
- [bounds] 100% coverage


packages/vx-bounds/test/withBoundingRects.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    expect(RenderedComponent.prop('rect')).toEqual(undefined);
+    expect(RenderedComponent.prop('parentRect')).toEqual(undefined);
+  });
+
+  test('it should set rect and parentRect to empty state if no getBoundingClient()', () => {
+    const Component = () => <>''</>;
+    const HOC = withBoundingRects(Component);
+    const wrapper = mount(<HOC />);
+    const RenderedComponent = wrapper.find(Component);
+    expect(RenderedComponent.prop('rect')).toEqual(emptyRect);
```
hshoff, line 78: 🤔 wasn't sure how to mock the RenderComponent.node.parentNode to not have `getBoundingRect` which triggers the `parentRect = emptyRect` branch... so skipped it for now.

package.json
```diff
@@ -91,6 +91,19 @@
       "coverageReporters": [
         "json"
       ],
+      "collectCoverageFrom": [
+        "packages/src/**/*{.ts,tsx}",
```
williaster, line 95: should this be `packages/src/**/*.{ts,tsx}`? (period outside the brackets)

might be worth looking at the generated `jest.config.js`, locally mine already looks like this so some of this may be duplicated. might be worth ignoring `test/**/*.{ts,tsx}` also?

```
  ...
  "collectCoverageFrom": [
    "**/src/**/*.{ts,tsx,js,jsx}",
    "**/test/**/*.{ts,tsx,js,jsx}"
  ],
  "coverageDirectory": "./coverage",
  "coveragePathIgnorePatterns": [
    "coverage/",
    "node_modules/",
    "public/",
    "esm/",
    "lib/",
    "tmp/",
    "dist/"
  ],
  ...
```

packages/vx-bounds/test/withBoundingRects.test.ts
```diff
@@ -1,54 +0,0 @@
-import { withBoundingRects } from '../src';
-
-// const expectedRectShape = expect.objectContaining({
```
williaster, line 3: didn't know this was all commented out 😮

packages/vx-annotation/test/LinePathAnnotation.test.tsx
```diff
@@ -65,4 +65,15 @@ describe('<LinePathAnnotation />', () => {
       ),
     ).toBe(true);
   });
+
+  test('it should have default x and y accessors', () => {
```
kristw, line 69: We could use `it(...)` function instead of `test(...)` too so it will read more like

```ts
it('should have default x and y accessors', ...)
```

hshoff: sure, i don't have a preference

PR #786, opened 2020-08-20
test(vx-brush,vx-event): add code coverage

hshoff:
#### :weight_lifting: Test

- [brush] add utils test
- [event] add tests

### DIFF
--- a/packages/vx-brush/test/utils.test.ts
+++ b/packages/vx-brush/test/utils.test.ts
@@ -1,0 +1,75 @@
+import { createScale } from '@vx/scale';
+import { getDomainFromExtent, scaleInvert } from '../src/utils';
+
+describe('getDomainFromExtent()', () => {
+  test('it should return { start, end } if scale.invert', () => {
+    const scale = createScale({ domain: [0, 10], range: [2, 4] });
+    const start = 0;
+    const end = 1;
+    const tolerentDelta = 0.5;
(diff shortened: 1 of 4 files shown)


packages/vx-brush/test/utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+describe('scaleInvert()', () => {
+  test('it should return scale.invert(value) if scale.invert', () => {
+    const scale = createScale({ domain: [0, 10], range: [2, 4] });
+    const value = 3;
+    const result = scaleInvert(scale, value);
+    expect(result).toEqual(scale.invert(value));
+  });
+
+  test('it should return the index of domain itmem for scales without invert (like band)', () => {
```
williaster, line 52: noting a typo `itmem` => `item`

packages/vx-event/test/localPoint.test.ts
```diff
@@ -1,7 +1,49 @@
...
+  test('it should return null if called with no arguments', () => {
+    // @ts-ignore
+    expect(localPoint()).toBeNull();
+    // @ts-ignore
+    expect(localPointGeneric(document.createElement('div'))).toBeNull();
+  });
+
+  test('it should handle localPoint(event) and get node from event.target', () => {
+    const e = new MouseEvent('test', {
+      clientX: 0,
```
williaster, line 18: only thought on these is if we should update the values to be non-zero since `0` is the default behavior of `localPoint`

PR #788, opened 2020-08-25
new(vx-brush): update brush position when initialBrushPosition changes

williaster:
#### :rocket: Enhancements

This PR 
- updates `@vx/brush` to update the `Brush`s `state.start/end` when `props.initialBrushPosition` changes. This makes the brush somewhere between a controlled and uncontrolled component. 
- updates the `/brush` demo to show how you could use this to snap the brush position to the nearest month. 
- adds tests

A couple caveats
- this probably shouldn't be used with `onBrushChange` as the # of callbacks is insane. Note in the demo I snap the brush `onBrushEnd` so that the update only happens on mouse up.
- there is an edge case I can't figure out a solution for with this approach: say `initialBrushPosition=[Jan 1, Feb 1]`, then you move the `Brush` slightly to `[Jan 10, Feb 10]`. When you release the brush, `onBrushEnd` is invoked but the snapping logic results in the same `[Jan 1, Feb 1]` position. The `Brush` gets this new `initialBrushPosition`, but because it `===` the previous `initialBrushPosition`, it does **not** update the `Brush` position `state` which stays at `[Jan 10, Feb 10]` 😱 
  - one option is updating `Brush` `state.start/end` whenever `initialBrushPosition` does not match it, but unless the renderer proactively updates `initialBrushPosition` themself, the brush will snap back to `initialBrushPosition` upon `onBrushEnd`

^ would love thoughts on this!

snappy-brush

[Code sandbox link](https://codesandbox.io/s/github/hshoff/vx/tree/chris--controlled-brush/packages/vx-demo/src/sandboxes/vx-brush) to play with it (this will break after the PR is merged)

@kristw @hshoff @emeeks

### DIFF
--- a/packages/vx-brush/src/BaseBrush.tsx
+++ b/packages/vx-brush/src/BaseBrush.tsx
@@ -7,8 +7,15 @@ import BrushHandle from './BrushHandle';
 import BrushCorner from './BrushCorner';
 import BrushSelection from './BrushSelection';
 import { MarginShape, Point, BrushShape, ResizeTriggerAreas, PartialBrushStartEnd } from './types';
+import { shouldUpdateInitialBrushPosition } from './utils';
 
 const BRUSH_OVERLAY_STYLES = { cursor: 'crosshair' };
+const EMPTY_EXTENT = {
+  x0: -1,
+  x1: -1,
+  y0: -1,
+  y1: -1,
+};
 
 type MouseHandlerEvent =
   | React.MouseEvent<SVGRectElement, MouseEvent>
@@ -52,12 +59,7 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     const { initialBrushPosition } = props;
     const extent = initialBrushPosition
       ? this.getExtent(initialBrushPosition.start, initialBrushPosition.end)
-      : {
-          x0: -1,
-          x1: -1,
-          y0: -1,
-          y1: -1,
-        };
+      : EMPTY_EXTENT;
     this.state = {
       start: { x: Math.max(0, extent.x0), y: Math.max(0, extent.y0) },
       end: { x: Math.max(0, extent.x1), y: Math.max(0, extent.y1) },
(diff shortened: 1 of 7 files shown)


williaster:
maybe another alternative would be to introduce a new prop `brushPosition`, which could make `Brush` behave more like a controlled component i.e., when `props.brushPosition` changes from the `Brush`s `state.start/end`, `state` would be updated to the `props` value. This could likely be updated `onChange` and/or `onBrushEnd`.

`initialBrushPosition` would then really only be for the **_initial position_**.

kristw:
I am not sure I fully understand the main problem that this PR is trying to solve. Is it to update brush state when `initialBrushPosition` changes? When would `initialBrushPosition` be dynamically updated? Shouldn't it be set only initially?

> updates @vx/brush to update the Brushs state.start/end when props.initialBrushPosition changes. This makes the brush somewhere between a controlled and uncontrolled component.

williaster:
@kristw sorry, to clarify there are two use cases @emeeks has that you cannot currently do with the brush (these are definitely common so we'd like to support them):

1. Not just set the initial position of the brush but to override the brushed position with a new position sent from some other place (resize the brush region, say, based on a button somewhere else that says "move brush to x")

2. Resize the brush right after brushing. For instance, if you're brushing a histogram and you want the brush to snap to the beginning of the left bar and the end of the right bar.

emeeks:
I agree with @kristw, if we're updating it, it would seem like `initialBrushPosition` becomes a problematic way to express that property. Could we change the name? Or perhaps add a property that is `controlledBrushPosition` or something similar where if updated it overrides the position? I also agree that callback hell is invoked if the brush position changes anywhere but on `end` so perhaps it should be only limited to an update after `end` and not available on the other events.

williaster:
Thanks for the feedback 🙏 I'll try to refactor this so we have a `Brush` (as it stands now / handles its own state) and a `ControlledBrush` (which takes position in as props). It may be tricky because of the way things are implemented now but will give it a shot.

williaster:
going to close this for now as it was more complex than anticipated

Stuhl:
Any way this is going to be implemented? 

It's a very needed feature and without it, it's hard to update the state of the brush.

My suggestion would be to keep it simple and add new props like `brushPositionStart `and `brushPositionEnd` or something similar (with maybe a less daunting name) to not break old code.

PR #789, opened 2020-08-26
new(xychart): add DataProvider, XYChart, LineSeries

williaster:
#### :rocket: Enhancements

This PR adds the initial implementation of the `XYChart` `DataProvider`, and a simple `LineSeries` component to validate that the scale computation works.

**TODO**
- [x] add tests ([PR](https://github.com/hshoff/vx/pull/795) into this branch for ease of review)
- [x] fix types / get CI to pass

@kristw @techniq @hshoff


packages/vx-demo/src/sandboxes/vx-xychart/Example.tsx
```diff
@@ -1,20 +1,32 @@
...
-  const [theme, setTheme] = useState<XYChartTheme>(lightTheme);
+  const [theme, setTheme] = useState<XYChartTheme>(darkTheme);
 
   return (
     <>
-      <ThemeProvider theme={theme}>
-        <XYChart />
-      </ThemeProvider>
+      <DataProvider theme={theme} xScale={xScaleConfig} yScale={yScaleConfig}>
+        <XYChart height={400}>
```
williaster, line 25: note auto width

packages/vx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -0,0 +1,62 @@
...
+import DataContext from '../../context/DataContext';
+
+type LineSeriesProps<Datum> = {
+  dataKey: string;
+  data: Datum[];
+  xAccessor: (d: Datum) => unknown;
+  yAccessor: (d: Datum) => unknown;
+};
+
+function isValidNumber(_: unknown): _ is number {
```
williaster, line 12: will pull this out into `typeguards/`

packages/vx-xychart/src/providers/DataProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const xMin = margin.left;
+    const xMax = width - margin.right;
+    const yMin = margin.top;
+    const yMax = height - margin.bottom;
+
+    const { type: xType, ...restXScaleConfig } = xScaleConfig;
+    const { type: yType, ...restYScaleConfig } = yScaleConfig;
+
+    let xScale = createScale({
+      type: xType,
```
williaster, line 74: this is failing types still, @kristw any thoughts?

image

packages/vx-xychart/src/providers/DataProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+          : (d3Extent(yValues, v => v) as YScaleInput[]),
+    }) as YScale;
+
+    // apply any scale updates from the registy
+    registryEntries.forEach(entry => {
+      if (entry?.xScale) xScale = entry.xScale(xScale);
+      if (entry?.yScale) yScale = entry.yScale(yScale);
+    });
+
+    xScale = updateScale<AxisScaleOutput, XScaleInput>(xScale, restXScaleConfig) as XScale;
```
williaster, line 99: this may be unnecessary, but currently a registry entry may override scale config specified by lib consumer. not sure how else to preserve this cleanly.

packages/vx-xychart/src/context/DataContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+/** Utilities for infering context generics */
+export type InferXScaleConfig<X extends AnyDataContext> = X extends DataContextType<
+  infer T,
+  any,
+  any,
+  any
+>
+  ? T
+  : any;
```
kristw, line 15: can change this `any` to default type for XScaleConfig. May be `AxisScaleConfig`?

packages/vx-xychart/src/providers/DataProvider.tsx
kristw: What if we do another way around?

```ts
const xScale = createScale(xScaleConfig);
type XScale = typeof xScale;
type XScaleInput = ScaleInput<XScale>;
// update domain and range
xScale.domain(...).range(...);
```

packages/vx-xychart/src/classes/DataRegistry.ts
```diff
@@ -0,0 +1,54 @@
...
+import { DataRegistryEntry } from '../types/data';
+
+/** A class for holding data entries */
+export default class DataRegistry<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  YScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  Datum = unknown,
+  DataKey extends string = string
+> {
+  private registry = new Map<
```
kristw, line 12: if key is only `string`, plain map `{ [key: string]: DataRegistryEntry }` is more lightweight than `Map` type.

packages/vx-xychart/src/classes/DataRegistry.ts
```diff
@@ -0,0 +1,54 @@
+import { AxisScaleOutput } from '@vx/axis';
+import { ScaleConfig } from '@vx/scale';
+import { DataRegistryEntry } from '../types/data';
+
+/** A class for holding data entries */
+export default class DataRegistry<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  YScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  Datum = unknown,
+  DataKey extends string = string
```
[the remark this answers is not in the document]
williaster: yeah maybe we can just skip this for now, was trying to allow more specificity like `'sf' | 'ny'` but prob easier to add later on if needed

packages/vx-xychart/src/context/DataContext.tsx
williaster: ohhh I see what you mean now, sorry I didn't get this yesterday 👌

packages/vx-xychart/src/providers/DataProvider.tsx
williaster: oh interesting. I get an error even when passing `xScaleConfig: XScaleConfig extends ScaleConfig<AxisScaleOutput>` directly to `createScale` 🤔 

image

williaster: it looks like it might be inferring that it is the first function overload signature (linear scale) based on `typeof xScale`

image

similar interpretation based on the accessor return value 
image

and re-assigning the `config.domain` later on
image


and

williaster: seems like it struggles to pick the right overload, it complains with `AnyD3Scale` too 🤔 

image

kristw: Hmm, I might have to add another overload at the end to catch all.

kristw: Try adding this to `createScale.ts` (line 110, above the comment `// Actual implementation)

```ts
function createScale<
  Output = DefaultOutput,
  DiscreteInput extends StringLike = StringLike,
  ThresholdInput extends DefaultThresholdInput = DefaultThresholdInput
>(
  config: ScaleConfig<Output, DiscreteInput, ThresholdInput>,
): D3Scale<Output, DiscreteInput, ThresholdInput>;

// Actual implementation
```

williaster: ^I thiiiink I got this to work, but only if that's the _first_ overload, not the last. I'll try to see if the other overloads work as expected still (e.g., linear config => linear scale)

kristw: hmm. usually the broader overload should be the last one. i'll try to see if there are ways to improve this. agree it is not ideak,

williaster: I updated this to be the first overload for now

packages/vx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -0,0 +1,74 @@
...
+  // register data on mount
+  // @TODO(chris) make this easier with HOC
+  useEffect(() => {
+    if (dataRegistry) dataRegistry.registerData({ key: dataKey, data, xAccessor, yAccessor });
+    return () => dataRegistry?.unregisterData(dataKey);
+  }, [dataRegistry, dataKey, data, xAccessor, yAccessor]);
+
+  const getScaledX = useCallback(
+    (d: Datum) => {
+      const x = xScale?.(xAccessor(d));
```
williaster, line 42: @kristw I'm curious if you have any thoughts on a final remaining type issue, which is for the accessors here.

image

I'm not sure if the `InferDataContext` is the issue (i.e., it's not working and is falling back to `AnyDataContext`), or if the way the `xAccessor` return type is defined ([here](https://github.com/hshoff/vx/pull/789/files#diff-74b1e99c28f45e74a09146815a7831e7R31)) is problematic: 

```ts
xAccessor: (d: Datum) => ScaleInput<ScaleConfigToD3Scale<XScaleConfig, AxisScaleOutput>>;
```

kristw: i'm pulling the branch to see locally. types are hard to debug without ide

kristw: Add `, any, any`

```ts
  xAccessor: (
    d: Datum,
  ) => // eslint-disable-next-line @typescript-eslint/no-explicit-any
  ScaleInput<ScaleConfigToD3Scale<XScaleConfig, AxisScaleOutput, any, any>>;
  yAccessor: (
    d: Datum,
  ) => // eslint-disable-next-line @typescript-eslint/no-explicit-any
  ScaleInput<ScaleConfigToD3Scale<YScaleConfig, AxisScaleOutput, any, any>>;
```

kristw: There are some funkiness with the `DiscreteInput` and `ThresholdInput` generics which are the last two parameters. The `StringLike` has some issue matching.

williaster: ooooO yes this fixed it, thanks SO MUCH for the help debugging it! 🙏 ❤️

williaster:
aside from tests (separate PR), this should be in a good place for review now 🎉

packages/vx-xychart/src/hooks/useDataRegistry.ts
```diff
@@ -0,0 +1,13 @@
+import { AxisScaleOutput } from '@vx/axis';
+import { useMemo } from 'react';
+import { ScaleConfig } from '@vx/scale';
+import DataRegistry from '../classes/DataRegistry';
+
+/** Hook that returns a constant instance of a DataRegistry.  */
+export default function useDataRegistry<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  YScaleConfig extends ScaleConfig<AxisScaleOutput>,
+  Datum = unknown
```
kristw, line 10: (optional) Not sure if `Datum` should be `object` instead of `unknown`? But that won't allow primitive types to be `Datum`.

packages/vx-xychart/src/typeguards/isValidNumber.ts
```diff
@@ -0,0 +1,3 @@
+export default function isValidNumber(_: unknown): _ is number {
+  return _ != null && typeof _ === 'number' && !isNaN(_) && isFinite(_);
```
kristw, line 2: Use `Number.isNaN` and `Number.isFinite`

packages/vx-xychart/src/types/data.ts
```diff
@@ -0,0 +1,66 @@
...
+  mouseEvents?: boolean;
+  /** Optionally update the xScale. */
+  xScale?: <XScale extends D3Scale<AxisScaleOutput>>(xScale: XScale) => XScale;
+  /** Optionally update the yScale. */
+  yScale?: <YScale extends D3Scale<AxisScaleOutput>>(yScale: YScale) => YScale;
+  /** Legend shape for the data key. */
+  legendShape?: LegendShape;
+}
+
+export interface DataContext<
```
kristw, line 44: Same here. + Have to modify `DataRegistry` to take scales types instead of scaleconfig types as well.

```ts
export interface DataContext<
  XScale extends AxisScale,
  YScale extends AxisScale,
  Datum
> {
  xScale: XScale;
  yScale: YScale;
  colorScale: PickD3Scale<'ordinal', string, string>;
  width: number;
  height: number;
  innerWidth: number;
  innerHeight: number;
  margin: Margin;
  dataRegistry: DataRegistry<XScale, YScale, Datum>;
  registerData: (
    data:
      | DataRegistryEntry<XScale, YScale, Datum>
      | DataRegistryEntry<XScale, YScale, Datum>[],
  ) => void;
  unregisterData: (keyOrKeys: string | string[]) => void;
  setDimensions: (dims: { width: number; height: number; margin: Margin }) => void;
  theme: XYChartTheme;
}
```

packages/vx-xychart/src/types/data.ts
```diff
@@ -0,0 +1,66 @@
...
+  top: number;
+  right: number;
+  bottom: number;
+  left: number;
+};
+
+export type LegendShape = 'rect' | 'line' | 'dashed-line' | 'circle';
+
+export interface DataRegistryEntry<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
```
kristw, line 22: Since `XScaleConfig` and `YScaleConfig` are not used directly, accepting generics `XScale` and `YScale` instead might be simpler. 

```ts
type AxisScale = D3Scale<AxisScaleOutput>; // not sure if `,any,any` is needed.

export interface DataRegistryEntry<
  XScale extends AxisScale,
  YScale extends AxisScale,
  Datum
> {
  /** unique data key */
  key: string;
  /** array of data for the key. */
  data: Datum[];
  /** function that returns the x value of a datum. */
  xAccessor: (d: Datum) => ScaleInput<XScale>;
  /** function that returns the y value of a datum. */
  yAccessor: (d: Datum) => ScaleInput<YScale>;
  /** whether the entry supports mouse events. */
  mouseEvents?: boolean;
  /** Optionally update the xScale. */
  xScale?: (xScale: XScale) => XScale;
  /** Optionally update the yScale. */
  yScale?: (yScale: YScale) => YScale;
  /** Legend shape for the data key. */
  legendShape?: LegendShape;
}
```

packages/vx-xychart/src/classes/DataRegistry.ts
```diff
@@ -0,0 +1,55 @@
+import { AxisScaleOutput } from '@vx/axis';
+import { ScaleConfig } from '@vx/scale';
+import { DataRegistryEntry } from '../types/data';
+
+/** A class for holding data entries */
+export default class DataRegistry<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
```
kristw, line 7: Can change to

```ts
  XScale extends D3Scale<AxisScaleOutput>,
  YScale extends D3Scale<AxisScaleOutput>,
```

because the config types are not used directly

packages/vx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useContext, useCallback, useEffect } from 'react';
+import LinePath from '@vx/shape/lib/shapes/LinePath';
+import { ScaleConfig, ScaleConfigToD3Scale, ScaleInput } from '@vx/scale';
+import { AxisScaleOutput } from '@vx/axis';
+import DataContext from '../../context/DataContext';
+import isValidNumber from '../../typeguards/isValidNumber';
+
+type LineSeriesProps<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
```
kristw, line 9: This seems like it could take 

```ts
  XScale extends AxisScale,
  YScale extends AxisScale,
```

instead of the configs, similar to the types below

packages/vx-xychart/src/hooks/useScales.ts
```diff
@@ -0,0 +1,81 @@
...
+  dataRegistry: DataRegistry<XScaleConfig, YScaleConfig, Datum>;
+  xRange: [number, number];
+  yRange: [number, number];
+}) {
+  // pull out memoization keys that are less likely to change
+  const registryKeys = dataRegistry.keys();
+  const [xMin, xMax] = xRange;
+  const [yMin, yMax] = yRange;
+
+  return useMemo(() => {
```
kristw, line 30: should this be broken into multiple `useMemo`? Otherwise `x` will also update when only `y` is updated

packages/vx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ) => // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ScaleInput<ScaleConfigToD3Scale<XScaleConfig, AxisScaleOutput, any, any>>;
+  yAccessor: (
+    d: Datum,
+  ) => // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ScaleInput<ScaleConfigToD3Scale<YScaleConfig, AxisScaleOutput, any, any>>;
+};
+
+export default function LineSeries<
+  XScaleConfig extends ScaleConfig<AxisScaleOutput>,
```
kristw, line 26: One thing I am not sure here is how to pass generic from the `XYChart` to the `LineSeries` for stronger type-checking without manually doing

```
<LineSeries<A, B, C> data={} xAccessor={} yAccessor={} />
```

If I understand correctly, if generics are not passed like it is now, TS only tries to infer `XScaleConfig` and `YScaleConfig` back from `xAccessor` and `yAccessor`. Not sure if that backward inference is successful. It might give somewhat broad types.

Not blocking though. Worst-case is type is not 100% strict. Probably ok to revisit later.

packages/vx-xychart/src/hooks/useDataRegistry.ts
williaster: good question, could `extend object = object`. I also thought about primitives, I'm not sure how that could be useful in an xychart context unless you had a constant `x` or `y` (in which case the accessor could return that constant) ... which might be valid. maybe should go for `object` for now and can loosen the restriction in the future.

packages/vx-xychart/src/components/series/LineSeries.tsx
williaster: good question, I'm not sure of this either. I may play around with what the types end up being locally but can revisit when building more robust Series components (this was MVP)

williaster:
@kristw thanks for the great comments + suggestions. addressed them all / CI passed 🎉

packages/vx-xychart/src/hooks/useScales.ts
```diff
@@ -0,0 +1,90 @@
...
+}) {
+  // pull out memoization keys that are less likely to change
+  const registryKeys = dataRegistry.keys();
+  const [xMin, xMax] = xRange;
+  const [yMin, yMax] = yRange;
+
+  const memoizedXScale = useMemo(() => {
+    const registryEntries = registryKeys.map(key => dataRegistry.get(key));
+
+    let xScale = createScale(xScaleConfig) as XScale;
```
williaster, line 33: note: when passing `Scale` in as the generic, the only way I could this right was to coerce here. while we can infer `Scale` from `ScaleConfig`, we can't currently do the opposite.

kristw (approved):

PR #790, opened 2020-08-26
minor fix to <Text> and re-enable unit tests.

kristw:
#### :bug: Bug Fix

- Fix complains when `NaN` or invalid values are passed as `x` or `y`

```
Warning: Received NaN for the `%s` attribute. If this is expected, cast the value to a string.%s
```

#### :house: Internal

- Remove `prop-types` from dependency. (no longer used)
- List `lodash.memoize` in dependency of `vx/text` which is smaller than entire `lodash`. 
- Fix issue with jsdom and re-enable unit tests.

### DIFF
--- a/packages/vx-text/package.json
+++ b/packages/vx-text/package.json
@@ -29,11 +29,10 @@
   "homepage": "https://github.com/hshoff/vx#readme",
   "dependencies": {
     "@types/classnames": "^2.2.9",
-    "@types/lodash": "^4.14.146",
+    "@types/lodash.memoize": "^4.1.6",
     "@types/react": "*",
     "classnames": "^2.2.5",
-    "lodash": "^4.17.15",
-    "prop-types": "^15.7.2",
+    "lodash.memoize": "^4.1.2",
     "reduce-css-calc": "^1.3.0"
   },
   "peerDependencies": {
--- a/packages/vx-text/package.json
+++ b/packages/vx-text/package.json
@@ -29,10 +29,11 @@
   "homepage": "https://github.com/hshoff/vx#readme",
   "dependencies": {
     "@types/classnames": "^2.2.9",
-    "@types/lodash.memoize": "^4.1.6",
+    "@types/lodash": "^4.14.160",
     "@types/react": "*",
     "classnames": "^2.2.5",
-    "lodash.memoize": "^4.1.2",
+    "lodash": "^4.17.20",
+    "prop-types": "^15.7.2",
     "reduce-css-calc": "^1.3.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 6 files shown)


packages/vx-text/package.json
```diff
@@ -29,11 +29,10 @@
   "homepage": "https://github.com/hshoff/vx#readme",
   "dependencies": {
     "@types/classnames": "^2.2.9",
-    "@types/lodash": "^4.14.146",
+    "@types/lodash.memoize": "^4.1.6",
     "@types/react": "*",
     "classnames": "^2.2.5",
-    "lodash": "^4.17.15",
-    "prop-types": "^15.7.2",
+    "lodash.memoize": "^4.1.2",
```
williaster, line 35: I've gone through and done this before but had forgot this was intentional, see #66

packages/vx-text/package.json
```diff
@@ -29,11 +29,10 @@
   "homepage": "https://github.com/hshoff/vx#readme",
   "dependencies": {
     "@types/classnames": "^2.2.9",
-    "@types/lodash": "^4.14.146",
+    "@types/lodash.memoize": "^4.1.6",
     "@types/react": "*",
     "classnames": "^2.2.5",
-    "lodash": "^4.17.15",
-    "prop-types": "^15.7.2",
```
williaster, line 36: we actually need this because we have a babel plugin that adds proptypes based on the types defined for a component.

this is useful for consumers I think (proptypes complain in dev / in the console), but @hshoff also was surprised by this recently so maybe we could consider removing it througout.

packages/vx-text/src/Text.tsx
```diff
@@ -176,6 +185,11 @@ class Text extends React.Component<TextProps, TextState> {
     const { wordsByLines } = this.state;
     const { x, y } = textProps;
 
+    // Cannot render <text> if x or y is invalid
+    if (!isValidXOrY(x) || !isValidXOrY(y)) {
+      return <svg ref={innerRef} x={dx} y={dy} fontSize={textProps.fontSize} style={SVG_STYLE} />;
```
williaster, line 190: do you think we should return `null` or an `svg`?

packages/vx-text/test/svgMock.ts
```diff
@@ -0,0 +1,25 @@
+// @ts-ignore
+let originalFn: typeof SVGElement.prototype.getComputedTextLength;
+
+/**
+ * JSDom does not implement getComputedTextLength()
+ * so this function add mock implementation for testing.
+ */
+export function addMock() {
```
williaster, line 8: nice!

packages/vx-text/package.json
kristw: oh ok sure. will revert.

packages/vx-text/src/Text.tsx
kristw: I was thinking about return null but think if at least having the svg may keep the layout from changing if consumer depends on having some element there.

packages/vx-text/package.json
kristw: ok!

PR #791, opened 2020-08-26
feat: scale overload

kristw:
#### :rocket: Enhancements

- Add fallback overload for `createScale`


williaster:
I think this may not work as the last overload, I only got it to work as
the first overload 🤔

https://github.com/hshoff/vx/pull/789#discussion_r477827250

PR #808, opened 2020-09-23
new(xychart): add BarSeries

williaster:
This PR adds `BarSeries` to the `xychart` package and refactors / fixes other issues in the process

#### :rocket: Enhancements

- adds `BarSeries` component
- adds `withRegisteredData` HOC which swaps `data`/`xAccessor`/`yAccessor` props for those in the `DataRegistry` in context. This makes `*Series` implementation more straight forward because they should always use values from context so they are in sync with the context scales.
- re-writes `useDataRegistry`
  - previously this hook just returned a constant instance of `DataRegistry`, which did not result in the necessary `hook` => `scales` => `context` => `component` updates when the registry was updated.
  - the new implementation mirrors the `DataRegistry` API and uses it for its implementation

- improves the light/dark theme colors:

#### :memo: Documentation

for the `/xychart` demo:
- adds checkbox controls to toggle `line` or `bar` series
- adds `render horizontally` toggle

Adds tests for all new components.

@kristw @techniq @hshoff

### DIFF
--- a/packages/visx-xychart/src/components/series/BarSeries.tsx
+++ b/packages/visx-xychart/src/components/series/BarSeries.tsx
@@ -1,0 +1,103 @@
+import React, { useContext, useCallback, useMemo } from 'react';
+import { AxisScale } from '@visx/axis';
+import DataContext from '../../context/DataContext';
+import { SeriesProps } from '../../types';
+import withRegisteredData, { WithRegisteredDataProps } from '../../enhancers/withRegisteredData';
+import getScaledValueFactory from '../../utils/getScaledValueFactory';
+import isValidNumber from '../../typeguards/isValidNumber';
+import getScaleBandwidth from '../../utils/getScaleBandwidth';
+
+type BarSeriesProps<
+  XScale extends AxisScale,
+  YScale extends AxisScale,
+  Datum extends object
+> = SeriesProps<XScale, YScale, Datum> & {
+  /** Whether bars should be rendered horizontally instead of vertically. */
+  horizontal?: boolean;
+  /**
+   * Specify bar padding when bar thickness does not come from a `band` scale.
+   * Accepted values are [0, 1], 0 = no padding, 1 = no bar, defaults to 0.1.
+   */
+  barPadding?: number;
+};
+
+function BarSeries<XScale extends AxisScale, YScale extends AxisScale, Datum extends object>({
+  barPadding = 0.9,
+  data,
+  dataKey,
+  horizontal,
+  xAccessor,
+  xScale,
+  yAccessor,
+  yScale,
+}: BarSeriesProps<XScale, YScale, Datum> & WithRegisteredDataProps<XScale, YScale, Datum>) {
+  const { colorScale, theme, innerWidth = 0, innerHeight = 0 } = useContext(DataContext);
+  const getScaledX = useCallback(getScaledValueFactory(xScale, xAccessor), [xScale, xAccessor]);
+  const getScaledY = useCallback(getScaledValueFactory(yScale, yAccessor), [yScale, yAccessor]);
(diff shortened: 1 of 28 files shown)


packages/visx-xychart/src/components/series/BarSeries.tsx
```diff
@@ -0,0 +1,103 @@
...
+type BarSeriesProps<
+  XScale extends AxisScale,
+  YScale extends AxisScale,
+  Datum extends object
+> = SeriesProps<XScale, YScale, Datum> & {
+  /** Whether bars should be rendered horizontally instead of vertically. */
+  horizontal?: boolean;
+  /**
+   * Specify bar padding when bar thickness does not come from a `band` scale.
+   * Accepted values are [0, 1], 0 = no padding, 1 = no bar, defaults to 0.1.
```
williaster, line 19: this is backwards and was updated to match d3's band scale logic: `1=no bar, 0=no padding`. will fix.

packages/visx-xychart/src/components/series/BarSeries.tsx
```diff
@@ -0,0 +1,103 @@
...
+   */
+  barPadding?: number;
+};
+
+function BarSeries<XScale extends AxisScale, YScale extends AxisScale, Datum extends object>({
```
williaster, line 24: most of this logic can become `BaseBarSeries` in the future to power animated and non-animated bars.

packages/visx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -1,55 +1,29 @@
-import React, { useContext, useCallback, useEffect } from 'react';
+import React, { useContext, useCallback } from 'react';
```
kristw, line 1: nice. The code in `LineSeries` looks much cleaner with the HOC.

packages/visx-xychart/src/components/series/BarSeries.tsx
```diff
@@ -0,0 +1,103 @@
...
+    : Math.min(xMin, xMax);
+  const yZeroPosition = isValidNumber(maybeYZero)
+    ? Math.min(maybeYZero, Math.max(yMin, yMax))
+    : Math.max(yMin, yMax);
+
+  const color = colorScale?.(dataKey) ?? theme?.colors?.[0] ?? '#222';
+
+  const bars = useMemo(
+    () =>
+      data.map(datum => {
```
kristw, line 68: This can be computed outside of the loop.
```ts
(scaleBandwidth || !horizontal ? 0 : barThickness / 2)
```

kristw (commented):
overall lgtm. have small comments and wait for green ci to approve

packages/visx-xychart/src/hooks/useDataRegistry.ts
```diff
@@ -1,12 +1,42 @@
...
   XScale extends AxisScale,
   YScale extends AxisScale,
-  Datum = unknown
+  Datum extends object
 >() {
-  return useMemo(() => new DataRegistry<XScale, YScale, Datum>(), []);
+  const [, forceUpdate] = useState(Math.random());
+  const privateRegistry = useMemo(() => new DataRegistry<XScale, YScale, Datum>(), []);
+
+  const registerData = useCallback(
```
kristw, line 14: as all of these 5 functions depends only on `privateRegistry`, perhaps a single `useMemo` block is enough?
or should the first two also depend on the value of `forceUpdate`?

kristw (approved):

packages/visx-xychart/src/hooks/useDataRegistry.ts
williaster: good call! initially I didn't have the force update, but you're right that now a single memo works 👌 had to wrap some of the functions for the appropriate `this` context to work.

Issue #813, opened 2020-09-26
@vx/text doesn't work with 0 as a number


ilariaventurini:
Example:

```js
<svg x={0} y={0} width={200} height={200}>
  {/* doesn't work */}
  <Text x={0} y={10} fill="black">
    {0}
  </Text>

  {/* works */}
  <Text x={0} y={10} fill="black">
    0
  </Text>
</svg>
```

The result is:

```js
<svg x="0" y="0" width="200" height="200">
  <svg x="0" y="0" style="overflow: visible;">
    <text x="0" y="10" fill="black" text-anchor="start">
      <tspan x="0" dy="0em"></tspan> // <-- no child element
    </text>
  </svg>
  <svg x="0" y="0" style="overflow: visible;">
    <text x="0" y="10" fill="black" text-anchor="start">
      <tspan x="0" dy="0em">0</tspan>
    </text>
  </svg>
</svg>
```

PR #814 should fix this problem.

PR #814, opened 2020-09-26
fix(demo/text): render 0 as number (#813)

ilariaventurini:
#### :bug: Bug Fix

- closes: #813


hshoff (approved):
Thanks for the contribution! Looks good to me.

packages/visx-text/src/Text.tsx
```diff
@@ -115,7 +115,7 @@ class Text extends React.Component<TextProps, TextState> {
     // Only perform calculations if using features that require them (multiline, scaleToFit)
     if (props.width || props.scaleToFit) {
       if (needCalculate) {
-        const words: string[] = props.children
+        const words: string[] = (props.children !== null && props.children !== undefined)
```
williaster, line 118: this could be simplified to `props.children != null` which captures both conditions

packages/visx-text/src/Text.tsx
```diff
@@ -138,7 +138,7 @@ class Text extends React.Component<TextProps, TextState> {
   }
 
   updateWordsWithoutCalculate(props: TextProps) {
-    const words = props.children ? props.children.toString().split(/(?:(?!\u00A0+)\s+)/) : [];
+    const words = (props.children !== null && props.children !== undefined) ? props.children.toString().split(/(?:(?!\u00A0+)\s+)/) : [];
```
williaster, line 141: ditto `!=`

williaster:
@ilariaventurini thanks for the fix! 

looks like the build is failing due to `eslint` errors. The seem to all be auto-fixable so if you run `yarn lint --fix` from the monorepo root it should fix everything ✨

ilariaventurini:
I edited the code using `props.children !== null` (as @williaster suggested me) instead of `props.children !== null && props.children !== undefined`.
Tests all pass. 
Then I tried running `yarn fix:lint` but I get this error:

```bash
119:11  error  Unexpected negated condition  no-negated-condition
141:7   error  Unexpected negated condition  no-negated-condition
```

So the build is failing.

packages/visx-text/src/Text.tsx
williaster: `!=` I think is still ideal, but there is another lint rule which favors **not** having negated conditions (because they are harder to understand logically). So I think this should fix things:

```suggestion
    const words = props.children == null ? [] : props.children.toString().split(/(?:(?!\u00A0+)\s+)/);
```

williaster (commented):
Hey @ilariaventurini , sorry for the misdirection in suggestiong `eslint --fix`, it looks like that converted a lot of TypeScript `any`s to `unknown` which aren't equivalent, and now failed the build.

I think it's probably easiest to revert back to your first commit [9f98551](https://github.com/airbnb/visx/pull/814/commits/9f9855124f917382a3b93cb4c3a3f9d251a081ee), and then apply my suggested fix to reverse the negated condition. Then this should be good to go 🙏

williaster:
@ilariaventurini sorry looks like you still have all the auto lint fix stuff in here. when I said revert back to your first commit I meant run `git reset --hard 9f98551`. 

That commit had lint issues, so from there you should simply be able to make the change to `const words = props.children == null ? [] : props.children.toString().split(/(?:(?!\u00A0+)\s+)/)` then force push your branch.

williaster (approved):
Thanks @ilariaventurini !

PR #828, opened 2020-10-01
fix TooltipWithBounds positioning

### LINKED ISSUE #767: TooltipWithBounds bug
image

singhanurag05:
#### :bug: Bug Fix

- This PR should close #767 
Please correct me if I did something wrong as I am new to Open Source.

### DIFF
--- a/packages/visx-tooltip/src/tooltips/Tooltip.tsx
+++ b/packages/visx-tooltip/src/tooltips/Tooltip.tsx
@@ -41,9 +41,9 @@ export default function Tooltip({
       style={{
         left: 0,
         top: 0,
-        transform: `translate(${left == null || offsetLeft == null ? left : left + offsetLeft}px, ${
-          top == null || offsetTop == null ? top : top + offsetTop
-        }px)`, 
+        transform: `translate(${
+          left == null || offsetLeft == null ? left ?? 0 : left + offsetLeft
+        }px, ${top == null || offsetTop == null ? top ?? 0 : top + offsetTop})`,
         ...(!unstyled && style),
       }}
       {...restProps}
--- a/packages/visx-tooltip/src/tooltips/Tooltip.tsx
+++ b/packages/visx-tooltip/src/tooltips/Tooltip.tsx
@@ -43,7 +43,7 @@ export default function Tooltip({
         top: 0,
         transform: `translate(${
           left == null || offsetLeft == null ? left ?? 0 : left + offsetLeft
-        }px, ${top == null || offsetTop == null ? top ?? 0 : top + offsetTop})`,
+        }px, ${top == null || offsetTop == null ? top ?? 0 : top + offsetTop}px)`,
         ...(!unstyled && style),
       }}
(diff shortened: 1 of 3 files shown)


singhanurag05:
@williaster Was this correct or not? Did I do something wrong? Please reply.

williaster:
hey @singhanurag05 thanks for the fix! just to check I pulled this branch and played with it locally, then realized that the ideal solution is a bit more complicated. 

I opened #831 with what I think is the needed fix. If you'd like to implement this yourself for more coding experience I'm happy to close it if you make those updates! Otherwise we can close this PR and use that one.

singhanurag05:
I had thought of the exact same thing! please let me do this @williaster

williaster:
@singhanurag05 great! One comment on my changes: let's **not** [set `position: 'absolute' by default](https://github.com/airbnb/visx/pull/831/files#diff-cac01199643221c3bdea60ceeb5e06d2R42) as this would be a breaking change. This means that when you set `unstyled`, you must set `.visx-tooltip { position: absolute; }` for correct positioning but we want to improve this in another PR in a non-breaking way.

singhanurag05:
@williaster Sorry I didn't get you. Can you please explain?

williaster:
In my PR I made the change to remove `position: absolute` from `defaultStyles`, and **always** set it in `Tooltip`. This is technically needed for the tooltip positioning to work properly (without it, consumers who set `unstyled` **must** set their own css style rule `.visx-tooltip { position: absolute; }`), but it's technically a breaking change (which would require a `2.0` release). 

so if you mirror the changes in my PR, I just meant to **not** make that change (i.e., leave `position: absolute` in `defaultStyles`, and don't apply it in `Tooltip`). make sense?

singhanurag05:
Absolutely yes! i'll make the changes in sometime. Thanks for the help!

williaster (commented):
Looks good to me! had one last suggestion but then I think this is good to go.

singhanurag05:
Thank you! Hope to contribute further.

packages/visx-tooltip/src/tooltips/Tooltip.tsx
```diff
@@ -39,8 +39,11 @@ export default function Tooltip({
     <div
       className={cx('visx-tooltip', className)}
       style={{
-        top: top == null || offsetTop == null ? top : top + offsetTop,
-        left: left == null || offsetLeft == null ? left : left + offsetLeft,
+        left: 0,
+        top: 0,
+        transform: `translate(${
+          left == null || offsetLeft == null ? left ?? 0 : left + offsetLeft
+        }px, ${top == null || offsetTop == null ? top ?? 0 : top + offsetTop})`,
```
williaster, line 46: oops forgot a unit here, I may just commit this to get it merged

```suggestion
        }px, ${top == null || offsetTop == null ? top ?? 0 : top + offsetTop}px)`,
```

williaster (approved):

PR #831, opened 2020-10-01
fix(tooltip): correct positioning when unstyled is applied

### LINKED ISSUE #767: TooltipWithBounds bug

williaster:
#### :bug: Bug Fix

This PR
- fixes #767 where setting `unstyled` on `TooltipWithBounds` breaks positioning because it is passed to `Tooltip`, which then doesn't respect the `style` prop which includes the needed offset
- fixes an issue where if `unstyled` is set on _either_ `Tooltip` or `TooltipWithBounds`, lib consumers must also minimally set `.visx-tooltip { position: absolute; }` themselves in order for the tooltips to be correctly positioned since they rely on `absolute` positioning.
  - this was fixed by removing `position: absolute` from `defaultStyles`, and always setting it (along with the positional offset)

**Testing**
Verified in `/tooltip` demo that tooltip is correctly positioned when `unstyled` is set for all combinations 
- boundary detection + ~portal~
- ~boundary detection~ + ~portal~
- boundary detection + portal
- ~boundary detection~ + portal

@hshoff @kristw @singhanurag05

### DIFF
--- a/packages/visx-tooltip/src/tooltips/Tooltip.tsx
+++ b/packages/visx-tooltip/src/tooltips/Tooltip.tsx
@@ -13,7 +13,6 @@ export type TooltipProps = {
 };
 
 export const defaultStyles: React.CSSProperties = {
-  position: 'absolute',
   backgroundColor: 'white',
   color: '#666666',
   padding: '.3rem .5rem',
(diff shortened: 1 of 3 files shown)


singhanurag05:
@williaster Sorry I didn't get you. Can you please explain?

torkelo:
Nice! just ran into this today!

williaster:
Closing in favor of #828

PR #834, opened 2020-10-02
new(responsive): add `ignoreDimensions` prop to handle rerenders

### LINKED ISSUE #247: ParentSize disableHieght disableWidth
Often children interested only in one of dimensions or positions, so its handy to limit redraw events with dome flags so redraw not fired if `disableHieght = true` and parent changed height only

wyze:
#### :rocket: Enhancements

This PR

- add a new prop to ParentSize (`disableFor`) to skip when a component doesn't want to be notified by a size change of a certain prop
- also converted the class component to a functional component (hopefully was ok). I did that in a separate commit so you could see what was needed for the feature

Note: I did attempt a test and was unsuccessful, so more than happy to add one provided some direction. I use react-testing-library, so unsure how to simulate resize of an element with enzyme.

~Also, `disableFor` might not be the best name for the prop, so I am open to changing it.~

Close #247

### DIFF
--- a/packages/visx-responsive/src/components/ParentSize.tsx
+++ b/packages/visx-responsive/src/components/ParentSize.tsx
@@ -7,10 +7,10 @@ export type ParentSizeProps = {
   className?: string;
   /** Child render updates upon resize are delayed until `debounceTime` milliseconds _after_ the last resize event is observed. */
   debounceTime?: number;
-  /** Optional dimensions provided won't trigger a state change when changed. */
-  disableFor?: keyof ParentSizeState | (keyof ParentSizeState)[];
   /** Optional flag to toggle leading debounce calls. When set to true this will ensure that the component always renders immediately. (defaults to true) */
   enableDebounceLeadingCall?: boolean;
+  /** Optional dimensions provided won't trigger a state change when changed. */
+  ignoreDimensions?: keyof ParentSizeState | (keyof ParentSizeState)[];
   /** Optional `style` object to apply to the parent `div` wrapper used for size measurement. */
   parentSizeStyles?: React.CSSProperties;
   /** Child render function `({ width, height, top, left, ref, resize }) => ReactNode`. */
@@ -35,7 +35,7 @@ export default function ParentSize({
   className,
   children,
   debounceTime = 300,
-  disableFor,
+  ignoreDimensions = [],
   parentSizeStyles = { width: '100%', height: '100%' },
   enableDebounceLeadingCall = true,
   ...restProps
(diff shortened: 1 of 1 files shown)


packages/visx-responsive/src/components/ParentSize.tsx
```diff
@@ -29,74 +31,65 @@ type ParentSizeState = {
...
+  const resize = useMemo(() => {
+    const normalized = disableFor ? (Array.isArray(disableFor) ? disableFor : [disableFor]) : [];
+
+    return debounce(
+      (incoming: ParentSizeState) => {
+        setState(existing => {
+          const stateKeys = Object.keys(existing) as (keyof ParentSizeState)[];
+          const keysWithChanges = stateKeys.reduce<(keyof ParentSizeState)[]>((diff, key) => {
+            return existing[key] === incoming[key] ? diff : [...diff, key];
+          }, []);
```
williaster, line 57: Could we use `filter` here instead? I think it's a little more readable

```suggestion
          const keysWithChanges = stateKeys.filter(key => existing[key] !== incoming[key])
```

packages/visx-responsive/src/components/ParentSize.tsx
```diff
@@ -1,12 +1,14 @@
...
-import React from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import ResizeObserver from 'resize-observer-polyfill';
 
 export type ParentSizeProps = {
   /** Optional `className` to add to the parent `div` wrapper used for size measurement. */
   className?: string;
   /** Child render updates upon resize are delayed until `debounceTime` milliseconds _after_ the last resize event is observed. */
   debounceTime?: number;
+  /** Optional keys provided won't trigger a state change when changed */
```
williaster, line 10:
```suggestion
  /** Optional dimensions provided won't trigger a state change when changed. */
```

packages/visx-responsive/src/components/ParentSize.tsx
```diff
@@ -1,12 +1,14 @@
...
+  disableFor?: keyof ParentSizeState | (keyof ParentSizeState)[];
```
williaster, line 11: Agree the name may not be ideal here, what about `ignoreDimensions`? `disable` doesn't quite seem right since we are still passing those dimensions, just ignoring their changes.

williaster (commented):
Thanks for the addition @wyze, and the `FunctionComponent` refactor! 💯 

I had a couple of small suggestions but otherwise lgtm. I think for now don't worry about new tests, mocking these things in enzyme isn't super easy / I can add the boilerplate later.

packages/visx-responsive/src/components/ParentSize.tsx
wyze: Whoops, I agree. I think I had a started with some different data structure and finally settled on this where my reduce end up being an over-engineered filter.

packages/visx-responsive/src/components/ParentSize.tsx
wyze: Sure, ignoreDimensions works! Sounds better to me.

williaster (approved):
lgtm, thanks for the addition @wyze!

PR #836, opened 2020-10-02
Add `initialHeight` and `initialWidth` to `withParentSize`

### LINKED ISSUE #554: Feature: specify default width/height for `withParentSize` from @visx/responsive
Hello, I would like to suggest to add a way to specify default width/height for `withParentSize` enahncer from `@fx/repsponsive`.

e.g.:

```js
export default withParentSize(Component, {
  initialHeight: 400,
  initialWidth: 400,
});
```

This will help to avoid layout shifting during component mounting.

As alternative we can add another property e.g. `parentSizeMeasured` (or a way to render placeholder in place of original `Component`) to allow Component to provide stub view, till we get actual size.

If this feature will be accepted, I can prepare PR

shannonrothe:
#### :rocket: Enhancements

Adds the ability to set `initialWidth/initialHeight` to `@visx/responsive`s `withParentSize`. Closes #554.

### DIFF
--- a/packages/visx-responsive/src/enhancers/withParentSize.tsx
+++ b/packages/visx-responsive/src/enhancers/withParentSize.tsx
@@ -24,14 +24,14 @@ export default function withParentSize<BaseComponentProps extends WithParentSize
   return class WrappedComponent extends React.Component<
     BaseComponentProps & WithParentSizeProvidedProps,
     WithParentSizeState
-  > {
+    > {
     static defaultProps = {
       debounceTime: 300,
       enableDebounceLeadingCall: true,
     };
     state = {
-      parentWidth: this.props.initialWidth,
-      parentHeight: this.props.initialHeight,
+      parentWidth: undefined,
+      parentHeight: undefined,
     };
     animationFrameID: number = 0;
     resizeObserver: ResizeObserver | undefined;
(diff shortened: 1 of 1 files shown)


wyze (changes requested):
In my opinion, this change should be reverted. It is an anti-pattern as you can see here https://reactjs.org/docs/react-component.html#constructor if you scroll down a little bit to the `Note` section.

A better solution would be to modify the `render` function like so:

```diff
  render() {
+   const { initialWidth, initialHeight } = this.props;
+   const { parentWidth = initialWidth, parentHeight = initialHeight } = this.state;
-   const { parentWidth, parentHeight } = this.state;
    return (
      <div style={CONTAINER_STYLES} ref={this.setRef}>
```

williaster:
thanks for the PR @shannonrothe! 

I'm okay with this implementation or @wyze 's suggestion. The react docs highlight that it's an anti-pattern _if you are expecting `state` to update when `props` change_ and you get into a mess with `componenDidUpdate`, but here the naming of the `props` with `initialXXX` makes is clear that's not the expectation.

Let me know which you prefer and we can merge this puppy 🚀 .

shannonrothe:
> thanks for the PR @shannonrothe!
> 
> I'm okay with this implementation or @wyze 's suggestion. The react docs highlight that it's an anti-pattern _if you are expecting `state` to update when `props` change_ and you get into a mess with `componenDidUpdate`, but here the naming of the `props` with `initialXXX` makes is clear that's not the expectation.
> 
> Let me know which you prefer and we can merge this puppy 🚀 .

Yeah I thought it would be implied that those values wouldn't update but I think @wyze's suggestion is a little nicer. I've fixed up so should be ready to merge :+1:

Issue #466, opened 2019-06-07
TooltipWithBounds overflowing its parent on small screens


sun-tea:
I use the [TooltipWithBounds](https://github.com/hshoff/vx/blob/master/packages/vx-tooltip/src/tooltips/TooltipWithBounds.js) to display tooltips on vx charts. I noticed that on small screens, the tooltip may overflow if neither left or right of the touch point (I'm on mobile) has enough place to fit the tooltip.
In my case, the left prop thus becomes negative. Maybe we could make the tooltip go above or below the touch point/cursor in these cases ?

<img width="404" alt="Screenshot 2019-06-07 at 16 05 32" src="https://user-images.githubusercontent.com/6019103/59110132-b3db9280-893e-11e9-9550-52251fc2cb1c.png">

hshoff:
Hi @sun-tea, first off, that's a very nice look chart. Great work!

Thanks for reporting this, we can get this fixed in `<TooltipWithBounds />`. Until the fix lands, the best work around would be to make your own `<CustomTooltipWithBounds />` -- something like:

```jsx
// CustomTooltipWithBounds.js

import React from 'react';
import { Tooltip } from '@vx/tooltip';
import { withBoundingRects } from '@vx/bounds';

function CustomTooltipWithBounds({
  left: initialLeft,
  top: initialTop,
  offsetLeft = 10,
  offsetTop = 10,
  rect,
  parentRect,
  getRects,
  children,
  style,
  ...otherProps
}) {

  let left = initialLeft;
  let top = initialTop;

  // custom implementation

  return (
    <Tooltip
      style={{ top: 0, transform: `translate(${left}px, ${top}px)`, ...style }}
      {...otherProps}
    >
      {children}
    </Tooltip>
  );
}
export default withBoundingRects(CustomTooltipWithBounds);
```

sun-tea:
Yep, I was aiming to a solution like this. Thank you for the quick response Harrison, vx is an awesome lib 🙂

sun-tea:
I'm sorry for the question which is not entirely relevant to this issue, but I was wondering if there was any other workaround other than putting a `key={Math.random()}` to force the tooltip to have its bounds up-to-date? If I remove that it indeed does not work properly but perf-wise I'm afraid it is not the best.

th0th:
I came up with something like this for preventing horizontal overflow, I am sharing in case it works for somebody else, too:

```
if (left > parentRect.width / 2) {
[...]

PR #837, opened 2020-10-02
[TooltipWithBounds] fix : overflowing its parent on small screens

LethalPants:
#### :bug: Bug Fix

-  #466 : change logic to calculate position

image

image

### DIFF
--- a/packages/visx-tooltip/src/tooltips/TooltipWithBounds.tsx
+++ b/packages/visx-tooltip/src/tooltips/TooltipWithBounds.tsx
@@ -4,58 +4,68 @@ import { withBoundingRects, WithBoundingRectsProps } from '@visx/bounds';
 import Tooltip, { TooltipProps, defaultStyles } from './Tooltip';
 
 export type TooltipWithBoundsProps = {
-  offsetLeft?: number;
-  offsetTop?: number;
+	offsetLeft?: number;
+	offsetTop?: number;
 } & TooltipProps &
-  React.HTMLProps<HTMLDivElement> &
-  WithBoundingRectsProps;
+	React.HTMLProps<HTMLDivElement> &
+	WithBoundingRectsProps;
 
 function TooltipWithBounds({
-  left: initialLeft = 0,
-  top: initialTop = 0,
-  offsetLeft = 10,
-  offsetTop = 10,
-  children,
-  rect: ownBounds,
-  parentRect: parentBounds,
-  getRects,
-  style = defaultStyles,
-  unstyled = false,
-  ...otherProps
+	left: initialLeft = 0,
+	top: initialTop = 0,
+	offsetLeft = 10,
+	offsetTop = 10,
+	children,
+	rect: ownBounds,
+	parentRect: parentBounds,
+	getRects,
+	style = defaultStyles,
+	unstyled = false,
+	...otherProps
 }: TooltipWithBoundsProps) {
-  let left = initialLeft;
-  let top = initialTop;
+	let left = initialLeft;
+	let top = initialTop;
 
-  if (ownBounds && parentBounds) {
-    const placeTooltipLeft =
-      offsetLeft + ownBounds.right > parentBounds.right ||
-      offsetLeft + ownBounds.right > window.innerWidth;
+	if (ownBounds && parentBounds) {
+		if (left > parentBounds.width / 2) {
+			left = left - offsetLeft - ownBounds.width;
 
-    const placeTooltipUp =
-      offsetTop + ownBounds.bottom > parentBounds.bottom ||
-      offsetTop + ownBounds.bottom > window.innerHeight;
+			if (left < 0) {
+				left = 0;
+			}
+		} else {
+			if (left + offsetLeft + ownBounds.width > parentBounds.width) {
+				left = parentBounds.width - ownBounds.width;
+			}
+		}
+		if (top > parentBounds.height / 2) {
+			top = top - offsetTop - ownBounds.height;
+			if (top < 0) {
+				top = 0;
+			}
+		} else {
+			if (top + offsetTop + ownBounds.height > parentBounds.height) {
+				top = parentBounds.height - ownBounds.height;
+			}
+		}
+	}
 
-    left = placeTooltipLeft ? left - ownBounds.width - offsetLeft : left + offsetLeft;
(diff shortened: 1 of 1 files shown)


williaster:
Hey @LethalPants thanks for the PR! In playing with this locally there are a couple of issues:

- `leftOffset/topOffset` are not applied correctly when the tooltip is positioned to the bottom or the right (they are flush with the lines but there should be 10px of offset padding)



- to `tooltipInPortal` is always positioned to the bottom left regardless of its position

LethalPants:
Okay I'll take a look at it and make the changes. Thanks!

joshxyzhimself:
Hi @LethalPants this looks gangsta, just curious do you think this is also related to https://github.com/airbnb/visx/issues/839?

LethalPants:
> to `tooltipInPortal` is always positioned to the bottom left regardless of its position

I don't think this and the overflow issue my commit fixes are related as it is persistent in the current version too, just that the tooltip is in the bottom right not sure if that was planned. 

Unrelated but the tooltip is under the plot 👀 

image

LethalPants:
I've changed the logic so that the position of the tooltip is change when the gap is less than 1% of the width of `parentBounds`
Which eliminated the offset issue mentioned by @williaster.

williaster (commented):
@LethalPants sorry for the delay, thanks for iterating on this! After playing with this version, I think something like the following might be the ideal fix:

```tsx
const placeTooltipLeft = parentBounds.width
  ? left + offsetLeft > parentBounds.width / 2
  : left + offsetLeft > window.innerWidth / 2;

const placeTooltipUp = parentBounds.height
  ? top + offsetTop > parentBounds.height / 2
  : top + offsetTop > window.innerHeight / 2;
```

I think this has two benefits versus the 1% implementation:
1) This always puts the tooltip in the quadrant with the most space (even if it's clipped; to avoid this I agree the Tooltip needs to have a dynamic width). With the 1% logic, if the tooltip will overflow on either side, it will still choose the side with less space (which is the bug report)

**1% logic (current)**


**largest quadrant logic proposed)**


2) This logic doesn't depend on `ownBounds.left/right`, which actually is a great performance win. If you check the example code, we often use `key={Math.random()}` for the tooltip to force update `ownBounds`, but that re-mounts the component on every mouse move. If we don't need `ownBounds`, we can remove that and improve perf pretty significantly.

williaster:
Thanks for the 🐛 report for the z-index issue in the airbnb.io/visx site, I noticed this as well and created an issue #850

williaster:
Sorry, playing with my suggested code more also isn't 100% perfect for the `Portal` case because `parentBounds` is basically the entire page. So if you do `> parentBounds.[width or height] / 2` it has weird effects depending where on the page the tooltip is used 🤦 .

I think this is better, but it isn't as robust to not needing the `key={Math.random}`:

```tsx
const placeTooltipLeft = parentBounds.width
  ? left + offsetLeft > parentBounds.width / 2
  : left + offsetLeft + ownBounds.width > window.innerWidth;
const placeTooltipUp = parentBounds.height
  ? top + offsetTop + ownBounds.height > parentBounds.height
  : top + offsetTop + ownBounds.height > window.innerHeight;
```

LethalPants:
My initial solution used `left > parentBounds.width / 2` which seemed to work almost fine but seeing the tooltip shift position at the halfway mark didn't feel natural 😅 
> So if you do `> parentBounds.[width or height] / 2` it has weird effects depending where on the page the tooltip is used 🤦.

Can you please elaborate on this, I couldn't recreate the issue you mentioned.

williaster:
@LethalPants yeah I agree that the shift at the midway point doesn't feel as great UX wise. okay what about this, it doesn't rely on `left/right/top/bottom`, swaps only at the edge, and swaps to the side where the fewest px will be clipped:

```tsx
let placeTooltipLeft = false;

if (parentBounds.width) {
  const rightPlacementClippedPx = left + offsetLeft + ownBounds.width - parentBounds.width;
  const leftPlacementClippedPx = ownBounds.width - left - offsetLeft;
  placeTooltipLeft =
    rightPlacementClippedPx > 0 && rightPlacementClippedPx > leftPlacementClippedPx;
} else {
  placeTooltipLeft = left + offsetLeft + ownBounds.width > window.innerWidth;
}

const placeTooltipUp = parentBounds.height
  ? top + offsetTop + ownBounds.height > parentBounds.height
  : top + offsetTop + ownBounds.height > window.innerHeight;
```

> I'm not sure whether the tooltip is supposed to stay in the bounds of the plot but that isn't the case in the code snippet you pasted.

When `Portal` is enabled, the tooltip is rendered outside of the container, and so its true parent container is `<body />`. So in this case it's _not_ supposed to detect its _immediate container's bounds_, instead it should detect the page bounds.

> > So if you do > parentBounds.[width or height] / 2 it has weird effects depending where on the page the tooltip is used 🤦.
>
> Can you please elaborate on this, I couldn't recreate the issue you mentioned.

This is confusing, sorry. It only applies to the `Portal` case where `parentRect` is the `<body />`, in some of the solutions we discussed where we flipped position based on parent width/height, there was weird behavior on `/gallery` and `/docs/tooltip` where the tooltip example is further down the page. I think we don't have to worry about this if we go with the above solution. sorry this turned out to be more complicated than initially anticipated 🙉 .

LethalPants:
Sorry for the delay.
> what about this, it doesn't rely on left/right/top/bottom, swaps position only when it encounters the right/bottom edge, and swaps to the side where the fewest px will be clipped (the original bug).

This seems fine. 

>for very wide tooltips that don't fit on either side this will swap the position at the parent's mid-width point (in this case the user should render the Tooltip in a Portal, or make its width responsive) 

I think changing to `Portal` would be a better option, incase the tooltip is wider than `parentBounds.width * 0.75` we can switch it to `Portal`, sounds good?

williaster (approved):
@LethalPants this looks great 👌  going to go ahead and merge, thanks for all the iteration on this!

> I think changing to Portal would be a better option, incase the tooltip is wider than parentBounds.width * 0.75 we can switch it to Portal, sounds good?

I think auto-switching is pretty tricky at the moment because rendering in the `Portal` currently requires a `ResizeObserver` polyfill so it'd be a breaking change. I think in most cases a `Portal` is actually more ideal, and `useTooltipInPortal` is compatible with `TooltipWithBounds` so users can leverage these enhancements.

PR #838, opened 2020-10-02
Remove code duplication

anastasiia-gontarieva:
#### 🏠 Internal

The body of this case clause on a line 26 duplicates the body of this case clause on a line 14. This may be caused by a copy-paste error. If this usage is intentional, consider using fall-through to remove code duplication.

### DIFF
--- a/packages/visx-pattern/src/patterns/Lines.tsx
+++ b/packages/visx-pattern/src/patterns/Lines.tsx
@@ -11,8 +11,6 @@ export function pathForOrientation({
   orientation: PatternOrientationType;
 }) {
   switch (orientation) {
-    case PatternOrientation.vertical:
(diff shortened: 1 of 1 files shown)


williaster:
Hi @anastasiia-gontarieva thanks for the PR. Happy to incorporate this, but for readability let's keep the case you removed and just merge it with the default. something like

```tsx
case PatternOrientation.vertical:
default:
  return `M ${height / 2}, 0 l 0, ${height}`;
```

anastasiia-gontarieva:
@williaster thanks. I made suggested changes. Could you please have a look when you have a free sec and checks have completed?

hshoff (approved):
👍 thanks for the contribution!

PR #841, opened 2020-10-05
fix(computeStats) #427 Exception thrown when firstQuartile and thirdQuartile are equal

### LINKED ISSUE #427: Exception thrown when firstQuartile and thirdQuartile are equal in @vx/stats/computeStats
In a way, it's a very uncommon case. On the other hand it can happen more often than not if your dashboard lets users filter data freely.

My data array turned out as follows: `[10000, 2400, 10000, 10000]`

If I feed it into `computeStats`, the function will compute:
`firstQuartile: 10000`
`thirdQuartile: 10000`
`IQR: 0` (firstQuartile - thirdQuartile)

Now, IQR is used as a multiplier and as a divisor in several places, turning up 0's and NaN's that eventually just crash the program.

Should computeStats "exit gracefully", for example returning an empty `{ boxPlot, binData }` object and a readable console error stating that the stats provided do not allow to compute a boxPlot?

I guess I can easily wrap the function call in a try/catch block for now, but I wouldn't rely on the fact that every developer out there will test their app with such a strange array. But their apps will crush in production once fed with real world data...

LethalPants:
#### :bug: Bug Fix

- Fixes #427 by changing method to calculate `firstQuartile` and `thirdQuartile`

test-result
*Test result*

### DIFF
--- a/packages/visx-stats/src/util/computeStats.ts
+++ b/packages/visx-stats/src/util/computeStats.ts
@@ -3,8 +3,23 @@ import { BoxPlot, BinDatum } from '../types';
 export default function computeStats(numericalArray: number[]) {
   const points = [...numericalArray].sort((a, b) => a - b);
   const sampleSize = points.length;
-  const firstQuartile = points[Math.round(sampleSize / 4)];
-  const thirdQuartile = points[Math.round((3 * sampleSize) / 4)];
+
+  function calcMedian(dataSet: number[]) {
+    const half = Math.floor(dataSet.length / 2);
+    if (dataSet.length % 2) return dataSet[half];
+    return (dataSet[half - 1] + dataSet[half]) / 2;
+  }
+  const median = calcMedian(points);
+
+  // calculate median of first half i.e. firstQuartile
+  const lowerHalfLength = Math.floor(sampleSize / 2);
+  const lowerHalf = points.slice(0, lowerHalfLength);
+  const firstQuartile = calcMedian(lowerHalf);
+
+  // calculate median of first half i.e. secondQuartile
+  const upperHalfLength = Math.ceil(sampleSize / 2);
+  const upperHalf = points.slice(upperHalfLength);
+  const thirdQuartile = calcMedian(upperHalf);
   const IQR = thirdQuartile - firstQuartile;
 
   const min = firstQuartile - 1.5 * IQR;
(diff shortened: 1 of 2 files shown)


packages/visx-stats/src/util/computeStats.ts
```diff
@@ -3,8 +3,23 @@ import { BoxPlot, BinDatum } from '../types';
 export default function computeStats(numericalArray: number[]) {
   const points = [...numericalArray].sort((a, b) => a - b);
   const sampleSize = points.length;
-  const firstQuartile = points[Math.round(sampleSize / 4)];
-  const thirdQuartile = points[Math.round((3 * sampleSize) / 4)];
+
+  function calcMedian(dataSet: number[]) {
```
williaster, line 7: since it has no close dependencies, this could be pulled outside the `computeStats` function

williaster (commented):
Hey @LethalPants thanks for the fix! I had one suggestion but otherwise lgtm.

hshoff:
One thing that tripped me up on this is `computeStates(data).boxPlot.min` and `computeStates(data).boxPlot.max` are not the min/max value in the dataset but the the low (<var>**Q**<sub>**1**</sub></var> - <var>**1.5**</var> * <var>**IQR**</var>) & high (<var>**Q**<sub>**3**</sub></var> + <var>**1.5**</var> * <var>**IQR**</var>) values which are used for calculating the outliers and **are** the min/max when showing outliers. But when there are no outliers min/max should be the `Math.min(data)` and `Math.max(data)`.

 For example:

```js
data = [10000, 2400, 10000, 10000]
computeStats(data).boxPlot;
{
  "min": 500,
  "firstQuartile": 6200,
  "median": 10000,
  "thirdQuartile": 10000,
  "max": 15700,
  "outliers": []
}
```

The [expected output](https://www.wolframalpha.com/input/?i=boxplot%28%5B10000%2C2400%2C10000%2C10000%5D%29) should be
```js
{
  "min": 2400,
  "firstQuartile": 6200,
  "median": 10000,
  "thirdQuartile": 10000,
  "max": 10000,
  "outliers": []
}
```

No need to fix this here with this PR, but wanted to flag it as I stumbled on this while reviewing.

williaster (approved):
thanks again @LethalPants, looks good.

I created #851 to track the min/max issue @hshoff noted.

PR #846, opened 2020-10-06
new(xychart): refine TooltipContext + Tooltip functionality

williaster:
**TODO**
- [x] add tests (separate PR for review #852)

#### :rocket: Enhancements

**`@visx/tooltip`**
allow function signatures for `useTooltip`'s `updateTooltip` + `showTooltip`
```tsx
// before
showTooltip({ tooltipData, ... });

//after 
showTooltip(currentState => ({ _logic based on currentState_ }));
```

**why?** this mirrors how `react` `setState` works, and was needed to implement tooltip data merging properly in `xychart/TooltipProvider`. when `TooltipProvider` tries to merge `tooltipData` when the hook state is updated in quick succession, the hook-provided value may be stale.

<hr />

**`@visx/xychart`**
Builds on #825 and adds more fully-featured `Tooltip` functionality:

- `TooltipProvider` no longer simply returns the `useTooltip` hook state, but adds additional functionality so that `TooltipContext.tooltipData` now includes both the _**globally closest datum**_ as well as the closest datum **_per-Series_**. Functionally this allows
  - for more comprehensive tooltip data to be rendered in the tooltip (i.e., data for _**all**_ Series, see demo)
  - `Tooltip` to snap to data values instead of simply following the cursor
- `Tooltip` now supports the following `props`
  - `snapTooltipToDatumX`, `snapTooltipToDatumY` (tooltip position snaps to data coord instead of mouse coord)
  - `showVerticalCrosshair`, `showHorizontalCrosshair`
  - `showDatumCircle`, `showSeriesCircles`
  - `verticalCrosshairStyle`, `horizontalCrosshairStyle`, `circleStyle`
- adds a `debounce` to `hideTooltip` for better UX

<hr />

**`@visx/demo`**

Updates the `XYChart` demo with all `Tooltip` controls

Also want to demo how you can do cool things like linked tooltips trivially with a shared `EventEmitterContext` which is TIGHT 🔥🔥

@kristw @hshoff @techniq

### DIFF
--- a/packages/visx-xychart/src/components/Tooltip.tsx
+++ b/packages/visx-xychart/src/components/Tooltip.tsx
@@ -7,6 +7,7 @@ import { UseTooltipPortalOptions } from '@visx/tooltip/lib/hooks/useTooltipInPor
 import TooltipContext from '../context/TooltipContext';
 import DataContext from '../context/DataContext';
 import { TooltipContextType } from '../types';
+import getScaleBandwidth from '../utils/getScaleBandwidth';
 
 const CROSSHAIR_STYLE: React.CSSProperties = {
   position: 'absolute',
@@ -26,6 +27,9 @@ export type TooltipProps<Datum extends object> = {
    * Content will be rendered in an HTML parent.
    */
   renderTooltip: (params: RenderTooltipParams<Datum>) => React.ReactNode;
+  /** */
+  snapTooltipToDatumX?: boolean;
+  snapTooltipToDatumY?: boolean;
   /** Whether to show a vertical line at tooltip position. */
   showVerticalCrosshair?: boolean;
   /** Whether to show a horizontal line at tooltip position. */
@@ -59,18 +63,21 @@ const INVISIBLE_STYLES: React.CSSProperties = {
 export default function Tooltip<Datum extends object>({
   debounce,
   detectBounds,
+  horizontalCrosshairStyle,
+  pointStyle,
   renderTooltip,
   resizeObserverPolyfill,
   scroll = true,
   showHorizontalCrosshair = true,
   showPoint = true,
   showVerticalCrosshair = true,
+  snapTooltipToDatumX = false,
+  snapTooltipToDatumY = true,
   verticalCrosshairStyle,
-  horizontalCrosshairStyle,
-  pointStyle,
   ...tooltipProps
 }: TooltipProps<Datum>) {
-  const { colorScale, theme, innerHeight, innerWidth, margin } = useContext(DataContext) || {};
+  const { colorScale, theme, innerHeight, innerWidth, margin, xScale, yScale, dataRegistry } =
+    useContext(DataContext) || {};
   const tooltipContext = useContext(TooltipContext);
   const { containerRef, TooltipInPortal } = useTooltipInPortal({
     debounce,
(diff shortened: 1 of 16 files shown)


packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -7,18 +7,54 @@ import { UseTooltipPortalOptions } from '@visx/tooltip/lib/hooks/useTooltipInPor
...
+  /** Whether to snap tooltip + crosshair x-coord to the nearest Datum x-coord instead of the event x-coord. */
+  snapTooltipToDatumX?: boolean;
+  /** Whether to snap tooltip + crosshair y-coord to the nearest Datum y-coord instead of the event y-coord. */
+  snapTooltipToDatumY?: boolean;
+  /** Whether to show a vertical line at tooltip position. */
+  showVerticalCrosshair?: boolean;
+  /** Whether to show a horizontal line at tooltip position. */
+  showHorizontalCrosshair?: boolean;
+  /** Whether to show a circle at the tooltip position for the (single) nearest Datum. */
+  showDatumCircle?: boolean;
```
kristw, line 49: Perhaps name this `showDatumMark` or `showDatumMarker` ? In case you want to support other shapes (that is not circle) in the future.

packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -7,18 +7,54 @@ import { UseTooltipPortalOptions } from '@visx/tooltip/lib/hooks/useTooltipInPor
...
+  /** Whether to show a circle for the nearest Datum in each series. */
+  showSeriesCircles?: boolean;
```
kristw, line 51: `showSeriesMarks`? (similar reason with above)

packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -7,18 +7,54 @@ import { UseTooltipPortalOptions } from '@visx/tooltip/lib/hooks/useTooltipInPor
...
+  /** Optional styles for the vertical crosshair, if visible. */
+  verticalCrosshairStyle?: React.SVGProps<SVGLineElement>;
+  /** Optional styles for the vertical crosshair, if visible. */
+  horizontalCrosshairStyle?: React.SVGProps<SVGLineElement>;
+  /** Optional styles for the point, if visible. */
+  circleStyle?: React.SVGProps<SVGCircleElement>;
```
kristw, line 57: `markStyle`

packages/visx-xychart/src/components/series/LineSeries.tsx
```diff
@@ -34,7 +39,7 @@ function LineSeries<XScale extends AxisScale, YScale extends AxisScale, Datum ex
     (params: HandlerParams | undefined) => {
       const { svgPoint } = params || {};
       if (svgPoint && width && height && showTooltip) {
-        const datum = findNearestDatumXY({
+        const datum = (horizontal ? findNearestDatumX : findNearestDatumY)({
```
kristw, line 42: how to specify if the nearest XY is what developer want?

kristw (approved):

packages/visx-xychart/src/components/Tooltip.tsx
williaster: good call. maybe I should make these use `@visx/glyph` 🤔  maybe I'll do that as a followup PR.

packages/visx-xychart/src/components/series/LineSeries.tsx
williaster: these are controlled with the `horizontal` prop right now. I removed `xy` because I realized that series should use the same logic for x vs y. Previously if bar used `x` and line used `xy`, you might have mismatched `x` values in the tooltip context (because `voronoi` don't map cleanly to x, it might not match the same `x` that the bar finds).

I agree this needs a better hook for dev experience or could be provided by `DataContext` at some point (you suggested this in the POC, maybe it'd be straightforward to infer from scale types 🤔 ).

packages/visx-xychart/src/components/Tooltip.tsx
williaster: updated to `glyph` throughout with the future glyph support in mind

PR #852, opened 2020-10-07
test(xychart): add refined Tooltip, TooltipProvider tests

williaster:
#### :house: Internal

Builds on #846, adding tests for the new `Tooltip` + `TooltipProvider` functionality (new tests highlighted). Separated for review-ability.

@kristw @hshoff

### DIFF
--- a/packages/visx-xychart/test/components/Tooltip.test.tsx
+++ b/packages/visx-xychart/test/components/Tooltip.test.tsx
@@ -2,29 +2,42 @@ import React from 'react';
 import ResizeObserver from 'resize-observer-polyfill';
 import { mount } from 'enzyme';
 import { Tooltip as BaseTooltip } from '@visx/tooltip';
-import { Tooltip, TooltipContext, TooltipContextType } from '../../src';
+import { DataContext, Tooltip, TooltipContext, TooltipContextType } from '../../src';
 import { TooltipProps } from '../../src/components/Tooltip';
+import getDataContext from '../mocks/getDataContext';
 
 describe('<Tooltip />', () => {
   type SetupProps =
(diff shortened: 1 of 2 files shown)


packages/visx-xychart/test/components/Tooltip.test.tsx
```diff
@@ -2,29 +2,42 @@ import React from 'react';
...
+          value={{
+            tooltipOpen: false,
+            showTooltip: jest.fn(),
+            updateTooltip: jest.fn(),
+            hideTooltip: jest.fn(),
+            ...context,
+          }}
+        >
+          <Tooltip
+            resizeObserverPolyfill={ResizeObserver}
```
kristw, line 35: What happens if the polyfill is not passed?

kristw (approved):

packages/visx-xychart/test/components/Tooltip.test.tsx
williaster: if it's not passed and it's not available globally on `window`, it will throw (this logic lives in `react-use-measure). Throwing may seem extreme but I think it's the cleanest way to polyfill, we import it globally in `@visx/responsive`'s `ParentSize` and have gotten negative feedback about it.

I think we just need to be clear about it in the docs like in `@visx/tooltip` + `react-use-measure`.

kristw: ic. a bit more gear towards power users i guess. 
If doc emphasizes it enough probably ok. 
Perhaps can make it a bit more convenient later to do this once instead of passing to every `Tooltip` instance.

PR #853, opened 2020-10-07
fix(stats): update min/max in computeStats to handle no outliers

### LINKED ISSUE #851: [stats/computeStats] min/max should match the dataset when there are no outliers
Noted in #841:

When there are no outliers, min/max should be the `Math.min(data)` and `Math.max(data)` of the passed `data`. Currently `min/max` are always computed as the low (**<var>Q<sub>1</sub></var>** - **<var>1.5</var>** * **<var>IQR</var>**) & high (**<var>Q<sub>3</sub></var>** + **<var>1.5</var>** * **<var>IQR</var>**) values which are used for calculating the outliers. These **are** the min/max when showing outliers, but shouldn't be used when there are no outliers.

For example:

```js
data = [10000, 2400, 10000, 10000]
computeStats(data).boxPlot;
{
  "min": 500,
  "firstQuartile": 6200,
  "median": 10000,
  "thirdQuartile": 10000,
  "max": 15700,
  "outliers": []
}
```

The [expected output](https://www.wolframalpha.com/input/?i=boxplot%28%5B10000%2C2400%2C10000%2C10000%5D%29) should be

```js
{
  "min": 2400,
  "firstQuartile": 6200,
  "median": 10000,
  "thirdQuartile": 10000,
  "max": 10000,
  "outliers": []
}
```

kanhegaonkarsaurabh:
#### 🐛 Bug fix

Closes #851.

### DIFF
--- a/packages/visx-stats/test/computeStats.test.ts
+++ b/packages/visx-stats/test/computeStats.test.ts
@@ -1,26 +1,26 @@
-import { computeStats } from "../src";
+import { computeStats } from '../src';
 
 const data = [1, 2, 3, 4, 5, 6, 6, 7, 8, 9, 1];
 const edgeCaseData = [10000, 2400, 10000, 10000];
 
-describe("computeStats", () => {
-  test("it should be defined", () => {
+describe('computeStats', () => {
+  test('it should be defined', () => {
     expect(computeStats).toBeDefined();
   });
 
-  test("it should have boxPlot and binData", () => {
+  test('it should have boxPlot and binData', () => {
     const stats = computeStats(data);
     expect(stats.boxPlot).toBeDefined();
     expect(stats.binData).toBeDefined();
   });
 
-  test("it should have boxPlot and binData when first and third quartile are equal", () => {
+  test('it should have boxPlot and binData when first and third quartile are equal', () => {
     const stats = computeStats(edgeCaseData);
     expect(stats.boxPlot).toBeDefined();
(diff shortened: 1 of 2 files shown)


packages/visx-stats/test/computeStats.test.ts
```diff
@@ -1,22 +1,28 @@
-import { computeStats } from '../src';
+import { computeStats } from "../src";
```
williaster, line 1: I think the addition of double quotes throughout here is failing `prettier` in the build. you can update them manually or run `yarn format` from the monorepo root.

williaster (commented):
thanks for the fix @kanhegaonkarsaurabh! this looks good to me except the build is failing due to a prettier issue I commented on.

kanhegaonkarsaurabh:
@williaster Got it! I've updated the test file to include single quotes and fix the `prettier` issue

packages/visx-stats/test/computeStats.test.ts
kanhegaonkarsaurabh: Had an issue running `yarn format` due to a `nimbus` driver not found error. Thus, I ended up doing it manually

williaster: interesting, should work if you run `yarn` in the root to install all deps. in any case glad you got it!

williaster (approved):
thanks for the contribution @kanhegaonkarsaurabh 🙌 !

PR #856, opened 2020-10-07
build: bump @types/d3-scale

### LINKED ISSUE #855: d3-scale out of sync with typings
`@visx/scale` depends on `d3-scale@^3.0.1` and `@types/d3-scale@^2.1.1`. This means that a v3 scale passed to `@visx/axis` won't type-check correctly even though the runtime version is v3. It should depend on `@types/d3-scale@^3.0.0` instead.

https://github.com/airbnb/visx/blob/c0eb0b2aa5a8689e4b44d2642d56384a9a34ed30/packages/visx-scale/package.json#L33-L38

kristw:
#### :rocket: Enhancements

- bump `@types/d3-scale` 

Fix #855

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-area/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-area/Example.tsx
@@ -127,8 +127,8 @@ export default withTooltip<AreaProps, TooltipData>(
           />
           <AreaClosed<AppleStock>
             data={stock}
-            x={d => dateScale(getDate(d))}
-            y={d => stockValueScale(getStockValue(d))}
+            x={d => dateScale(getDate(d)) ?? 0}
+            y={d => stockValueScale(getStockValue(d)) ?? 0}
             yScale={stockValueScale}
             strokeWidth={1}
             stroke="url(#area-gradient)"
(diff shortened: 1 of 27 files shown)


williaster (approved):
thanks for making all the fixes needed for this 🙏 lgtm. may have to figure out if this will count as a breaking change or not.

razh:
Might be able to revert this? `@types/d3-scale` v3.2.1 no longer returns unknown values.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49196

williaster:
@kristw wdyt?

kristw:
> Might be able to revert this? `@types/d3-scale` v3.2.1 no longer returns unknown values.
> 
> [DefinitelyTyped/DefinitelyTyped#49196](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49196)

@razh Could you elaborate a bit more about your concerns with the `unknown` and why we should revert the change? 

Also, things are a bit more complicated as the previous version of the types was for `d3-scale` 2 and mistyped a few things and did not include several functions that are available API. I would prefer to have types that better match the `d3-scale` to avoid mistakes in the library.

razh:
It's mostly a matter of correctness versus convenience.

Given a known domain and range, `scaleLinear`/`scaleTime`/etc. should not return `undefined`.

This comment explains the rationale:

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/48299#issuecomment-717543887

I don't have a stake in the game here, just FYI!

PR #857, opened 2020-10-08
fix(tooltip): fallback to left/top positioning, add applyPositionStyle

### LINKED ISSUE #767: TooltipWithBounds bug

williaster:
#### :bug: Bug Fix

This partially reverts #828 (which fixed #767) because the re-implementation of in `Tooltip` to use `transform` instead of `top/left` for positioning breaks cases where users themselves set `style.transform` and would thus be a 💥  breaking change. This was true in the `/areas` demo:

Before

After

#### :rocket: Enhancements

**`@visx/tooltip`** 
- adds a new `applyPositionStyle` which applies `position: absolute` so that users who set `unstyled=true` don't have to set this themselves via a CSS selector 🤦 . In a future breaking change this will be the default behavior.

@hshoff @kristw

### DIFF
--- a/packages/visx-tooltip/src/tooltips/Tooltip.tsx
+++ b/packages/visx-tooltip/src/tooltips/Tooltip.tsx
@@ -19,7 +19,7 @@ export type TooltipProps = {
    * Whether to omit applying any style, except `left` / `top`.
    * In most cases if this is `true` a developer must do one of the following
    * for positioning to work correctly:
-   * - set `applyPositionStyles=true`
+   * - set `applyPositionStyle=true`
    * - create a CSS selector like: `.visx-tooltip { position: 'absolute' }`
    */
   unstyled?: boolean;
--- a/packages/visx-tooltip/src/tooltips/Tooltip.tsx
+++ b/packages/visx-tooltip/src/tooltips/Tooltip.tsx
@@ -2,13 +2,19 @@ import React from 'react';
 import cx from 'classnames';
 
 export type TooltipProps = {
+  /** Tooltip content. */
   children?: React.ReactNode;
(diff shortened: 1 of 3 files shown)


packages/visx-tooltip/src/tooltips/TooltipWithBounds.tsx
```diff
@@ -3,23 +3,20 @@ import { withBoundingRects, WithBoundingRectsProps } from '@visx/bounds';
...
-export type TooltipWithBoundsProps = {
-  offsetLeft?: number;
-  offsetTop?: number;
-} & TooltipProps &
+export type TooltipWithBoundsProps = TooltipProps &
   React.HTMLProps<HTMLDivElement> &
   WithBoundingRectsProps;
 
 function TooltipWithBounds({
+  children,
```
williaster, line 11: abc

packages/visx-tooltip/src/tooltips/Tooltip.tsx
```diff
@@ -2,13 +2,26 @@ import React from 'react';
 import cx from 'classnames';
 
 export type TooltipProps = {
+  children?: React.ReactNode;
```
williaster, line 5: abc

packages/visx-tooltip/src/tooltips/TooltipWithBounds.tsx
kristw: what is abc?

williaster: sorry updated

kristw (approved):

PR #859, opened 2020-10-08
fix(network): apply fill and radius props to circle element in DefaultNode

tombarton:
#### Bug Fix

Apply the `r` and `fill` props to the `circle` svg element. Please feel free to close if this isn't the direction the library was planning to go, but it seems like one might wish to use the DefaultNode but simply change the radius and colour, especially as it was already being defaulted to a value.


williaster (approved):
thanks for the fix @tombarton! 🙏  I think this must have just been an oversight.

PR #864, opened 2020-10-09
add touch events for brush component

### LINKED ISSUE #845: Brush interaction not working on mobile
(Enjoying the library, thanks for making it available)

I noticed the brush doesn't work with touch interactivity on mobile.  Example here: https://vx-demo.now.sh/brush

After digging into the code a bit, I noticed that mobile events don't seem to be wired up for the `Drag` component.  Shouldn't drag include touch and touchmove the same way as mouse and mousemove?

https://github.com/airbnb/visx/blob/master/packages/visx-drag/src/Drag.tsx#L121-L122

I could give it a shot if this is unintended and no one has time to look into it or is currently doing so.  Lmk

BrianRosamilia:
Fixes #845 

The fix seems fairly simple and I think covers all scenarios.

One thing to note: it's hard to DRY out this code into member functions because of the hooks that the event handlers rely on :O I could create local functions in the jsx expression itself to DRY it up a little bit, or use higher order member functions, but I figured that this PR is actually the simplest/best to at least get started on fixing this and get some feedback first since I am totally new to this project 😄 

#### :boom: Breaking Changes

- none

#### :bug: Bug Fix

- adds touch events for brush component

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -376,6 +376,7 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
           onDragMove={this.handleDragMove}
           onDragEnd={this.handleDragEnd}
         >
+          {/* resize */}
           {({ dragStart, isDragging, dragMove, dragEnd }) => (
             <Bar
               className="visx-brush-overlay"
@@ -401,6 +402,19 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
                 if (isDragging) dragMove(event);
               }}
               onMouseUp={(event: MouseHandlerEvent) => {
+                this.mouseUpTime = Date.now();
+                if (onMouseUp) onMouseUp(event);
+                dragEnd(event);
+              }}
+              onTouchStart={(event: MouseHandlerEvent) => {
+                this.mouseDownTime = Date.now();
+                dragStart(event);
+              }}
+              onTouchMove={(event: MouseHandlerEvent) => {
+                if (!isDragging && onMouseMove) onMouseMove(event);
+                if (isDragging) dragMove(event);
+              }}
+              onTouchEnd={(event: MouseHandlerEvent) => {
                 this.mouseUpTime = Date.now();
                 if (onMouseUp) onMouseUp(event);
                 dragEnd(event);
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -376,7 +376,6 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
           onDragMove={this.handleDragMove}
           onDragEnd={this.handleDragEnd}
         >
-          {/* resize */}
           {({ dragStart, isDragging, dragMove, dragEnd }) => (
             <Bar
               className="visx-brush-overlay"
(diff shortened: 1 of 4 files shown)


sakulstra:
just wanted to report this as an issue, glad a fix is already on the way :+1:

>  I could create local functions in the jsx expression itself to DRY it up a little bit

I'd personally prefer that as it makes the code easier to follow and is in line with how the non functional components handle it.
```js
const handleStart = (event: MouseHandlerEvent) => {
  this.mouseDownTime = Date.now();
  dragStart(event);
}
```

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -405,6 +405,19 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
                 if (onMouseUp) onMouseUp(event);
                 dragEnd(event);
               }}
+              onTouchStart={(event: MouseHandlerEvent) => {
```
williaster, line 408: on thought I have here for DRY-ness is to update all of these to `PointerEvent` handlers, which are a combination of `MouseEvent`s + `TouchEvent`s introduced in [`react@16.4`](https://reactjs.org/blog/2018/05/23/react-v-16-4.html).  so `onMouseDown => onPointerDown`, `onMouseLeave => onPointerLeave`, `onMouseMove => onPointerMove`. I think that `MouseHandlerEvent` will also need to be updated to support `React.PointerEvent`. 

To avoid breaking changes we could could keep the props as `onMouseXXX`, but could consider also adding `onPointerXXX` handlers to match, and then deprecate `onMouseXXX` in a future `2.0` release. This would also require that we bump the `react` `peerDependency` in `@visx/brush` to `^16.4.0`, which is theoretically breaking but I just checked `@visx/drag` (which brush relies on) and it already requires `^16.8.0` so I think technically `@visx/brush` should already actually be `^16.8.0`.

cc @hshoff

packages/visx-drag/src/Drag.tsx
```diff
@@ -120,6 +120,8 @@ export default class Drag extends React.Component<DragProps, DragState> {
             height={height}
             onMouseMove={this.handleDragMove}
             onMouseUp={this.handleDragEnd}
+            onTouchMove={this.handleDragMove}
```
williaster, line 123: Since this change `Drag` was re-implemented/re-factored to use the `useDrag` hook so there are unfortunately conflicts now. I think an easy change might be to simply update `onMouseMove/onMouseUp` to `onPointerMove/onPointerUp`, similar to my suggestion in `@visx/brush`

packages/visx-brush/src/BrushSelection.tsx
```diff
@@ -146,6 +148,15 @@ export default class BrushSelection extends React.Component<
               onClick={event => {
                 if (onClick) onClick(event);
               }}
+              onTouchStart={disableDraggingSelection ? undefined : dragStart}
```
williaster, line 151: same comment on remove these additions and updating the above `mouse` handlers to the `pointer` equivalents 👍

williaster (commented):
Hey @BrianRosamilia apologies for such a delay in reviewing this, we really appreciate the PR! 🙌 

I had a couple suggestions here for DRY-ness, lmk what you think and I'll keep an eye out for any updates to merge this ASAP hopefully for the `1.3.0` release.

williaster:
Hi @BrianRosamilia just wanted to ping this PR again to see if we can get it past the finish line, or if I should take over. 

Thanks & happy New Year!

RemaaBdair:
@williaster we've been facing the same issue, and we need this pr to get approved and merged ASAP. Is there any way to speed things up? 
Should we do the reviews in another pr?

RemaaBdair:
@williaster Hi again, i created a new pr to address the reviews if you could check it please: https://github.com/airbnb/visx/pull/1155
Thank you

williaster:
This will be superseded by #1155

PR #865, opened 2020-10-09
new(xychart): add BarStack series

williaster:
**TODO**
- [x] Add tests (separate PR for reviewability #866 )

#### :rocket: Enhancements

This PR adds `BarStack` to `@visx/xychart` with an API that wraps `BarSeries` (this was planned in the POC, but could simplify in the future): 

```tsx
<BarStack>
  <BarSeries dataKey="sf" data={..} />
  <BarSeries dataKey="ny" data={..} />
</BarStack>
```

It also updates the `/xychart` demo
- adds option to render `BarStack` (in the demo this is not compatible with `LineSeries`/`BarSeries` because they share `dataKey`s so controls are disabled accordingly)
- adds option to render _negative values_. this lets me test that negative stacks work correctly (ditto for other Series types)

**Tooltip**

**Negative values**

**Horizontal**

Some notes on the implementation
- The general flow is:
  1) collect `props.data` from child `BarSeries`
  2) create combined stacked data structure (using child `props.dataKey`)
  3) register data in the `DataContext` for each child `dataKey`, but update `data` + `x/yAccessor` to use the stacked data (so that scales are computed correctly)
  4) render `BarStack` using data + scales from `DataContext`

- I first wrote this using `@visx/shape`s `BarStack(Horizontal)`, but it doesn't allow us to use the stacked data independently of their scaled values (i.e., it creates the stack data structure **and** scales the stacked values). For the `DataContext` scales + `findNearestDatum` functions (`TooltipContext`), we need to add the stacked data itself as described in prev point

- `BarStack`s implementation is currently is incompatible with `Tooltip.snapTooltipToDatumX/Y`. 
  - This is tricky because `Tooltip` attempts to use `DataContext` for `data` + `x/yAccessor`s to map the `TooltipContext.tooltipData` to the scaled positions. However `tooltipData` is typed as `Datum`, and the `x/yAccessor`s + `x/yScale`s from `DataContext` use the stacked `SeriesDatum` type. Will noodle on this more.

@kristw @techniq

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
-import cityTemperature, { CityTemperature } from '@visx/mock-data/lib/mocks/cityTemperature';
+import { CityTemperature } from '@visx/mock-data/lib/mocks/cityTemperature';
 import {
   AnimatedAxis,
   AnimatedGrid,
   DataProvider,
   BarSeries,
+  BarStack,
   LineSeries,
   Tooltip,
   XYChart,
@@ -17,20 +18,19 @@ type Props = {
   height: number;
 };
 
-const xScaleConfig = { type: 'band', paddingInner: 0.3 } as const;
-const yScaleConfig = { type: 'linear' } as const;
-const numTicks = 4;
-const data = cityTemperature.slice(150, 225);
-const getDate = (d: CityTemperature) => d.date;
-const getSfTemperature = (d: CityTemperature) => Number(d['San Francisco']);
-const getNyTemperature = (d: CityTemperature) => Number(d['New York']);
+type City = 'San Francisco' | 'New York' | 'Austin';
 
 export default function Example({ height }: Props) {
   return (
     <ExampleControls>
       {({
+        accessors,
         animationTrajectory,
+        config,
+        data,
+        numTicks,
         renderBarSeries,
+        renderBarStack,
         renderHorizontally,
         renderLineSeries,
         sharedTooltip,
@@ -45,11 +45,7 @@ export default function Example({ height }: Props) {
         xAxisOrientation,
         yAxisOrientation,
       }) => (
-        <DataProvider
-          theme={theme}
-          xScale={renderHorizontally ? yScaleConfig : xScaleConfig}
-          yScale={renderHorizontally ? xScaleConfig : yScaleConfig}
-        >
+        <DataProvider theme={theme} xScale={config.x} yScale={config.y}>
           <XYChart height={Math.min(400, height)}>
             <CustomChartBackground />
             <AnimatedGrid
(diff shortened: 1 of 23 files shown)


packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
```diff
@@ -17,20 +18,19 @@ type Props = {
   height: number;
 };
 
-const xScaleConfig = { type: 'band', paddingInner: 0.3 } as const;
```
williaster, line 20: I moved a lot of the `Example` params into `ExampleControls`

packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
```diff
@@ -40,24 +74,63 @@ export default function ExampleControls({ children }: ControlsProps) {
...
+        renderLineSeries: renderBarOrBarStack === 'bar' && renderLineSeries,
         sharedTooltip,
         showGridColumns,
         showGridRows,
         showHorizontalCrosshair,
         showTooltip,
         showVerticalCrosshair,
-        snapTooltipToDatumX,
-        snapTooltipToDatumY,
+        snapTooltipToDatumX: renderBarOrBarStack === 'bar' && snapTooltipToDatumX,
```
williaster, line 132: moving stuff in here allows us to handle couple control logic like this

packages/visx-xychart/src/components/series/BarStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const barThickness = getBandwidth(horizontal ? yScale : xScale);
+
+  return (
+    <g className="visx-bar-stack">
+      {stackedData.map((barStack, stackIndex) => {
+        const entry = dataRegistry.get(barStack.key);
+
+        return entry
+          ? barStack.map((bar, index) => {
```
williaster, line 174: this mirrors the logic in `BarStack/BarStackHorizontal`

kristw: Perhaps do `horizontal` and other conditions outside the loop when possible.

```ts
const thickness = getBandwidth(horizontal ? yScale : xScale);
const halfThickness = thickness / 2;

let widthFn;
let heightFn;
let xFn;
let yFn;

if (horizontal) {
  widthFn = bar => (xScale(getSecondItem(bar)) || 0) - (xScale(getFirstItem(bar)) || 0);
  heightFn = () => thickness;
  xFn = bar => xScale(getFirstItem(bar));
  yFn = 'bandwidth' in yScale 
    ? bar => yScale(getStackValue(bar.data))
    : bar => Math.max((yScale(getStackValue(bar.data)) || 0) - halfThickness);
} else {
  widthFn = () => thickness;
  heightFn = bar => (yScale(getFirstItem(bar)) || 0) - (yScale(getSecondItem(bar)) || 0);
  xFn = 'bandwidth' in xScale
    ? bar => xScale(getStackValue(bar.data))
    : bar => Math.max((xScale(getStackValue(bar.data)) || 0) - halfThickness);
} 

return (
  <g className="visx-bar-stack">
    {stackedData.map((barStack, stackIndex) => {
      const entry = dataRegistry.get(barStack.key);
      if (!entry) return null;

      const y = yFn ?? (bar => yScale(entry.yAccessor(bar)));

      return barStack.map((bar, index) => {
        const barX = xFn(bar);
        if (!isValidNumber(barX)) return null;
        const barY = y(bar);
        if (!isValidNumber(barY)) return null;

        return (
          <rect
            key={`${stackIndex}-${barStack.key}-${index}`}
            x={barX ?? 0}
            y={barY ?? 0}
            width={widthFn(bar)}
            height={heightFn(bar)}
            fill={colorScale(barStack.key)}
            stroke="transparent"
          />
        );        
      });
    })}
  </g>
);
```

packages/visx-xychart/src/utils/combineBarStackData.ts
```diff
@@ -0,0 +1,46 @@
...
+  const dataByStackValue: {
+    [stackValue: string]: CombinedStackData<XScale, YScale>;
+  } = {};
+
+  barSeriesChildren.forEach(child => {
+    const { dataKey, data, xAccessor, yAccessor } = child.props;
+
+    // this should exist but double check
+    if (!xAccessor || !yAccessor) return;
+
```
kristw, line 32: Can do condition outside of loop
```ts
const [stackFn, valueFn] = horizontal ? [xAccessor, yAccessor] : [yAccessor, xAccessor];
```

packages/visx-xychart/src/utils/getBarStackRegistryData.ts
```diff
@@ -0,0 +1,48 @@
...
+  stackedData: BarStackData<XScale, YScale>,
+  comprehensiveDomain: [number, number],
+  horizontal?: boolean,
+) {
+  return stackedData
+    .map((data, index) => {
+      const entry: DataRegistryEntry<XScale, YScale, BarStackDatum<XScale, YScale>> = {
+        key: data.key,
+        data,
+        xAccessor: horizontal ? getNumericValue : getStack,
```
kristw, line 25: Could do condition before map

packages/visx-xychart/src/utils/getBarStackRegistryData.ts
```diff
@@ -0,0 +1,48 @@
...
+        yAccessor: horizontal ? getStack : getNumericValue,
+      };
+
+      // update the numeric domain to account for full data stack
+      // only need to do this for one key
+      if (comprehensiveDomain.length > 0 && index === 0) {
+        if (horizontal) {
+          entry.xScale = scale =>
+            scale.domain(
+              extent([...scale.domain(), ...comprehensiveDomain], d => d),
```
kristw, line 35:
```ts
scale.domain().concat(comprehensiveDomain)
``` 
is faster than spread if I remember correctly.

Also, the second argument of `extent` is optional (from doc).

kristw (approved):
Mostly minor perf suggestion.

packages/visx-xychart/src/components/series/BarStack.tsx
```diff
@@ -0,0 +1,209 @@
...
+                ? (xScale(getSecondItem(bar)) || 0) - (xScale(getFirstItem(bar)) || 0)
+                : (yScale(getFirstItem(bar)) || 0) - (yScale(getSecondItem(bar)) || 0);
+
+              const barY = horizontal
+                ? 'bandwidth' in yScale
+                  ? yScale(getStackValue(bar.data))
+                  : Math.max((yScale(getStackValue(bar.data)) || 0) - barThickness / 2)
+                : yScale(entry.yAccessor(bar));
+
+              const barX: number | undefined = horizontal
```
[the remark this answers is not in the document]
williaster: type was an oversight, I think we have nested ternary disabled in this repo because it's so often useful

packages/visx-xychart/src/components/series/BarStack.tsx
williaster: great suggestion, also the special `yFn` handling made me realize I didn't actually need the `entry` accessor function 👍

PR #867, opened 2020-10-13
refactor: remove dead link and add one link

ArvinH:
#### :memo: Documentation

- Remove some dead link in the readme (some of them might just need to re-deploy 🤔  so if we shouldn't remove them, I can put them back)
- Add one link of mine, the UI of the demo and the tutorial is written in Traditional Chinese and has been shared by the newsletter I'm running

screencast


hshoff (approved):
Awesome! thanks for the contribution and great work on the demo!

PR #892, opened 2020-10-23
Remove data requirement from arc

### LINKED ISSUE #886: Arc example
I've been using visx for some time now but I've never actually been able to render an Arc. I'm not sure whether it's a bug or if I'm doing something wrong.

The Arc can be quite difficult one to understand. It has tons of radius and angle attributes. Even when passing in random arguments like this:
```html
<svg width={250} height={250}>
  <Arc
      startAngle={10}
      endAngle={180}
      innerRadius={30}
      cornerRadius={30}
      outerRadius={50}
      padAngle={10}
      padRadius={10}
      fill={"#FF0000"}
   ></Arc>
</svg>
```
I can't get anything to show up. (Not even anything in Inspect Element which is why I have the idea that it might just be a bug)

It'd be nice to have a simple example, which I could copy paste and then modify to understand what each attribute does.

oscar-broman:
`arc` from d3-shape does not take any arguments: https://github.com/d3/d3-shape/blob/master/src/arc.js#L87

Fixes #886

### DIFF
--- a/packages/visx-shape/src/shapes/Arc.tsx
+++ b/packages/visx-shape/src/shapes/Arc.tsx
@@ -7,8 +7,6 @@ import { arc } from '../util/D3ShapeFactories';
 export type ArcProps<Datum> = {
   /** className applied to path element. */
   className?: string;
-  /** A Datum for which to generate an arc. */
-  data?: Datum;
   /** Override render function which is passed the configured arc generator as input. */
   children?: (args: { path: ArcType<$TSFIXME, Datum> }) => React.ReactNode;
   /** React ref to the path element. */
@@ -17,7 +15,6 @@ export type ArcProps<Datum> = {
 
 export default function Arc<Datum>({
   className,
-  data,
   innerRadius,
   outerRadius,
   cornerRadius,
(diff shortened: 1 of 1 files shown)


williaster (changes requested):
From my comment in #886 

> I think `data` should be optional, not removed because it supports both types of usage. 
>
> `arc` optionally takes arguments [as you can see in the readme](https://github.com/d3/d3-shape/#_arc) and it's confusing but a few lines down from the line you linked you can see that [`arguments` are passed along to the accessors](https://github.com/d3/d3-shape/blob/master/src/arc.js#L90).
>
> `Arc` is functionally as it is used in the [`/chord` demo](https://airbnb.io/visx/chord). [If you omit the `data` prop, it fails to draw anything](https://codesandbox.io/s/bold-goldstine-muqc2?file=/Example.tsx). I think this is confusing because the `d3` API supports two types of usage (and thus so does `visx`):

```ts
const arc1 = d3.arc();
arc1({ innerRadius, ... }); // "M0,-100A100,100,0,0,1,100,0L0,0Z"

const arc2 = d3.arc().innerRadius(...)...;
arc2(); // "M0,-100A100,100,0,0,1,100,0L0,0Z"
```

packages/visx-shape/src/shapes/Arc.tsx
```diff
@@ -41,13 +38,12 @@ export default function Arc<Datum>({
 
   // eslint-disable-next-line react/jsx-no-useless-fragment
   if (children) return <>{children({ path })}</>;
-  if (!data) return null;
```
williaster, line 44: since `data` is already optional, I think this is probably the only line that needs to be removed since it is valid to call `path` without arguments.

williaster:
This has been inactive for a while, closing in favor of #937

PR #898, opened 2020-10-28
new(xychart): handle rendering + tweening missing values

williaster:
#### :house: Internal

This PR is a mix of improvements + new functionality for `xychart` which are all related by the rabbit hole 🐰 ⚫    I went down trying to fix one problem:

> when transitioning between `horizontal` and `vertical` orientation in the `/xychart` demo, or transitioning between more or fewer data points, `AnimatedPath` would throw due to a cryptic `react-spring` error

The improvements are:
1) does a sweep to handle missing values across all `*Series` and in scale domain calculation
  - tested by updating the `/xychart` demo 
2) (✅  fix for above error) updates `AnimatedPath` to handle missing values and a _**changing number of values**_ by using [`d3-interpolate-path`](https://github.com/pbeshai/d3-interpolate-path) for interpolating `path` `d` attributes
  - caveats: the only animation case this doesn't handle well is `Area` shapes with missing data (see below) but I think it's the best option
  - 💬  reasoning
    - `react-spring` does this beautifully [**except** when the # of points changes](https://www.react-spring.io/docs/props/spring) which is the cause of the above error 😭  
      - I spent a long time trying to test `react-spring@9.0.0-rc.3` to see if it did ... but it does not
    - `d3-interpolate` does not handle path string interpolation well (well documented see e.g., [here](https://bocoup.com/weblog/improving-d3-path-animation))
    - [`flubber`](https://github.com/veltman/flubber) is the best at handling all edge cases, but it assumes _**closed polygon shapes**_ (doesn't apply for lines) and its bundle size is 10x what `d3-interpolate-path`s is.
3) moves `props.horizontal` from all `*Series` types to `DataProvider`s `context.horizontal` (the start of the 🐰 ⚫  )
  - `horizontal` determines orientation for `Bar`-type series, and it controls nearest `datum` logic for mouse events across _all `*Series`_
  - the goal of this was to make it easier for `dev`s, most of the time `horizontal` can be inferred from `scale` types and this could be done once in context vs specified / computed in each `*Series`

Note issue with missing area data

@kristw @techniq

### DIFF
--- a/packages/visx-xychart/src/components/series/private/AnimatedPath.tsx
+++ b/packages/visx-xychart/src/components/series/private/AnimatedPath.tsx
@@ -1,6 +1,6 @@
 import React, { useRef } from 'react';
 import { animated, useSpring } from 'react-spring';
-// @ts-ignore
+// @ts-ignore no types
 import { interpolatePath } from 'd3-interpolate-path';
 
 export default function AnimatedPath({
@@ -15,11 +15,16 @@ export default function AnimatedPath({
   // d3-interpolate-path is better at interpolating extra/fewer points so we use that
   const interpolator = interpolatePath(previousD.current, d);
   previousD.current = d;
-  const frame = useSpring({ from: { t: 0 }, to: { t: 1 }, reset: true });
+  // @ts-ignore t is not in CSSProperties
+  const { t } = useSpring({
+    from: { t: 0 },
+    to: { t: 1 },
+    reset: true,
+  });
   const tweened = useSpring({ stroke, fill });
   return (
     <animated.path
-      d={frame.t.interpolate(interpolator)}
+      d={t.interpolate(interpolator)}
       stroke={tweened.stroke}
       fill={tweened.fill}
       {...lineProps}
--- a/packages/visx-xychart/src/components/series/private/AnimatedPath.tsx
+++ b/packages/visx-xychart/src/components/series/private/AnimatedPath.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import { animated, useSpring } from 'react-spring';
+// @ts-ignore
+import { interpolatePath } from 'd3-interpolate-path';
 
 export default function AnimatedPath({
   d,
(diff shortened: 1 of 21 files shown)


packages/visx-xychart/src/hooks/useScales.ts
```diff
@@ -32,15 +34,18 @@ export default function useScales<
...
     let xScale = createScale(xScaleConfig) as XScale;
     type XScaleInput = ScaleInput<XScale>;
+    const xScaleIsDiscrete = isDiscreteScale(xScaleConfig);
+    const valueIsDefined = xScaleIsDiscrete ? (v: XScaleInput) => v != null : isValidNumber;
 
     const xValues = registryEntries.reduce<XScaleInput[]>(
       (combined, entry) =>
-        entry ? combined.concat(entry.data.map((d: Datum) => entry.xAccessor(d))) : combined,
+        entry
+          ? combined.concat(entry.data.map((d: Datum) => entry.xAccessor(d)).filter(valueIsDefined))
```
williaster, line 43: verified that `d3Extent` handles this so no need to filter, will remove

packages/visx-xychart/src/components/series/private/AnimatedPath.tsx
```diff
@@ -1,12 +1,33 @@
...
+  // d3-interpolate-path is better at interpolating extra/fewer points so we use that
+  const interpolator = interpolatePath(previousD.current, d);
+  previousD.current = d;
+  // @ts-ignore t is not in CSSProperties
+  const { t } = useSpring({
+    from: { t: 0 },
+    to: { t: 1 },
+    reset: true,
+  });
+  const tweened = useSpring({ stroke, fill });
```
felixcodes, line 24: TIL `tweened` because 
>Inbetweening or tweening is a key process in all types of animation, including computer animation.

packages/visx-xychart/src/components/series/private/BaseBarGroup.tsx
```diff
@@ -132,8 +126,8 @@ export default function BaseBarGroup<
   const bars = registryEntries.flatMap(({ xAccessor, yAccessor, data, key }) => {
     const getLength = (d: Datum) =>
       horizontal
-        ? (xScale(xAccessor(d)) ?? 0) - xZeroPosition
-        : (yScale(yAccessor(d)) ?? 0) - yZeroPosition;
+        ? (xScale(xAccessor(d)) ?? NaN) - xZeroPosition
```
felixcodes, line 129: nice

hshoff (approved):

PR #902, opened 2020-10-29
new(drag): add `useDrag` hook

williaster:
#### :rocket: Enhancements

This PR 
- adds a `useDrag` hook to `@visx/drag` to complement the render props `<Drag />` component 
  - **(I need this for some `@visx/annotation` work)**
- refactors `<Drag />` to use `useDrag`
- bumps `react` `peerDependency` to `^16.8.0-0` to support hooks
- adds support for `PointerEvents` (in addition to `Mouse` and `Touch` events) in drag callbacks
- updates `@visx/demo` `/drag-ii` to use `useDrag` so we have examples of both usages
- updates `/docs/drag` to include `useDrag`
- adds minimal tests for all `@visx/drag` apis

Tested via
- verified `/drag-ii` is still functional
- Updated docs

@hshoff @kristw

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-drag-i/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-drag-i/Example.tsx
@@ -64,27 +64,25 @@ export default function DragI({ width, height }: DragIProps) {
               setDraggingItems(raise(draggingItems, i));
             }}
           >
-            {({ dragStart, dragEnd, dragMove, isDragging, dx, dy }) =>
-              (false && isDragging && console.log(d.id, d.x, dx)) || (
-                <circle
-                  key={`dot-${d.id}`}
-                  cx={d.x}
-                  cy={d.y}
-                  r={isDragging ? d.radius + 4 : d.radius}
-                  fill={isDragging ? 'url(#stroke)' : colorScale(d.id)}
-                  transform={`translate(${dx}, ${dy})`}
-                  fillOpacity={0.9}
-                  stroke={isDragging ? 'white' : 'transparent'}
-                  strokeWidth={2}
-                  onMouseMove={dragMove}
-                  onMouseUp={dragEnd}
-                  onMouseDown={dragStart}
-                  onTouchStart={dragStart}
-                  onTouchMove={dragMove}
-                  onTouchEnd={dragEnd}
-                />
-              )
-            }
+            {({ dragStart, dragEnd, dragMove, isDragging, dx, dy }) => (
+              <circle
+                key={`dot-${d.id}`}
+                cx={d.x}
+                cy={d.y}
+                r={isDragging ? d.radius + 4 : d.radius}
+                fill={isDragging ? 'url(#stroke)' : colorScale(d.id)}
+                transform={`translate(${dx}, ${dy})`}
+                fillOpacity={0.9}
+                stroke={isDragging ? 'white' : 'transparent'}
+                strokeWidth={2}
+                onMouseMove={dragMove}
+                onMouseUp={dragEnd}
+                onMouseDown={dragStart}
+                onTouchStart={dragStart}
+                onTouchMove={dragMove}
+                onTouchEnd={dragEnd}
+              />
+            )}
           </Drag>
         ))}
       </svg>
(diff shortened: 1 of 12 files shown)


packages/visx-demo/src/sandboxes/visx-drag-i/Example.tsx
```diff
@@ -64,27 +64,25 @@ export default function DragI({ width, height }: DragIProps) {
               setDraggingItems(raise(draggingItems, i));
             }}
           >
-            {({ dragStart, dragEnd, dragMove, isDragging, dx, dy }) =>
-              (false && isDragging && console.log(d.id, d.x, dx)) || (
```
williaster, line 68: just removed this guy 😱

packages/visx-demo/src/sandboxes/visx-drag-ii/Example.tsx
```diff
@@ -33,63 +58,45 @@ export default function DragII({ data = [], width, height }: DragIIProps) {
...
-              nextLines[lastIndex] = [...(nextLines[lastIndex] || []), newPoint];
-              return nextLines;
-            });
-          }}
-        >
-          {({ x = 0, y = 0, dx, dy, isDragging, dragStart, dragEnd, dragMove }) => (
+
+        <g>
+          {isDragging && (
+            /* capture mouse events (note: <Drag /> does this for you) */
```
williaster, line 64: 💬  this is the key missing feature from `useDrag`

packages/visx-drag/src/util/useStateWithCallback.ts
```diff
@@ -0,0 +1,33 @@
...
+  const setStateCallback = useCallback<SetStateWithCallback<State>>(
+    (nextState, callback) => {
+      callbackRef.current = callback || null;
+      setState(nextState);
+    },
+    [setState],
+  );
+
+  // if we use useEffect, some callback invocations are skipped
+  useLayoutEffect(() => {
```
williaster, line 24: the hook is the general approach I saw for trying to re-create the `setState(state, callback)` component api in hooks, except they all recommend `useEffect`.  However, when I used `useEffect` I noticed in the `/drag-ii` demo that some of the callbacks were skipped (e.g., the `useDragOptions.onDragStart` [the callback] would be skipped sometimes, which resulted in the previous line being connected to the new line you were starting to draw [`onDragStart` creates a new line in `/drag-ii`]).

I don't have a ton of experience using `useLayoutEffect` vs `useEffect`, I just know that `useLayoutEffect` is blocking while `useEffect` isn't as documented in the react api and we get a warning in `next` now that `useLayoutEffect` cannot be used server-side (for this functionality I don't think it matters).  

Would love any thoughts on this!

packages/visx-drag/src/useDrag.tsx
```diff
@@ -0,0 +1,131 @@
...
+  dragStart: (event: MouseOrTouchEvent) => void;
+};
+
+/** Hook for dragging, returns a `UseDrag` object. */
+export default function useDrag({
+  resetOnStart = false,
+  onDragEnd,
+  onDragMove,
+  onDragStart,
+}: UseDragOptions): UseDrag {
```
williaster, line 51: note to make this optional to enable `useDrag()` if desired

packages/visx-drag/src/useDrag.tsx
```diff
@@ -0,0 +1,131 @@
+import React, { useCallback } from 'react';
+import { localPoint } from '@visx/event';
+import useStateWithCallback from './util/useStateWithCallback';
+
+type MouseOrTouchEvent = React.MouseEvent | React.TouchEvent;
```
hshoff, line 5: Could add `React.PointerEvent`. Was added in React 16.4 https://reactjs.org/blog/2018/05/23/react-v-16-4.html#pointer-events

packages/visx-drag/src/util/useStateWithCallback.ts
hshoff: Hm. Not ideal but this could be something we revisit in the future.

hshoff (approved):

hshoff (changes requested):
Just remembered, need to bump react peerDep to `^16.8.0-0`. similar to [@visx/tooltip](https://github.com/airbnb/visx/blob/master/packages/visx-tooltip/package.json#L39).  (hooks were added in 16.8 https://reactjs.org/docs/hooks-intro.html). 
https://github.com/airbnb/visx/blob/master/packages/visx-tooltip/package.json#L39

hshoff (approved):

Issue #905, opened 2020-10-30
Drag: programmatically reset dragging position


m0t0r:
Hello, 

I would like to ask if it would be possible to programmatically reset dragging position? I was checking the current `<Drag />` component and unable to find how to reset the position (there is only option to reset on new drag start but I would like to control it without new drag). I also saw a new #902 PR for the hook and wondering if it would be possible to add such API.

Thanks

williaster:
Hey @m0t0r 👋  this is a great suggestion 💯  I was actually thinking about the need for this yesterday when working on the `useDrag` addition you mentioned 😄 

Let me noodle on how best to add support and will try and get it into #902 or a follow up 👍

m0t0r:
Hey @williaster, great! Thanks a lot! 👍

PR #907, opened 2020-10-30
new(annotation): add new Annotation components

williaster:
#### :rocket: Enhancements

This PR builds on required `@visx/drag` work (#902, #905) and adds several new components to `@visx/annotation`.

**Why?**

Previously we recommended using [`react-annotation`](https://react-annotation.susielu.com/) which is the most fully featured annotation library out there. We would like to build in a solid annotation layer into the new `@visx/xychart`, and when evaluating `react-annotation` I was a little discouraged because its bundle size is relatively high (`95kb` minified) and it has no typings so I would have had to re-write the types anyway.

I think we can keep this package size down by avoiding use of `d3-shape` which adds `30kb`, and doing some relatively simple math for the few shapes we need.

**What**

Example usage

```tsx
<Annotation x={subjectX} y={subjectY} dx={labelDx} dy={labelDy}>
  <CircleSubject radius={...} />
  <Connector stroke={...} />
  <Label title={...} subtitle={...} width={...} />
</Annotation>
```

Several components are added which are inspired by the `react-annotation` API which has a concept of subject, connector, and label:
- `Label` the annotation label, takes `title/subtitle` and handles `@visx/text` rendering, styles, and wrap
- `Connector` the connector from an annotation `Label` to the annotation `*Subject`
  - has `line` and `elbow` types (latter shown above)
- Subjects
  - `CircleSubject` to annotate a single point (shown above)
  - `LineSubject` to annotate a threshold (vertical or horizontal line)

In addition to these component parts there are two higher-level components
- `Annotation` simply takes `x/y/dx/dy` as props and provides them in `context` for consumption by child subject, connector, and label components
- `EditableAnnotation` wraps `Annotation` and allows draggable editing of an `Annotation`s position (subject + label)

Future work
- add `/annotation` demo to `@visx/demo` (sneak peek above 👀 ) + update `/docs/annotation` page
- future functionality
  - add `RectSubject` to annotate regions
  - add ability to resize `CircleSubject`
  - polish `Connector` to account for a `Subject`s fill region (above you can see that the connector line spills into the circle subject's radius, ideally it is aware of the subject region and doesn't enter it)

@kristw @hshoff


williaster:
I'll add better tests in a separate PR since coverage ⬇️

hshoff (approved):

kristw (approved):

PR #913, opened 2020-11-04
new(xychart): make DataProvider optional

williaster:
#### :rocket: Enhancements

This PR adds a usability improvement to `@visx/xychart`, making it optional for package consumers to render `DataProvider` which is arguably an internal implementation detail (note that it is already optional for consumers to render `TooltipProvider` and `EventEmitterProvider`). 

Users can still render `DataProvider` up the parent chain for advanced use cases like linked charts, where it's desirable to share a single `DataContext`.

```tsx
// before (still valid)
() => (
  <DataProvider xScale={..} yScale={..} theme={..}>
    <XYChart width={..} height={..}>
      {...}
    </XYChart>
  </DataProvider>
)

// after
() => (
  <XYChart 
    xScale={..} 
    yScale={..} 
    theme={..}
    width={..} 
    height={..}
  >
    {...}
  </XYChart>
)
```

**Testing**
- [x] new jest tests
- [x] verified that `/xychart` demo is still functional

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
@@ -9,7 +9,6 @@ import {
   AnimatedGlyphSeries,
   AnimatedGrid,
   AnimatedLineSeries,
-  DataProvider,
   Tooltip,
   XYChart,
 } from '@visx/xychart';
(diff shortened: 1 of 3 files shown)


packages/visx-demo/src/sandboxes/visx-xychart/Example.tsx
```diff
@@ -53,173 +52,171 @@ export default function Example({ height }: Props) {
...
-                  dataKey="Austin"
-                  data={data}
-                  xAccessor={accessors.x.Austin}
-                  yAccessor={accessors.y.Austin}
-                />
-              </AnimatedBarGroup>
-            )}
-            {renderBarSeries && (
+        <XYChart theme={theme} xScale={config.x} yScale={config.y} height={Math.min(400, height)}>
+          <CustomChartBackground />
```
williaster, line 56: 💬  only `DataProvider` was removed, everything else in this file is a de-indentation

packages/visx-xychart/test/components/XYChart.test.tsx
```diff
@@ -9,40 +9,66 @@ import {
...
+  it('should warn if DataProvider is not available and no x- or yScale config is passed', () => {
+    expect(
+      () =>
+        mount(
+          <XYChart>
+            <rect />
+          </XYChart>,
+        ),
+      // eslint-disable-next-line jest/require-to-throw-message
+    ).toThrow();
```
williaster, line 53: this throws only because nimbus transforms `console.warn`s to throw

kristw (approved):

PR #925, opened 2020-11-10
Remove resize observer polyfill

koddsson:
- Added new "ParentSize" components that don't include a polyfill for `ResizeObserver`.

---

My application doesn't require the polyfill for resize observer since my browser support policy only includes browsers that support `ResizeObserver`.

In keeping with this projects mantra of, "Keep your bundle sizes down and use only the packages you need.", I'm creating new components that don't include the polyfill. Maybe the old components can be removed in future major versions.


hshoff:
Hi @koddsson, I hear you on this. This has been a trade-off discussion for a couple years now. Here's the previous discussion: https://github.com/airbnb/visx/pull/254#issuecomment-415182999

This is going to continue coming up in the future as we get further away from legacy browser support and modern browser support for ResizeObserver widens.

Unfortunately, this is something we would have to land in a `v2` because of the breaking change. But we would like to avoid breaking changes as much as possible. 

One thing that is mentioned in the previous discussion is the idea of alias-ing out the polyfill dependency so it's not including in bundles. This isn't ideal though as it requires more work to do the paved road approach of not needing a polyfill.

My preference would be to avoid a v2 and breaking change release and to add new components to `@visx/responsive` that don't include the polyfill (and/or make it opt-in). Would be happy to review a PR that makes these additions.

koddsson:
> My preference would be to avoid a v2 and breaking change release and to add new components to `@visx/responsive` that don't include the polyfill (and/or make it opt-in). @williaster and I can prioritize landing these new polyfill-less components in `v1.2.0`. Would also be happy to review a PR that makes these additions.

Would this just be copies of the original? So in this case I'd just clone the `ParentSize.tsx` component, remove the polyfill and call it something like `ParentSizeWithoutPolyfills.tsx`?

williaster:
@hshoff do you have thoughts on the name of the component? maybe `ParentSizeOptInPolyfill` for now and the next major release we'd swap it to `ParentSize`.

Regarding implementation, I think [`react-use-measure`](https://github.com/pmndrs/react-use-measure#resize-observer-polyfills) has a great opt-in design. I think it'd translate to something like:

```tsx
// user passes in their own polyfill
import { ResizeObserver } from 'my-favorite-polyfill';
() => <ParentSize polyfill={ResizeObserver}>{...}</ParentSize>

// user polutes the `window` object (current implementation)
import ResizeObserver from 'resize-observer-polyfill';
() => <ParentSize >{...}</ParentSize>

// browser policy supports ResizeObserver already
() => <ParentSize >{...}</ParentSize>

// user does nothing, and browser doesn't support
() => <ParentSize >{...}</ParentSize>
// Error: This browser does not support ResizeObserver out of the box
// see example implementation https://github.com/pmndrs/react-use-measure/blob/master/src/web/index.ts#L47
```

Make sense?

hshoff:
`ParentSizeOptInPolyfill` or `ParentSizeModern` sound good to me

koddsson:
I've pushed some changes and update the PR description. I opted for not accepting a polyfill as a prop since I feel like polyfills should just patch the global object but people might disagree.

Let me know what y'all think and I can make changes accordingly.

hshoff (approved):
Looks good to me! thanks for this @koddsson 

We can merge this and make the linter happy in a follow up PR. Should land in 1.3.0

williaster:
@koddsson thanks for the changes 🙏  do you want to fix the lint issues or do you mind if I push them to this branch? (trying not to merge something that fails CI)

koddsson:
> @koddsson thanks for the changes 🙏 do you want to fix the lint issues or do you mind if I push them to this branch? (trying not to merge something that fails CI)

Feel free to push to the branch, I can also fix them tomorrow if you'd like.

williaster:
Thanks @koddsson! Will try to get this out today as `1.3.0`

williaster:
heads up this is out as a `1.3.0-alpha.0` pre-release if you really need it. trying to merge a couple more open PRs for `1.3.0`, tomorrow should be possible for that.

PR #926, opened 2020-11-11
add textPath prop to Text component

grsoares21:
This was implemented as a solution for https://github.com/airbnb/visx/issues/759
It is my first PR to a big open-source project. I'd appreciate your feeback.

#### :boom: Breaking Changes

- none

#### :rocket: Enhancements

- Adds an optional textPath props that can be used as a path for the text to follow along

### DIFF
--- a/packages/visx-text/src/Text.tsx
+++ b/packages/visx-text/src/Text.tsx
@@ -1,20 +1,20 @@
-import React from 'react';
-import reduceCSSCalc from 'reduce-css-calc';
-import getStringWidth from './util/getStringWidth';
-import uniqueId from 'lodash/uniqueId';
+import React from "react";
+import reduceCSSCalc from "reduce-css-calc";
+import getStringWidth from "./util/getStringWidth";
+import uniqueId from "lodash/uniqueId";
 
-const SVG_STYLE = { overflow: 'visible' };
+const SVG_STYLE = { overflow: "visible" };
 
 function isNumber(val: unknown): val is number {
-  return typeof val === 'number';
+  return typeof val === "number";
 }
 
 function isValidXOrY(xOrY: string | number | undefined) {
   return (
     // number that is not NaN or Infinity
-    (typeof xOrY === 'number' && Number.isFinite(xOrY)) ||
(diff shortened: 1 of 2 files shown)


packages/visx-text/src/Text.tsx
```diff
@@ -111,6 +117,12 @@ class Text extends React.Component<TextProps, TextState> {
     this.updateWordsByLines(this.props, needCalculate);
   }
 
+  updateTextPathId(props: TextProps) {
+    if(props.textPath) {
+      this.setState({ textPathId: uniqueId('text-path-') })
+    }
```
hshoff, line 123: I see this is called from `componentDidMount()`. Ideally we could move this to the `constructor()` because it doesn't rely on the DOM.

hshoff (changes requested):
Hi @grsoares21, thanks for the PR! This looks good to me except for a minor change.

packages/visx-text/src/Text.tsx
grsoares21: Sorry, it took me so long to fix this. I've moved the whole state initialization into the constructor with the id generation in there as well.

susiwen8:
@grsoares21 I'm very sorry, `Text` has been refactored by `hook`(#946 ) that means you should rewrite this feature...... I‘m sincerely  sorry about it........

grsoares21:
> @grsoares21 I'm very sorry, `Text` has been refactored by `hook`(#946 ) that means you should rewrite this feature...... I‘m sincerely sorry about it........

Hello @susiwen8 . No problems :D. I'm glad to hear it's on hooks now. I'll try to reimplement this next weekend then.

williaster:
Sorry for the delayed review, will try and review asap once I see any new commits to avoid any future conflicts 👍

derrickbeining:
Came here for this very thing. Sad to see it has stalled. Are you planning to pick it up again @grsoares21 ?

williaster:
If someone else wants to pick this up I'm happy to review! it shouldn't be too complicated, tho as noted `@visx/text` was refactored to use hooks so the approach will need to be tweaked slightly for that.

PR #931, opened 2020-11-17
add tokenizedbtc to in the wild list

sakulstra:
#### :memo: Documentation
Over the last few days I created this little demo which uses a lot of visx features in combination (stack, tooltip, brush + some interactivity)
image


hshoff (approved):
Awesome, nice work!

PR #934, opened 2020-11-18
new(brush): expose updateBrush method

peterwiebe:
#### :rocket: Enhancements

ezgif com-video-to-gif

@williaster - I know you already have a PR up but I was wondering what you think of this.
cc #577

### DIFF
--- a/packages/visx-brush/src/Brush.tsx
+++ b/packages/visx-brush/src/Brush.tsx
@@ -56,6 +56,7 @@ export type BrushProps = {
   resetOnEnd?: boolean;
   /** Size of Brush handles, applies to all `resizeTriggerAreas`. */
   handleSize: number;
+  innerRef?: React.Object<BaseBrush>;
 };
 
 class Brush extends Component<BrushProps> {
@@ -160,6 +161,7 @@ class Brush extends Component<BrushProps> {
       margin,
       brushDirection,
       initialBrushPosition,
+      innerRef,
       resizeTriggerAreas,
       brushRegion,
       yAxisOrientation,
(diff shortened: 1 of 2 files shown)


packages/visx-brush/src/Brush.tsx
```diff
@@ -216,20 +219,21 @@ class Brush extends Component<BrushProps> {
...
+        brushDirection={brushDirection}
+        disableDraggingSelection={disableDraggingSelection}
+        handleSize={handleSize}
         inheritedMargin={margin}
         initialBrushPosition={initialBrushPosition}
-        onChange={this.handleChange}
-        onBrushEnd={this.handleBrushEnd}
-        onBrushStart={this.handleBrushStart}
-        handleSize={handleSize}
+        ref={innerRef}
```
peterwiebe, line 227: this is the added line, the rest are just sorted by proptype & name

peterwiebe:
@williaster if there isn't any issues with this implementation can this be added to the next release?

packages/visx-brush/src/Brush.tsx
```diff
@@ -56,6 +56,8 @@ export type BrushProps = {
   resetOnEnd?: boolean;
   /** Size of Brush handles, applies to all `resizeTriggerAreas`. */
   handleSize: number;
+  /** Reference to the BaseBrush component */
```
williaster, line 59: nit – can we add `.` here for consistency with the other prop annotations?

packages/visx-demo/src/sandboxes/visx-brush/Example.tsx
```diff
@@ -113,6 +115,38 @@ function BrushChart({
...
+      setFilteredStock(stock);
+      brushRef.current.reset();
+    } else {
+      console.log(`innerRef Prop not properly set.`);
+    }
+  };
+
+  const handleLastWeekClick = () => {
+    if (brushRef?.current === null) {
+      console.log(`innerRef Prop not properly set.`);
```
williaster, line 129: the console logs are useful in debugging, but for the simplicity of the demo (which should work!) I wonder if we could remove them?

packages/visx-demo/src/sandboxes/visx-brush/Example.tsx
```diff
@@ -164,6 +199,8 @@ function BrushChart({
           />
         </AreaChart>
       </svg>
+      <button onClick={handleResetClick}>Reset</button>&nbsp;
```
williaster, line 202: thanks for the demo addition! wondering if we could make two minor changes:
- Instead of adding a `1 week` position which differs from `initialBrushPosition`, could we re-use `initialBrushPosition`? In this case we could re-name the button label to `Reset` (since it's the initial position, see next point)
- If we make the above^ change, `Reset` seems to behave more like `Clear` (even though the `Brush` method name is `.reset()`) , could we update to the button label + `handleResetClick` variable name to use `clear`?

williaster (commented):
Thanks for the PR and sorry for the delayed review @peterwiebe (catching up from being out 🌴  last week). This looks great and is much simpler than my previous thinking 💯 

I had a couple of minor suggestions for the demo, but then yes we should be able to get this out in the next release 👍

peterwiebe:
@williaster thank you for your review 🙏  I updated the PR with your feedback, please let me know if there is anything else you think could use some tweaking.
ezgif com-video-to-gif

williaster (approved):
This looks great @peterwiebe! 💯  Thanks for the addition, I'll merge it now and try and get a release out sometime today.

williaster:
@peterwiebe sorry for the delay, this is now out as `1.2.0`. demo site will be live tomorrow but [sandbox](https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-brush?file=/Example.tsx) works.

peterwiebe:
@williaster thank you for helping me get it in 🙏

PR #937, opened 2020-11-18
fix(shape): conditionally render Arc without data

### LINKED ISSUE #886: Arc example

williaster:
#### :bug: Bug Fix

This supersedes #892 and partially addresses #886, it
- conditionally renders `@visx/shape`s `Arc` when no `data` is specified _**if angles + radii are specified**_
- adds a warning if we render `null` due to missing `data` + angles + radii for better dev experience

As mentioned in both issues above, d3-shape supports two types of usage for `arc()`, it can be configured with constant angles + radii, or passed them (as `data` in `@visx/shape`):

```ts
const arc1 = d3.arc();
arc1({ innerRadius, ... }); // "M0,-100A100,100,0,0,1,100,0L0,0Z"

const arc2 = d3.arc().innerRadius(...)...;
arc2(); // "M0,-100A100,100,0,0,1,100,0L0,0Z"
```

@hshoff @kristw 
cc @oscar-broman @thecodeah

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12116,7 +12116,7 @@ reserved-words@^0.1.2:
   resolved "https://registry.yarnpkg.com/reserved-words/-/reserved-words-0.1.2.tgz#00a0940f98cd501aeaaac316411d9adc52b31ab1"
   integrity sha1-AKCUD5jNUBrqqsMWQR2a3FKzGrE=
 
-resize-observer-polyfill@1.5.1:
+resize-observer-polyfill@1.5.1, resize-observer-polyfill@^1.5.1:
   version "1.5.1"
   resolved "https://registry.yarnpkg.com/resize-observer-polyfill/-/resize-observer-polyfill-1.5.1.tgz#0e9020dd3d21024458d4ebd27e23e40269810464"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -12116,7 +12116,7 @@ reserved-words@^0.1.2:
   resolved "https://registry.yarnpkg.com/reserved-words/-/reserved-words-0.1.2.tgz#00a0940f98cd501aeaaac316411d9adc52b31ab1"
   integrity sha1-AKCUD5jNUBrqqsMWQR2a3FKzGrE=
 
-resize-observer-polyfill@1.5.1:
+resize-observer-polyfill@1.5.1, resize-observer-polyfill@^1.5.1:
```
williaster, line 12119: got a fresh laptop and this was changed

kristw (approved):

PR #946, opened 2020-11-25
Refactor(@visx/text): rewrite Text with hook

susiwen8:
#### :rocket: Enhancements

add `@visx/text` `useText` hook

#### :house: Internal

rewrite `@visx/text`  `Text` with `useText` hook


susiwen8:
@williaster Hi, I have rewrite Text by hook, please review it. Thanks

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export function useText(props: TextProps): [WordsWithWidth[], string, string] {
+  const [wordsByLines, setWordsByLines] = useState<WordsWithWidth[]>([]);
+  const wordsWithWidth = useRef<WordWithWidth[]>([]);
+  const spaceWidth = useRef(0);
+  const prevProps = usePreviousValue(props);
+  const prevWordsByLines = usePreviousValue(wordsByLines);
+
+  const calculateWordsByLines = useCallback(
```
susiwen8, line 18: Are those `useCallback` useless? Cause they all depend on `props`, which will not be reference equality at each `render`.

susiwen8:
@williaster Hi Chris, Did you got chance to review my code? :)

williaster:
Sorry for the delay @susiwen8 , was gone on holiday last week and should be able to review today 👀 👍

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,122 @@
...
+    const originY = y - sy * y;
+    transforms.push(`matrix(${sx}, 0, 0, ${sy}, ${originX}, ${originY})`);
+  }
+  if (angle) {
+    transforms.push(`rotate(${angle}, ${x}, ${y})`);
+  }
+
+  const transform = transforms.length > 0 ? transforms.join(' ') : '';
+
+  return [wordsByLines, startDy, transform];
```
williaster, line 121: since this hook has >2 return values, I'm wondering if we should return an object with named values instead. This shouldn't affect memoization of the values at all.

packages/visx-text/src/Text.tsx
```diff
@@ -17,216 +15,39 @@ function isValidXOrY(xOrY: string | number | undefined) {
...
+export default function Text(props: TextProps) {
+  const {
+    dx = 0,
+    dy = 0,
+    textAnchor = 'start',
+    innerRef,
+    verticalAnchor,
+    angle,
+    lineHeight = '1em',
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
williaster, line 27: this shouldn't be needed (typically if there is a `...rest`, the linter is smart enough not to flag)

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,122 @@
...
+      prevProps?.children !== props.children || prevProps?.style !== props.style;
+    updateWordsByLines(needCalculate);
+  });
+
+  const {
+    verticalAnchor = 'end',
+    scaleToFit = false,
+    angle,
+    lineHeight = '1em',
+    capHeight = '0.71em',
```
williaster, line 90: can we keep the previous comment for this value `// Magic number from d3`

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,122 @@
...
+      }, []);
+    },
+    [props],
+  );
+
+  const updateWordsWithoutCalculate = useCallback(() => {
+    const words =
+      props.children == null ? [] : props.children.toString().split(/(?:(?!\u00A0+)\s+)/);
+    setWordsByLines([{ words }]);
+  }, [props]);
```
williaster, line 50: should be able to update this dependency to just `[children]`

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        setWordsByLines(calculateWordsByLines(width));
+      } else {
+        updateWordsWithoutCalculate();
+      }
+    },
+    [calculateWordsByLines, props, updateWordsWithoutCalculate],
+  );
+
+  useEffect(() => {
```
williaster, line 75: I think previously this check was needed because that's how it needed to be done inside a component. Hooks implement this dependency check for us, so I'm wondering if we can refactor this to compute `wordsByLines` behind a `useMemo` hook, with a dependency on `[children, style]` because that's when it should be updated. 

That would make this code more readable/easy to follow now that we no longer need this imperative logic.

williaster (commented):
Thanks for the PR @susiwen8 ! It's looking great overall, I left a first round of comments and can take another look after that! 👀

packages/visx-text/src/Text.tsx
susiwen8: The reason why I put `scaleToFit` out of `textProps` is that `sacle-to-fit` isn't a legal SVG attribute
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

Otherwise `scaleToFit` wrapped in `textProps`, browser will throw an error.

packages/visx-text/src/hooks/useText.ts
susiwen8: @williaster Hi Chris, please take a look, your idea is great. :+1:

susiwen8:
@williaster Hi, any thoughts?

packages/visx-annotation/test/Label.test.tsx
```diff
@@ -9,27 +9,25 @@ describe('<Label />', () => {
...
   it('should render title Text', () => {
-    expect(
-      shallow(<Label title="title test" resizeObserverPolyfill={ResizeObserver} />)
-        .find(Text)
-        .dive()
-        .text(),
-    ).toBe('title test');
+    shallow(<Label title="title test" resizeObserverPolyfill={ResizeObserver} />)
+      .find('visx-annotationlabel')
+      .forEach(item => {
```
williaster, line 14: can you explain why this needed to be changed? in theory the API of `Text` shouldn't change with these changes right?

a `.forEach` is dangerous because if there are no matches for `visx-annotationlabel`, there would be no assertions and the test would "pass". so if this was needed we should add `expect.hasAssertions()` to the test.

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,123 @@
...
+    width,
+    lineHeight = '1em',
+    capHeight = '0.71em', // Magic number from d3
+    children,
+    style,
+    ...textProps
+  } = props;
+
+  const { x = 0, y = 0 } = textProps;
+  const isXOrYValid = !isValidXOrY(x) || !isValidXOrY(y);
```
williaster, line 39: naming nit – I think this actual should be `isXOrYInvalid` (invalid, not valid)

```suggestion
  const isXOrYInValid = !isValidXOrY(x) || !isValidXOrY(y);
```

packages/visx-text/src/Text.tsx
```diff
@@ -1,232 +1,41 @@
 import React from 'react';
-import reduceCSSCalc from 'reduce-css-calc';
-import getStringWidth from './util/getStringWidth';
+import useText from './hooks/useText';
+import { TextProps } from './types';
+
+export { TextProps } from './types';
```
williaster, line 5: 👍

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,123 @@
...
+    return [
+      {
+        words: children == null ? [] : children.toString().split(/(?:(?!\u00A0+)\s+)/),
+      },
+    ];
+  }, [isXOrYValid, width, scaleToFit, children, wordsWithWidth, spaceWidth]);
+
+  const startDy = useMemo(() => {
+    const startDyStr = isXOrYValid
+      ? ''
```
williaster, line 89: it looks like previously this case would result in `startyDy` being `undefined`, empty string is probably okay tho

packages/visx-text/src/hooks/useText.ts
```diff
@@ -0,0 +1,123 @@
...
+        )
+      : reduceCSSCalc(`calc(${wordsByLines.length - 1} * -${lineHeight})`);
+
+    return startDyStr;
+  }, [isXOrYValid, verticalAnchor, capHeight, wordsByLines.length, lineHeight]);
+
+  const transform = useMemo(() => {
+    const transforms: string[] = [];
+    if (isXOrYValid) {
+      return '';
```
williaster, line 104: same with this being `undefined` before, again probably fine

williaster (commented):
Hey @susiwen this is looking GREAT 😎  Thanks for simplifying the logic per my previous comments 👌 . 

I had one more small comment for a variable name, I'm also going to pull the branch to verify it functionally locally but then we should be able to get it in!

williaster:
Functionally looks good to me 👍  🎉  

Performance-wise seems similar or even slightly better (we should have CI checks for this ideally :)

packages/visx-text/src/Text.tsx
susiwen8: Actually It can't :joy: :joy: :joy:
Screen Shot 2020-12-04 at 09 23 09
Screen Shot 2020-12-04 at 09 24 15

packages/visx-annotation/test/Label.test.tsx
susiwen8: That was little embarrassing, cause I'm not familiar with `enzyme`.... I have update `label` test case inspired by `axis`
https://github.com/airbnb/visx/blob/master/packages/visx-axis/test/Axis.test.tsx#L177

williaster: no worries at all! 😅

packages/visx-text/src/Text.tsx
williaster: oh no! 🙉  sorry for the back and forth.

williaster (approved):
Thanks again @susiwen8 this is awesome 💯  

This should land in `1.3.0` which we can get out early next week unless you need it for something sooner 👍

susiwen8:
@williaster Thanks for guiding. I really learn something!😁

PR #947, opened 2020-11-25
new(xychart): add PointerEvent handlers to XYChart and *Series

williaster:
**TODO**
- [x] add tests in separate PR #952 

#### :rocket: Enhancements

This PR 
- adds the ability to set `PointerEvent` handlers on `XYChart`, or on individual `*Series` (previously the only mouse events which were handled were to show/hide tooltips)
  - `onPointerMove/Out/Up` are currently supported
    - hoping this will be compatible with focus/blur (a11y), and brush/zoom/pan in the future
- adds `pointerEvents: boolean` to all `*Series`, which enables users to prevent a `Series` from emitting, or responding to any `PointerEvent`s (including tooltips; see demo below)
- all `mousemove/out` + `touchmove/end` references were simplified to `pointermove/out`
- updates the `/xychart` example to demonstrate `onPointerUp` => if you click on the chart, the `Annotation` will move to the nearest `Datum` in the nearest `Series`

**Usage examples**

```tsx
// set handler on chart
// invoke for nearest Datum in _nearest_ `Series` to event, or nearest Datum in _all_ `Series`
<XYChart onPointerUp={() => ...} pointerEvents={'all' | 'nearest'}>
  <LineSeries {...} />
  <LineSeries {...} />
</XYChart>

// invoke handler for only a specific `Series`
// `XYChart` will capture and emit events when `captureEvents=true`, else Series DOM elements emit events
// note: Tooltips still work for Series regardless of `PointerEvent` handlers, unless pointerEvents={false}
<XYChart captureEvents={true | false}>
  <LineSeries onPointerMove={() => ...}  {...} />
  <LineSeries {...} />
</XYChart>
```

**Implementation notes**
- all events are dispatched through the `EventEmitterContext` rather than being handled within a given component. 
  - In order to scope e.g., `BarSeries` events to _that_ `BarSeries`, I added the notion of `EventEmitter` `source`s so that a `Series` can subscribe to only it's own events + events emitted by `XYChart`
- `Series` `PointerEvent` handlers are set on _individual_ bars, glyphs, or paths (rather than a container to capture events on all of them) to prevent one `Series` blocking events for another series.

**Demo of `pointerEvents={false}` for all Series except SF** (therefore tooltip only contains SF, even though mouse is closer to other `AreaSeries`)

@kristw @hshoff


packages/visx-xychart/src/components/XYChart.tsx
```diff
@@ -36,18 +43,52 @@ export type XYChartProps<
...
+    datum,
+    distanceX,
+    distanceY,
+    event,
+    index,
+    key,
+    svgPoint,
+  }: PointerEventParams<Datum>) => void;
+  /** Whether to invoke PointerEvent handlers for all dataKeys, or the nearest dataKey. */
+  pointerEvents?: 'all' | 'nearest';
```
williaster, line 72: maybe should be called `pointerEventDataKey` 🤔

packages/visx-xychart/src/components/XYChart.tsx
```diff
@@ -118,19 +152,28 @@ export default function XYChart<
...
   }
 
+  const pointerEventEmitters = usePointerEventEmitters({ source: XYCHART_EVENT_SOURCE });
+  const pointerEventHandlers = usePointerEventHandlers({
+    dataKey: pointerEvents === 'nearest' ? POINTER_EVENTS_NEAREST : POINTER_EVENTS_ALL,
+    onPointerMove,
+    onPointerOut,
+    onPointerUp,
+  });
+  useEventEmitter('pointermove', pointerEventHandlers.onPointerMove, eventSourceSubscriptions);
```
williaster, line 162: `useEventEmitter()` x3 calls are duplicated in all Series, I could push this into `usePointerEventHandlers`

packages/visx-xychart/src/components/series/private/BaseAreaSeries.tsx
```diff
@@ -26,23 +27,31 @@ export type BaseAreaSeriesProps<
   lineProps?: Omit<LinePathProps<Datum>, 'data' | 'x' | 'y' | 'children' | 'defined'>;
   /** Rendered component which is passed path props by BaseAreaSeries after processing. */
   PathComponent?: React.FC<Omit<React.SVGProps<SVGPathElement>, 'ref'>> | 'path';
-} & Omit<React.SVGProps<SVGPathElement>, 'x' | 'y' | 'x0' | 'x1' | 'y0' | 'y1' | 'ref'>;
+} & Omit<
+    React.SVGProps<SVGPathElement>,
+    'x' | 'y' | 'x0' | 'x1' | 'y0' | 'y1' | 'ref' | 'pointerEvents'
+  >;
+
+const eventSourceSubscriptions = [AREASERIES_EVENT_SOURCE, XYCHART_EVENT_SOURCE];
```
williaster, line 35: I think actually `eventSourceSubscriptions` must be equal to `${SERIESTYPE_EVENT_SOURCE}-${dataKey}`, otherwise Series could trigger events for other Series of the same type.

packages/visx-xychart/src/hooks/useEventEmitter.ts
```diff
@@ -2,34 +2,52 @@ import { useCallback, useContext, useEffect } from 'react';
...
+  eventType?: EventType,
+  /** Handler invoked on emission of EventType event.  */
+  handler?: Handler,
+  /** Optional valid sources for EventType subscription. */
+  sources?: string[],
+) {
   const emitter = useContext(EventEmitterContext);
 
-  /** wrap emitter.emit so we can enforce stricter type signature */
+  // wrap emitter.emit so we can enforce stricter type signature and filter events by source
```
williaster, line 31: todo: move bit about source to handler, not emitter

packages/visx-xychart/src/hooks/usePointerEventHandlers.ts
```diff
@@ -0,0 +1,117 @@
...
+import { PointerEventParams } from '../types';
+import findNearestDatumX from '../utils/findNearestDatumX';
+import findNearestDatumY from '../utils/findNearestDatumY';
+import { HandlerParams } from './useEventEmitter';
+
+export const POINTER_EVENTS_ALL = '__all';
+export const POINTER_EVENTS_NEAREST = '__nearest';
+
+type PointerEventHandlerParams<Datum extends object> = {
+  /** Controls whether callbacks are invoked for a given registry dataKey, the nearest dataKey, or all datakeys. */
```
williaster, line 12: `datakeys` => `dataKeys`

packages/visx-xychart/src/hooks/usePointerEventHandlers.ts
```diff
@@ -0,0 +1,117 @@
...
+      // nearest Datum across all dataKeys, if relevant
+      let nearestDatumPointerParams: PointerEventParams<Datum> | null = null;
+      let nearestDatumDistance = Infinity;
+
+      if (params && event && svgPoint && width && height && xScale && yScale) {
+        const considerAllKeys =
+          dataKey === POINTER_EVENTS_NEAREST || dataKey === POINTER_EVENTS_ALL;
+
+        // find nearestDatum for relevant dataKey(s)
+        (considerAllKeys
```
williaster, line 49: could pull this out to `dataKeys`

packages/visx-xychart/src/hooks/usePointerEventHandlers.ts
```diff
@@ -0,0 +1,117 @@
...
+              width,
+              xAccessor: entry.xAccessor,
+              xScale,
+              yAccessor: entry.yAccessor,
+              yScale,
+            });
+            if (nearestDatum) {
+              pointerParamsByKey[key] = { key, svgPoint, event, ...nearestDatum };
+
+              // compute nearest Datum if not emitting events for all keys\
```
williaster, line 70: remove `\`

packages/visx-xychart/src/hooks/usePointerEventHandlers.ts
```diff
@@ -0,0 +1,117 @@
...
+          }
+        });
+      }
+    },
+    [dataKey, xScale, yScale, width, height, horizontal, onPointerMove, onPointerOut, onPointerUp],
+  );
+
+  const handlePointerOut = useCallback(
+    (params?: HandlerParams) => {
+      if (params && onPointerOut) onPointerOut(params.event);
```
williaster, line 107: because we have to always use the same # of hooks, we can't conditionally create this and have to do the check in the handler. we conditionally return the handler below, tho.

packages/visx-xychart/src/components/XYChart.tsx
williaster: I guess this probably should mirror the implementation and could also support specific `dataKey | dataKey[]` as well

kristw (approved):

packages/visx-xychart/src/components/series/private/BaseBarStack.tsx
```diff
@@ -113,58 +134,36 @@ function BaseBarStack<
...
-    (params: HandlerParams | undefined) => {
-      const { svgPoint } = params || {};
-
-      // invoke showTooltip for each key so all data is available in context,
-      // and let Tooltip find the nearest point among them
-      dataKeys.forEach(key => {
-        const entry = dataRegistry.get(key);
-        const childData = barSeriesChildren.find(child => child.props.dataKey === key)?.props.data;
-        if (childData && svgPoint && width && height && showTooltip) {
-          const datum = findNearestStackDatum(
```
williaster, line 129: noting as a TODO that this logic was lost in this refactor

Issue #832, opened 2020-10-01
[feature request]: radial bar chart


alangrn:
Is there any plan to add circular barplot to visx?
https://www.d3-graph-gallery.com/circular_barplot

williaster:
This isn't on our short list of features (I'm primarily focused on the `@visx/xychart` package) but we'd happily review the feature! `@visx/shape` has [`LineRadial`](https://airbnb.io/visx/lineradial) currently so this would fit well there.

sarathps93:
Hi @williaster is this feature request up for grabs? If yes shall I work on this?

williaster:
Hey @sarathps93 👋  it **is** open for grabs if you're interested! 🙏

sarathps93:
@williaster Thanks for the confirmation. I will start working on this

sarathps93:
Hi @williaster 
After some initial analysis need some input from you.

The given `circular bar plot` can be easily constructed using the existing `Arc` shape from `visx-shapes`. So should I create the `circular bar plot` as a separate component or just include them in the examples / gallery to let the users know how to construct it from the existing packages?
One more addition would be the inclusion of `scaleRadial` from `d3-scales`

williaster:
Hey @sarathps93 thanks for digging into this more! 🙏  Seems like adding `scaleRadial` to `@visx/scale` is clear, my sense for just adding a demo using `Arc` vs a higher-level `RadialBarChart` is to start with the demo (this would be consistent with the approach for a cartesian Bar Chart).

Thinking about the primitives that might be needed for this, I was also reminded of #404 `AxisRadial` and #405 `GridRadial/GridPolar` which I started a long time ago and never wrapped up. That was pre-TypeScript so they'll need a bit of work to land. If you have any interest in either of those I could abandon the PRs in favor of newer ones.

Excited for this!

sarathps93:
Thanks for the response @williaster . I will add `scaleRadial` and a demo for the new chart.
And regarding #404 & #405, does these PRs just lack the typescript support? If everything else is intact shall I re-use the existing code you had written in those?

williaster:
Hey @sarathps93 I believe #404 + #405 were in pretty good shape except for TypeScript for the new functionality itself, and it used the old project name `vx` instead of `visx`. 

[...]

PR #958, opened 2020-12-04
feat(*): added scale radial

sarathps93:
#### :rocket: Enhancements

- Added `Radial Scale` which will help in creating charts like `circular bar plot`. Adding it as a part of #832

### DIFF
--- a/packages/visx-scale/src/operators/unknown.ts
+++ b/packages/visx-scale/src/operators/unknown.ts
@@ -11,6 +11,6 @@ export default function applyUnknown<
   config: ScaleConfigWithoutType<Output, DiscreteInput, ThresholdInput>,
 ) {
   if ('unknown' in scale && 'unknown' in config && typeof config.unknown !== 'undefined') {
-    scale.unknown(config.unknown);
+    (scale.unknown as any)(config.unknown);
   }
 }
--- a/packages/visx-scale/src/types/ScaleConfig.ts
+++ b/packages/visx-scale/src/types/ScaleConfig.ts
@@ -55,6 +55,13 @@ export type SymlogScaleConfig<Output = DefaultOutput> = CreateScaleConfig<
   'clamp' | 'constant' | 'nice' | 'zero'
 >;
 
+export type RadialScaleConfig<Output = DefaultOutput> = CreateScaleConfig<
+  'radial',
+  ContinuousDomain,
+  Output[],
+  'clamp' | 'nice' | 'round' | 'unknown'
+>;
+
 export type QuantileScaleConfig<Output = DefaultOutput> = CreateScaleConfig<
   'quantile',
   ContinuousDomain,
@@ -127,6 +134,7 @@ export interface ScaleTypeToScaleConfig<
   pow: PowScaleConfig<Output>;
   sqrt: SqrtScaleConfig<Output>;
   symlog: SymlogScaleConfig<Output>;
+  radial: RadialScaleConfig<Output>;
   time: TimeScaleConfig<Output>;
   utc: UtcScaleConfig<Output>;
   quantile: QuantileScaleConfig<Output>;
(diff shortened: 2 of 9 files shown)


sarathps93:
Had to update the version of `d3-scale` and it's types package to get the `scaleRadial` and type from it

packages/visx-scale/src/operators/unknown.ts
```diff
@@ -11,6 +11,6 @@ export default function applyUnknown<
   config: ScaleConfigWithoutType<Output, DiscreteInput, ThresholdInput>,
 ) {
   if ('unknown' in scale && 'unknown' in config && typeof config.unknown !== 'undefined') {
-    scale.unknown(config.unknown);
+    (scale.unknown as any)(config.unknown);
```
sarathps93, line 14: The latest type definitions for `d3-scales` are causing issues at here and that's why I have including the type casting in here

packages/visx-scale/src/types/ScaleConfig.ts
```diff
@@ -55,6 +55,13 @@ export type SymlogScaleConfig<Output = DefaultOutput> = CreateScaleConfig<
   'clamp' | 'constant' | 'nice' | 'zero'
 >;
 
+export type RadialScaleConfig<Output = DefaultOutput> = CreateScaleConfig<
+  'radial',
+  ContinuousDomain,
+  Output[],
+  'clamp' | 'nice' | 'round' | 'zero' | 'unknown'
```
williaster, line 62: should `zero` be omitted here since it's not applied in `scales/radial.ts`?

williaster (commented):
This looks great to me with one minor comment on the `RadialScaleConfig`.

Also cc-ing @kristw in case he notices anything else since he's very familiar with the scale package.

packages/visx-scale/src/types/ScaleConfig.ts
sarathps93: Yeah makes sense. Have done the changes

kristw: +1

packages/visx-scale/src/operators/unknown.ts
kristw: Would you mind sharing the specific issue? 
Just want to help think if there is a workaround that is more type-safe than casting to `any`.

sarathps93: ` /**
     * Returns the current unknown value, which defaults to undefined.
     */
    unknown(): UnknownReturnType<Unknown, undefined>;
    /**
     * Sets the output value of the scale for undefined (or NaN) input values and returns this scale.
     *
     * @param value The output value of the scale for undefined (or NaN) input values.
     */
    unknown<NewUnknown>(value: NewUnknown): ScaleLinear<Range, Output, NewUnknown>;`

Above is the latest type definition from `@types/d3-scale`. Somehow the `scale.unknown` is pointing to the function which accepts no arguments.

sarathps93: @kristw Please let me if any other workaround can be done to avoid this.

williaster: It looks like TS isn't correctly matching the second overload. is there any way to import `NewUnknown` and coerce to that?

`scale.unknown(config.unknown as NewUnknown)`

sarathps93: @williaster I have tried that option as well. They are not exposing the `NewUnknown` for us to import

williaster: @kristw any other thoughts? else we can probably merge this as is

williaster (approved):
Spoke with @kristw offline, let's merge this! Thanks again for the contribution 🙏

PR #961, opened 2020-12-05
Refactor(responsive): rewrite enhancer by hook

susiwen8:
#### :rocket: Enhancements

- Refac `responsive` by `hook`.


packages/visx-responsive/src/hooks/useSize.ts
```diff
@@ -0,0 +1,54 @@
...
+  props: BaseComponentProps & WithSizeProvidedProps,
+  isWindow: boolean,
+) {
+  const { debounceTime = 300, enableDebounceLeadingCall = true, initWidth, initHeight } = props;
+  const [eleWidth, setEleWidth] = useState<number | undefined>(undefined);
+  const [eleHeight, setEleHeight] = useState<number | undefined>(undefined);
+  const resize = useCallback(
+    debounce(
+      (params?: { width: number; height: number } | UIEvent) => {
+        if (assertParams(isWindow, params!)) {
```
susiwen8, line 35: Need to find more elegant way to determine whose size it is. :thinking::thinking::thinking:

susiwen8:
Sorry, I'm little 'arrogant' now :joy::joy::joy: 

I think this is the last one who can be refactored

packages/visx-demo/src/components/Show.tsx
```diff
@@ -21,15 +21,15 @@ type ShowProps = {
...
   packageJson?: PackageJson;
 };
 
 const padding = 40;
 
-export default withScreenSize<ShowProps & WithScreenSizeProvidedProps>(
+export default withScreenSize<ShowProps & WithSizeProvidedProps>(
   ({
-    screenWidth,
+    width,
```
williaster, line 32: hey @susiwen8 thanks for the PR! I'll try to take a proper review pass either today or tomorrow but one thing I'd note is that we should try to avoid any API changes because those are breaking changes and will require us to release a `2.0` version.

If you'd like to improve/unify APIs, I think that's great 🎉  but I'd recommend supporting _both_ sets of props for now, and then we'd deprecate/officially remove the old ones when we want to do the `2.0` release.

packages/visx-responsive/src/hooks/useParentSize.ts
```diff
@@ -0,0 +1,36 @@
+import { RefObject, useRef, useEffect } from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
+import { useSize, WithSizeProps, WithSizeProvidedProps } from './useSize';
+
+export function useParentSize<BaseComponentProps extends WithSizeProps = {}>(
```
williaster, line 5: one thing that we've been considering is using [`react-use-measure`](https://github.com/pmndrs/react-use-measure) to power this instead as it has some more advanced scroll listener functionality built in. we could consider exporting `useMeasure` almost directly for `useParentSize`.

I'm not sure if it could be used for `useScreenSize`, may need to 👀  more closely

packages/visx-demo/src/components/Show.tsx
susiwen8: @williaster  Hi, I had no confident on this one. Will take a look.

packages/visx-responsive/src/hooks/useParentSize.ts
susiwen8: Great, I'll look it up this weekend.

PR #963, opened 2020-12-09
docs(xychart): add README, add docs/xychart page, add XYChart gallery tile

williaster:
#### :memo: Documentation

This PR
- fleshes out the `README` for `@visx/xychart` to cover most aspects of the API, including some example code snippets
  - note I've added `<details />` sections so that the rest of the `/docs/xychart` page with example tile(s) + comprehensive API will be visible without needed to scroll a lot
- adds a `/docs/xychart` page to the demo site, and the `XYChartTile` to the gallery 🎉 

I plan to add a couple more examples but non-gallery codesandboxes will be easier to create when the package is published

image

image

@kristw @hshoff

### DIFF
--- a/packages/visx-demo/src/components/Gallery/XYChartTile.tsx
+++ b/packages/visx-demo/src/components/Gallery/XYChartTile.tsx
@@ -5,7 +5,6 @@ import GalleryTile from '../GalleryTile';
 export { default as packageJson } from '../../sandboxes/visx-xychart/package.json';
 
 const tileStyles = { background: '#222' };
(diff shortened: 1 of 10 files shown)


williaster:
hmm, may not be able to cleanly create the `@visx/xychart/animated` index, right now I think because we have `main: lib/index.js` for the package, and that `index` doesn't reference `lib/animated/index.js` we don't get the correct types, etc.

williaster:
Going to merge this since it's been open for a while and is just docs w(hich are blocking `1.3.0` as it stands), happy to address any feedback in a following PR

PR #966, opened 2020-12-09
new(xychart): move all Animated* exports to @visx/xychart/animated

williaster:
#### :rocket: Enhancements

This PR builds on #963 and creates a new `@visx/xychart/animated` `index` export file that is equivalent to `@visx/xychart`, but contains `Animated*` versions of components. 

This enables consumers who **don't** want animated components / don't want to add the `react-spring` peer dependency to still import files from the `@visx/xychart` index without errors.

```ts
// before
import { Axis } from '@visx/xychart';  // errors if no `react-spring` peer dep
import { AnimatedAxis } from '@visx/xychart';

// after
import { Axis } from '@visx/xychart';  // fine if no `react-spring` peer dep
import { Axis as AnimatedAxis } from '@visx/xychart/animated'; // still requires `react-spring` peer dep
```

@hshoff @kristw


kristw (approved):

williaster:
Noting here that `@visx/xychart/animated` currently exports everything exported from `@visx/xychart`, even non-animated components like `XYChart` (where there is no animated equivalent), context, types, etc.

Another option would be to make it strictly `Animated` components, but that requires consumers to basically always have multiple import lines, `XYChart` from `@visx/xychart`, animated components from `/animated`. So let's go with the single import statement.

williaster:
noting that I've reverted this in the #963 branch as it breaks types/index (would need to import from `@visx/xychart/lib/animated` without additional fixes)

PR #984, opened 2020-12-29
Add XYChart to @visx/visx

ShaneHudson:
This PR relates to https://github.com/airbnb/visx/issues/974 to ensure XYChart is in the package that includes all other packages.

#### :boom: Breaking Changes
- none

#### :rocket: Enhancements
- Adds XYChart to the @visx/visx package.

#### :memo: Documentation
- none

#### :bug: Bug Fix
- none

#### :house: Internal
- none

### DIFF
--- a/packages/visx-visx/package.json
+++ b/packages/visx-visx/package.json
@@ -60,6 +60,7 @@
     "@visx/text": "1.3.0",
     "@visx/tooltip": "1.3.0",
     "@visx/voronoi": "1.0.0",
+    "@visx/xychart": "1.3.0",
     "@visx/zoom": "1.3.0"
   }
 }
--- a/packages/visx-visx/src/index.ts
+++ b/packages/visx-visx/src/index.ts
@@ -25,4 +25,5 @@ export * from '@visx/shape';
 export * from '@visx/text';
 export * from '@visx/tooltip';
 export * from '@visx/voronoi';
+export * from '@visx/xychart';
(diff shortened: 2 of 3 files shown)


ShaneHudson:
The build failed, it says to regenerate types but not sure how.

williaster:
@ShaneHudson thanks for the PR! ❤️  sorry for the review delay (holidays 🎁 ). 

I think this is failing because some of the named exports from `XYChart` conflict with exports from different packages 😱  this is (surprisingly) the first time for name clashes, so may need to rethink the approach. I think probably the best thing is to move toward name-spacing the exports into **separate files** to avoid this. Need to be cognizant of breaking changes tho so I would propose:

1) (this PR, no breaking changes): create a new `src/xychart.ts` file which has the `export * from '@visx/xychart';` line and avoids name conflicts.
2) (follow up PR, consistency with no breaking changes): create a new `src/xxx.ts` for each package, e.g., `src/shape.ts` would `export * from '@visx/shape';`.  the `src/index.ts` is unchanged.
3) (follow up PR, for breaking `2.0.0` release): remove `src/index.ts` and remove `package.json` `main` + `module` entries which reference this index.

I'll probably create an issue to track the multiple parts. If you want to combine `1` and `2` into this PR feel free else I'm happy to take a stab 🔜 . There's no planned `2.0.0` yet, but I'll start a project with known todo items for it.

PR #987, opened 2021-01-03
Fix scale options in xychart

Janpot:
#### :bug: Bug Fix

Fixes scale options  not applying in `XYChart`. See https://github.com/airbnb/visx/issues/986

`zero`, `nice`,... options are applied in `createScale`. they were overwritten by setting the `domain` and `range` afterwards. this PR makes sure the `domain` and `range` are set in the `createScale` call directly so that other options are applied in the correct order.

### DIFF
--- a/packages/visx-xychart/src/hooks/useScales.ts
+++ b/packages/visx-xychart/src/hooks/useScales.ts
@@ -31,7 +31,6 @@ export default function useScales<
   const memoizedXScale = useMemo(() => {
     const registryEntries = registryKeys.map(key => dataRegistry.get(key));
 
-    let xScale = createScale(xScaleConfig) as XScale;
     type XScaleInput = ScaleInput<XScale>;
 
     const xValues = registryEntries.reduce<XScaleInput[]>(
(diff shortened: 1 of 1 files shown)


packages/visx-xychart/src/hooks/useScales.ts
```diff
@@ -42,8 +41,11 @@ export default function useScales<
 
     const xDomain = isDiscreteScale(xScaleConfig) ? xValues : d3Extent(xValues);
 
-    xScale.range(xScaleConfig.range || [xMin, xMax]);
-    xScale.domain(xScaleConfig.domain || xDomain);
+    let xScale = createScale({
+      range: [xMin, xMax],
+      domain: xDomain as [number, number],
```
williaster, line 46: I think depending on scale type this won't always be a `number`, but I can try to fix the typing in a followup PR as types can be tricky to debug here.

williaster (approved):
@Janpot thanks for the bug report and fix 🙏 ❤️   

I encountered this myself before the holidays but didn't quite have time to submit a fix, this LGTM 🎉

PR #991, opened 2021-01-05
[annotation] Fix LineSubject types and strokeWidth

Janpot:
#### :bug: Bug Fix

It was impossible to set `strokeWidth` on `LineSubject`s because the property was never passed through. I also augmented the types with the svg line properties because I was getting type errors when trying to set `strokeDasharray`

### DIFF
--- a/packages/visx-annotation/src/components/LineSubject.tsx
+++ b/packages/visx-annotation/src/components/LineSubject.tsx
@@ -29,7 +29,6 @@ export default function LineSubject({
   min,
   max,
   stroke = '#222',
-  strokeWidth = 2,
   ...restProps
 }: LineSubjectProps & Omit<React.SVGProps<SVGLineElement>, keyof LineSubjectProps>) {
   // if props are provided, they take precedence over context
--- a/packages/visx-annotation/src/components/LineSubject.tsx
+++ b/packages/visx-annotation/src/components/LineSubject.tsx
@@ -19,7 +19,7 @@ export interface LineSubjectProps extends React.SVGProps<SVGLineElement> {
   min: number;
   /** The maximum coordinate of the line. */
   max: number;
-};
+}
 
 export default function LineSubject({
   className,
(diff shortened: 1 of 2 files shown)


packages/visx-annotation/src/components/LineSubject.tsx
```diff
@@ -29,9 +29,8 @@ export default function LineSubject({
   min,
   max,
   stroke = '#222',
-  strokeWidth = 2,
   ...restProps
-}: LineSubjectProps & Omit<React.SVGProps<SVGLineElement>, keyof LineSubjectProps>) {
+}: LineSubjectProps) {
```
williaster, line 33: can you send a link that reproduced the TS error you are seeing? It works okay for me in the `/annotation` [codesandbox](https://codesandbox.io/s/dazzling-jang-zxpkl?file=/Example.tsx)

image

For other packages/components we have preferred to set the native SVG props on the component as a union with its own exported prop types, rather than on the exported prop types themselves to keep them separate and simplify re-use of the export where relevant.

williaster (commented):
Thanks for the fix @Janpot. The `strokeWidth` fix definitely makes sense but I had one question about the type change.

packages/visx-annotation/src/components/LineSubject.tsx
Janpot: Screenshot 2021-01-06 at 22 09 27

I actually saw it on `AnnotationLineSubject`. I admit I didn't research this thoroughly enough and just assumed it propagated from `LineSubject`

Janpot: can be reproduced in the `xychart` codesandbox https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-xychart?file=/Example.tsx:7459-7501

Janpot: @williaster Update the PR to update `AnnotationLineSubject` instead

williaster (approved):
Awesome, thanks @Janpot ! LGTM. Will get this out in `1.4.0` hopefully later today.

PR #993, opened 2021-01-06
fix(stats): passing value 0 to valueScale

rsbh:
#### :bug: Bug Fix

- right now there is a check for the values before calling `valueScale` which is supposed to check `undefined` and `null` value but it is filtering out the value `0`
when an inverted linear scale is passed as `valueScale` the value 0 is at the top, not the bottom. And values can be 0 and it is a valid value for `valueScale` function.

### DIFF
--- a/packages/visx-stats/src/BoxPlot.tsx
+++ b/packages/visx-stats/src/BoxPlot.tsx
@@ -93,10 +93,10 @@ export default function BoxPlot({
   const valueRange = valueScale.range();
 
   const minValue = valueScale(min ?? 0);
-  const firstQuartileValue = firstQuartile ? valueScale(firstQuartile) ?? 0 : 0;
-  const medianValue = median ? valueScale(median) ?? 0 : 0;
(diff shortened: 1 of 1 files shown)


williaster (approved):
Thanks for the fix @rsbh! 🙏   I agree this is the desired behavior 👍

rsbh:
@williaster what is the ETA on the 1.4.0 release?

williaster:
@rsbh I was trying to get a few more PRs in this week before release but they'll probably land early next week. If you need the release urgently I can probably get a release out today with this fix.

rsbh:
Thanks @williaster. This is a bug fix and it will be better to release a patch version instead of waiting for next major/minor release.

williaster:
👍   yep, this was a bug fix which would normally be a patch, but there were already new features merged by the time this got in so went ahead with the `minor` release. [`1.4.0`](https://github.com/airbnb/visx/releases/tag/v1.4.0) is out now 🎉

PR #1008, opened 2021-01-08
fix(xychart): include zero in scale domains by default

williaster:
#### :bug: Bug Fix

This fixes a default behavior issue with `XYChart` scales, where their domains don't include `zero` by default if the scale allows. Including `zero` in the axis domain is generally best practice, and users can still opt out by setting `x/yScaleConfig={{ zero: false }}`. 

**Before** 
This `AreaSeries` has 3 y-values which are all equal and non-zero. It renders like this because the y-domain is `[100, 100]`. This was also reported in #1003 

**After**
With the same data + chart config, the y-domain is now `[0, 100]`. 

Verified it works for negative values, and horizontal orientation.

#### :rocket: Enhancements

- Adds a `scaleCanBeZeroed` typeguard util to `@visx/scale` (needed for typing these changes)
- exposes `DataProvider`'s `horizontal` prop in `XYChart` (noticed a small TODO during this work)

@hshoff @kristw 
cc @RIP21

### DIFF
--- a/packages/visx-xychart/src/hooks/useScales.ts
+++ b/packages/visx-xychart/src/hooks/useScales.ts
@@ -41,18 +41,18 @@ export default function useScales<
 
     const xDomain = isDiscreteScale(xScaleConfig) ? xValues : d3Extent(xValues);
 
-    let xScale = scaleCanBeZeroed(xScaleConfig)
+    let xScale = (scaleCanBeZeroed(xScaleConfig)
       ? createScale({
           range: [xMin, xMax],
           domain: xDomain as [XScaleInput, XScaleInput],
           zero: true,
           ...xScaleConfig,
         })
-      : (createScale({
+      : createScale({
           range: [xMin, xMax],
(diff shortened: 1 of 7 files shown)


packages/visx-xychart/src/hooks/useScales.ts
```diff
@@ -41,11 +41,18 @@ export default function useScales<
...
 
-    let xScale = createScale({
-      range: [xMin, xMax],
-      domain: xDomain as [number, number],
-      ...xScaleConfig,
-    }) as XScale;
+    let xScale = (scaleCanBeZeroed(xScaleConfig)
+      ? createScale({
+          range: [xMin, xMax],
+          domain: xDomain as [XScaleInput, XScaleInput],
```
williaster, line 47: noting that the domain was updated from `number` => `ScaleInput` (leftover from my comment [here](https://github.com/airbnb/visx/pull/987#discussion_r551586262))

hshoff (approved):

PR #1019, opened 2021-01-13
new(xychart): add AreaStack

williaster:
#### :rocket: Enhancements

This closes #994 and 
- adds an `AreaStack` series to support stacked areas + stream graphs
- consolidates `BarStack` and `AreaStack` logic in a new `useStackedData` hook

#### :bug: Bug Fix
- fixes scale domain logic in `useStackedData` for different stack `offset`s (e.g., `wiggle` for streamgraphs; only `diverging` and `none` worked previously)

**TODO** (separate PRs to keep size down)
- [x] add to `/xychart` demo #1020
- [x] add tests #1036

Preview images from the demo branch:

**Basic**

**Tab-able tooltip**

**Stack offsets**

@hshoff @kristw

PR #1020, opened 2021-01-13
new(demo/xychart): add AreaStack, add stackOffset variable for stacks

williaster:
#### :rocket: Enhancements

This updates the `/xychart` demo to include `AreaStack` added in #1019 (see images there), and also adds a toggle for `stackOffset` for both `AreaStack` and `BarStack`.

@hshoff @kristw

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
@@ -65,6 +65,7 @@ type ProvidedProps = {
   setAnnotationDataKey: (key: DataKey | null) => void;
   setAnnotationLabelPosition: (position: { dx: number; dy: number }) => void;
   renderAreaSeries: boolean;
+  renderAreaStack: boolean;
   renderBarGroup: boolean;
   renderBarSeries: boolean;
   renderBarStack: boolean;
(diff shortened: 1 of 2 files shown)


packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
```diff
@@ -720,6 +792,11 @@ export default function ExampleControls({ children }: ControlsProps) {
         input[type='radio'] {
           height: 10px;
         }
+        .pattern-lines {
```
williaster, line 795: this currently pattern-only svg has a non-zero height on `/xychart`, this fixes it.

packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
```diff
@@ -440,24 +465,35 @@ export default function ExampleControls({ children }: ControlsProps) {
           <label>
             <input
               type="radio"
-              onChange={() => setRenderBarStackOrGroup('bar')}
+              onChange={() => {
+                if (renderAreaLineOrStack === 'areastack') {
+                  setRenderAreaLineOrStack('none');
```
williaster, line 470: previously I disabled updating `renderAreaLineOrStack ` depending on the value of `renderBarStackOrGroup`, I think it's better UX if I just update the value of `renderAreaLineOrStack` so it's valid.

hshoff (approved):

PR #1030, opened 2021-01-22
test(happo): add Happo for screenshot testing

williaster:
#### :house: Internal

This PR adds [happo.io](https://happo.io) to `visx` so that we have more robust visual testing in our CI 🍾  🥳 . 

It specifically

- adds a `happo` job to our github `ci` workflow
  - ~split `ci` from one job into `build`/`test`/`lint`/`happo` jobs – would love 👀  on this because I'm new to github actions~ 
    - EDIT: I undid this because the output of a `build` job can't be used by `lint`/`test`/`happo` 
  - enabled the `happo` app so it can post on PRs
  - added appropriate secrets to the repo
  - setup account details with @lencioni 
- configured `happo` to run on all of our `@visx/demo` **gallery tiles**
  - added `happo.io` dep + configuration 
    - specifically I am using the `next.js`-oriented configuration which mostly just leverages the `next.js` webpack config
  - fixed some copy/paste issues from the `sandbox` examples to ensure a unique test names (taken from sandbox `package.json` `name` field)

Locally I was able to test it on all examples (here's a sample report image), will see how CI goes 🤞 

@hshoff @kristw @lencioni


packages/visx-demo/package.json
```diff
@@ -66,6 +68,7 @@
     "@visx/xychart": "1.4.0",
     "@visx/zoom": "1.3.0",
     "@zeit/next-css": "^1.0.1",
+    "babel-loader": "8.2.2",
```
williaster, line 71: needed by `happo.io`

packages/visx-demo/src/sandboxes/visx-annotation/Example.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Label, Connector, CircleSubject, LineSubject } from '@visx/annotation';
 import { LinePath } from '@visx/shape';
+// import resizeObserverPolyfill from 'resize-observer-polyfill';
```
williaster, line 4: oops, remove me

packages/visx-demo/src/sandboxes/visx-annotation/package.json
```diff
@@ -16,6 +16,7 @@
     "react": "^16",
     "react-dom": "^16",
     "react-scripts-ts": "3.1.0",
+    "resize-observer-polyfill": "1.5.1",
```
williaster, line 19: remove

.github/workflows/ci.yml
```diff
@@ -2,41 +2,60 @@ name: CI
...
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    steps:
```
williaster, line 12: `build`/`test`/`lint` are unchanged except split into multiple jobs with `needs: build` added. looks like this might have broke CI

packages/visx-demo/.happo.js
```diff
@@ -0,0 +1,66 @@
...
+        delete config.resolve.alias[key];
+      }
+    });
+    config.module = base.module;
+    return config;
+  },
+
+  // happo is unable to resolve some imports if the tmpdir isn't located inside
+  // the project structure. The default is an OS provided folder, `os.tmpdir()`.
+  tmpdir: path.join(__dirname, happoTmpDir),
```
trotzig, line 65: Good catch, I've seen this happen myself. I'll make a note to change the default tmp folder to something like `./.happo-tmp`.

packages/visx-demo/happo/gallery.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export default snapshots.concat([
+  // needs timeout for animated axes
+  {
+    component: '@visx/demo-axis',
+    variants: {
+      default: renderInDom => {
+        return new Promise(resolve => {
+          renderInDom(<AxisTile />);
+          setTimeout(() => resolve(), MAX_TIMEOUT_MS);
```
trotzig, line 38: nit - `setTimeout(() => resolve(), MAX_TIMEOUT_MS)` could be written as `setTimeout(resolve, MAX_TIMEOUT_MS)`

trotzig (approved):
Nice! 

I can't see the CI build results, so I can't help you there unfortunately.

packages/visx-demo/happo/gallery.tsx
```diff
@@ -0,0 +1,55 @@
...
+        });
+      },
+    },
+  },
+  // needs timeout for animated axes
```
lencioni, line 43: One option that you might want to consider is to not use an animation if prefers-reduced-motion is set to reduce, and then add `prefersReducedMotion: true,` to your RemoteBrowserTarget:

```ts
'chrome-desktop': new RemoteBrowserTarget('chrome', {
  viewport: '800x552',
  prefersReducedMotion: true,
}),
```

.github/workflows/ci.yml
```diff
@@ -2,41 +2,60 @@ name: CI
...
-    - name: Report code coverage
-      uses: coverallsapp/github-action@v1.1.1
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-    - name: Run lint
-      run: yarn run lint
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm ci
+      - run: npm run happo-ci-github-actions
```
lencioni, line 58: It looks like you are using yarn in this repo, why use npm here?

packages/visx-demo/happo/gallery.tsx
```diff
@@ -0,0 +1,55 @@
...
+type HappoSnapshot = {
+  component: string;
+  variants: {
+    [key: string]: (
+      renderInDom: (component: React.ReactElement) => void,
+    ) => React.ReactNode | Promise<unknown>;
+  };
+};
+
+const MAX_TIMEOUT_MS = 400;
```
lencioni, line 15: Why 400? Is there a variable somewhere else that this could import and be based on?

.github/workflows/ci.yml
williaster: good catch 👍

packages/visx-demo/happo/gallery.tsx
williaster: ideally this is in sync with `.happo.js`, will update to import from there.

packages/visx-demo/happo/gallery.tsx
williaster: this is a cool idea. it'll be a bit of work to re-write those examples because the animations aren't css-based. looks like you can do the media query in JS to detect it, so I can swap out animated components for their static analogs. I may do this in a follow up PR depending on how big the change is.

.github/workflows/ci.yml
```diff
@@ -2,41 +2,54 @@ name: CI
...
+      - name: Run lint
+        run: yarn lint
+
+      - name: Run happo
+        run: yarn run happo-ci-github-actions
+        working-directory: './packages/visx-demo/'
+        env:
+          HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
+          HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
+          HAPPO_COMMAND: '../../node_modules/happo.io/build/cli.js'
```
williaster, line 55: noting that the `happo-ci` binary expects this path to be in the same directory, but because this is run in `./packages/visx-demo/` it needs to point to the monorepo root `node_modules` where `yarn` installs it.

packages/visx-demo/happo/gallery.tsx
williaster: couldn't import `asyncTimeout` directly from `.happo.js` because `node`'s `child_process` gets pulled into the webpack config step which throws. put it in a separate file.

williaster:
🎉  report looks good. I'm going to create one more PR to add a stable `Math.random` to use in examples so that we don't get spurious diffs (this is used across ~5 demos right now). That can merge and then we can merge this one.

packages/visx-demo/happo/gallery.tsx
```diff
@@ -0,0 +1,54 @@
...
+          setTimeout(resolve, asyncTimeout);
+        });
+      },
+    },
+  },
+  // needs timeout for animated axes
+  {
+    component: '@visx/demo-xychart',
+    variants: {
+      default: renderInDom => {
```
kristw, line 46: nit: can extract to function and use in both places?

kristw (approved):
super cool

PR #1036, opened 2021-01-26
test(xychart): add AreaStack tests

williaster:
#### :house: Internal

This adds tests for `useStackedData` and `AreaStack` added in #1019 

#### :bug: Bug Fix

In the process of adding tests I fixed two `DataProvider` issues which can result in invalid scales on initial render:
- an invalid object spread in `useDimensions` could result in `width`, `height`, or `margin` being `undefined` upon initialization, resulting in invalid `scale.range`
- `d3` scales initialized with `domain: []` result in `domain = [NaN, NaN]`, which previously could happen when no data was present in the data registry (now we defer `scale` initialization until data is present)

@kristw @hshoff

### DIFF
--- a/packages/visx-xychart/src/components/axis/BaseAxis.tsx
+++ b/packages/visx-xychart/src/components/axis/BaseAxis.tsx
@@ -66,7 +66,11 @@ export default function BaseAxis<Scale extends AxisScale>({
       ? (width ?? 0) - (margin?.right ?? 0)
       : 0;
 
-  return (
+  const scale = (orientation === 'left' || orientation === 'right' ? yScale : xScale) as
+    | Scale
+    | undefined;
+
+  return scale ? (
     <AxisComponent
       top={topOffset}
       left={leftOffset}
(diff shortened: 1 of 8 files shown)


packages/visx-xychart/src/components/axis/BaseAxis.tsx
```diff
@@ -66,7 +66,11 @@ export default function BaseAxis<Scale extends AxisScale>({
       ? (width ?? 0) - (margin?.right ?? 0)
       : 0;
 
-  return (
+  const scale = (orientation === 'left' || orientation === 'right' ? yScale : xScale) as
```
kristw, line 69: Would this look cleaner?
`const scale: Scale | undefined = ...`

kristw (approved):

packages/visx-xychart/src/components/axis/BaseAxis.tsx
williaster: Definitely ideal, but the type won't work without a coercion either here, or when I get the scales from `DataContext` (I haven't successfully figured out how to use generics with `Context` types; most Series coerce the `Context` values right now)

PR #1037, opened 2021-01-27
new(demo): respect prefersReducedMotion and leverage in happo

williaster:
This PR adds the following logic discussed in #1030

#### :memo: Documentation
- checks for the `prefersReducedMotion` media query in the `/xychart`, `/axis`, and `/patterns` demos, and uses animated or unanimated components accordingly ~(also removes `animationTrajectory` controls accordingly)~.

EDIT: even if `prefersReducedMotion=true`, users can still toggle animation on/off.

#### :house: Internal
- sets `prefersReducedMotion: true` for happo tests, and removes the happo rendering timeouts that were previously needed for animations

Tested the happo examples locally, and tested gallery examples with / without `prefersReducedMotion` set on my OS.

@lencioni @kristw @hshoff

### DIFF
--- a/packages/visx-demo/happo/gallery.tsx
+++ b/packages/visx-demo/happo/gallery.tsx
@@ -10,17 +10,13 @@ type HappoSnapshot = {
   };
 };
 
-const specialCases = new Set(['@visx/demo-xychart', '@visx/demo-axis']);
-
 const getComponentName = (Example: typeof examples[0]) =>
   Example.packageJson.name || 'missing-name';
 
-const snapshots: HappoSnapshot[] = examples
-  .filter(Example => specialCases.has(getComponentName(Example)))
-  .map(Example => ({
-    // note: this (reasonably) asserts Examples have unique names
-    component: getComponentName(Example),
-    variants: { default: () => <Example.default /> },
-  }));
+const snapshots: HappoSnapshot[] = examples.map(Example => ({
+  // note: this (reasonably) asserts Examples have unique names
+  component: getComponentName(Example),
+  variants: { default: () => <Example.default /> },
+}));
 
 export default snapshots;
--- a/packages/visx-demo/happo/gallery.tsx
+++ b/packages/visx-demo/happo/gallery.tsx
@@ -10,13 +10,13 @@ type HappoSnapshot = {
   };
 };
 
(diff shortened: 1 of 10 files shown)


packages/visx-demo/happo/gallery.tsx
```diff
@@ -13,17 +10,7 @@ type HappoSnapshot = {
...
-// renders an example with a timeout
-const renderWithTimeout: (
-  Example: React.ReactElement,
-) => HappoSnapshot['variants'][string] = Example => renderInDom => {
-  return new Promise(resolve => {
-    renderInDom(Example);
-    setTimeout(resolve, asyncTimeout);
-  });
-};
+const specialCases = new Set([]);
```
lencioni, line 13: Any reason to keep this around?

packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
```diff
@@ -742,41 +746,43 @@ export default function ExampleControls({ children }: ControlsProps) {
...
-          <label>
-            <input
-              type="radio"
-              onChange={() => setAnimationTrajectory('max')}
-              checked={animationTrajectory === 'max'}
-            />
-            from max
-          </label>
-        </div>
+        {!prefersReducedMotion && (
```
lencioni, line 749: I think it might be nicer to always render these controls, and then just disable them and add a message explaining why so that people who visit this page with prefers reduced motion are still able to see how it can be used. It would be even better if there was a way for these people to override and enable the animations anyway if they want to see the demo with the animations without having to mess with their OS settings or whatever.

lencioni (approved):
Love it!

packages/visx-demo/src/sandboxes/visx-xychart/ExampleControls.tsx
williaster: good call, I updated the examples to have an `enable animation` toggle with a default determined by `prefersReducedMotion`.

image

packages/visx-demo/happo/gallery.tsx
williaster: initially I thought it'd still be useful for testing specific examples quickly. removed.

PR #1044, opened 2021-02-03
new(axis): pass all tick values in tickLabelProps signature

williaster:
#### :rocket: Enhancements

This updates the signature of `@visx/axis`'s `Axis.props.tickLabelProps`:

```ts
// before
(tickValue, index) => TextProps

// after
(tickValue, index, tickValues) => TextProps
```

This is consistent with how many `d3` accessors work (gives you access to the data point, index, and all data points) as well as matches the signature of [`Axis.props.tickFormat`](https://github.com/airbnb/visx/blob/master/packages/visx-axis/src/types.ts#L16). 

I noticed this as a limitation when using `xychart` in an internal app (e.g., if you want to right-align the last tick, but don't know how many ticks you have since `numTicks` is approximate).

@kristw @hshoff

### DIFF
--- a/packages/visx-axis/src/types.ts
+++ b/packages/visx-axis/src/types.ts
@@ -19,7 +19,7 @@ export type TickFormatter<T> = (
   values: { value: T; index: number }[],
 ) => FormattedValue;
 
-export type TickLabelProps<T> = (value: T, index: number) => Partial<TextProps>;
+export type TickLabelProps<T> = (value: T, index: number, ticks: T[]) => Partial<TextProps>;
 
 export type TickRendererProps = Partial<TextProps> & {
   x: number;
--- a/packages/visx-axis/src/types.ts
+++ b/packages/visx-axis/src/types.ts
@@ -19,7 +19,11 @@ export type TickFormatter<T> = (
   values: { value: T; index: number }[],
 ) => FormattedValue;
 
-export type TickLabelProps<T> = (value: T, index: number, ticks: T[]) => Partial<TextProps>;
+export type TickLabelProps<T> = (
+  value: T,
+  index: number,
+  values: { value: T; index: number }[],
+) => Partial<TextProps>;
 
 export type TickRendererProps = Partial<TextProps> & {
   x: number;
--- a/packages/visx-axis/test/Axis.test.tsx
+++ b/packages/visx-axis/test/Axis.test.tsx
@@ -78,15 +78,22 @@ describe('<Axis />', () => {
     expect.hasAssertions();
   });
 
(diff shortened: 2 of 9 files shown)


packages/visx-axis/test/Axis.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('<Axis />', () => {
     expect(wrapper.find(`.${tickClassName}`).length).toBeGreaterThan(0);
   });
 
-  test('it should pass the output of tickLabelProps to tick labels', () => {
```
williaster, line 69: have been updating `test('it ...')` to `it('...')` as I've encountered them when updating tests

packages/visx-axis/src/types.ts
```diff
@@ -19,7 +19,11 @@ export type TickFormatter<T> = (
   values: { value: T; index: number }[],
 ) => FormattedValue;
 
-export type TickLabelProps<T> = (value: T, index: number) => Partial<TextProps>;
+export type TickLabelProps<T> = (
+  value: T,
+  index: number,
+  values: { value: T; index: number }[],
```
hshoff, line 25: Does this mean tsc will require folks to update the arguments for every tickLabelProps? guess I'm trying to figure out if a tsc warning/error on this would break builds without everyone going in and updating all of their functions to the new signature

kristw: Passing function that takes fewer arguments should be ok.
Passing function that requires more argument is not.

```ts
type MyFnType = ((a: string) => string);

// OK
const f1: MyFnType = () => 'a';
// OK
const f2: MyFnType = (a: string) => a;
// error
const f3: MyFnType = (a: string, b: string) => 'a';
```

kristw (approved):
👍

packages/visx-axis/src/types.ts
williaster: ^yeah I don't think that using the old signature will be problematic `(value, index) => TextProps` shouldn't throw. One thing that would break is if folks are using `TickLabelProps` explicitly in their typing:

```ts
// error
(value, index): TickLabelProps<ValueType> => TextProps
```

That would throw because the signature doesn't match (`XYChart` wraps `Axis` and uses that type, so I had to update the signature [on this line](https://github.com/airbnb/visx/pull/1044/files#diff-a1e01e13b948f291db2d3d3506abe738e077f948a5f4fecf2fb543d2e11c5624R37))

I'd guess that would be rare. I don't think we've considered type changes like that breaking historically (e.g., bumping d3 types wasn't breaking even though there were maybe breaking changes)

PR #1045, opened 2021-02-03
new(tooltip, xychart): force update TooltipInPortal bounds

### LINKED ISSUE #983: [xychart] Tooltip does not update when chart position changes
https://user-images.githubusercontent.com/1286001/103257155-2f4eda00-49e4-11eb-8eb9-0d98c701b96e.mov

This behaviour happens when using visx in the `react-grid-layout` library. When the chart is moved, the x and y positions are not updated and the tooltip is displayed where the chart used to be. After a resize, the since the width prop has changed, the tooltip is rendered correctly.

I can work around this by passing the chart's X and Y position in as a prop, but I feel like this may be something that the library can solve automatically for me. Thoughts?

williaster:
#### :rocket: Enhancements

This PR fixes an issue with `XYChart`'s `Tooltip` described in #983, where `Tooltip` position does **not** correctly update when the chart container changes position due to limitations with `react-use-measure` (note that position **does** update upon page scroll or container resize). It's not obvious how to fix this with `IntersectionObserver`, etc. so I used a relatively simple approach:

- expose `react-use-measure`'s `forceRefreshBounds` function in the `@visx/tooltip` `useTooltipInPortal` hook, which will force update the `Tooltip` container bounds
- invoke ^`forceRefreshBounds` whenever `@visx/xychart`'s `Tooltip` goes from the `unopen => open` state

Closes #983.

**Before**
In this case, the chart animates in from the left so the `Portal` thinks its bounds are way left/off screen until I scroll, at which point `react-use-measure` updates the position and the tooltip is correctly positioned; see #983 for a different example with drag & drop

**After**
Tooltip has correct position from the outset

@kristw @hshoff 
cc @valtism

### DIFF
--- a/packages/visx-tooltip/src/hooks/useTooltipInPortal.tsx
+++ b/packages/visx-tooltip/src/hooks/useTooltipInPortal.tsx
@@ -10,6 +10,7 @@ export type TooltipInPortalProps = TooltipProps & Pick<UseTooltipPortalOptions, 
 export type UseTooltipInPortal = {
   containerRef: (element: HTMLElement | SVGElement | null) => void;
   containerBounds: RectReadOnly;
+  forceRefreshBounds: () => void;
   TooltipInPortal: React.FC<TooltipInPortalProps>;
 };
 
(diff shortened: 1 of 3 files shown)


valtism:
Amazing! Thank you

kristw (approved):

VigneshTrantor:
Thank you

PR #1068, opened 2021-02-12
fix(xychart): improve positioning with missing data

### LINKED ISSUE #1054: Hovering on point with no data first renders glyph and tooltip outside chart
When I have a point with a y value that is undefined (representing a data gap) and I hover over it without first hovering over other points (e.g. moving mouse in from under x axis to this point), the glyph and tooltip do this:

Screenshot (29)

williaster:
#### :bug: Bug Fix

Closes #1054. 

Previously when either `snapTooltipToDataX`/`snapTooltipToDataY` were set on `<Tooltip />`, and a hovered `datum` had a missing value for the corresponding `x`/`y`, we rendered the tooltip at `0,0` and also rendered the tooltip glyphs (circles corresponding to the hovered points) incorrectly.

Also fixes the demo to not show `NaN` values for missing data.

**Before**

**After**

@kristw @hshoff 
cc @mira-t


mira-t:
Thank you very much! Really appreciate this change

PR #1073, opened 2021-02-18
new(workflows/push): automate releases

williaster:
#### 🏠  Internal

This PR adds automated releases to the github `push` workflow (i.e., merges into `master`). I wrote most everything custom because I couldn't find anything compelling for both the automated release + changelog updates without also adopting conventional commits which I think would be tricky. Npm auth with lerna was tricky 🥴 .

In the future the release cycle would work like the following:

- if a PR is merged **without** a release label (`patch-release`, `minor-release`, `major-release`, `alpha-release`), nothing will happen
- if a PR is merged **with** a release label, a release will be triggered
- a release includes the following:
  - git tags + npm packages are published using lerna (same as now) with a version tag (patch, minor, major, ±alpha) that matches any PR release labels found since the last release
  - for non-alpha releases, the changelog is updated using all PRs found since the last release:
    - PR titles are used for changelog entries, with a link to the PR
    - changelog categorization is done using PR labels (bug, enhancement, internal, docs, breaking). if the PR has multiple labels, it will be listed in multiple categories in the changelog
    - contributors are listed
  - after the release (including alpha releases), the `github-bot` posts on all PRs since the last release informing the author(s) which version the PR was released as

I tested this as much as I could on a branch, so hopefully it will go smootly upon merge.

Example commits

Example changelog diff

Example posting on PRs

@kristw @hshoff


scripts/performRelease/performLernaRelease.ts
```diff
@@ -0,0 +1,54 @@
...
+    pr.labels.some(label => label.name === MINOR_RELEASE),
+  );
+  const isMajor = releasePRsSinceLastTag.some(pr =>
+    pr.labels.some(label => label.name === MAJOR_RELEASE),
+  );
+
+  // perform release
+  try {
+    const version = `${isPreRelease ? 'pre' : ''}${
+      isMajor ? 'major' : isMinor ? 'minor' : 'patch'
```
williaster, line 34: if only `alpha-release` is set, this will default to `patch` which I think is safe

.github/workflows/push.yml
```diff
@@ -31,14 +32,29 @@ jobs:
...
       - name: 📐 Commit package sizes
+        # note: `git diff-index --quiet HEAD`
+        #       has exit code 0 if there are changes vs HEAD, else nothing to commit
         run: |
           yarn build:sizes	
           git config user.name github-actions	
           git config user.email github-actions@github.com	
           git add .	
-          git commit -m "build(${GITHUB_SHA}): auto-commit package sizes"	
+          git diff-index --quiet HEAD || git commit -m "build(${GITHUB_SHA}): auto-commit package sizes"
```
williaster, line 42: if there were no package size changes, this was failing before

scripts/performRelease/updateChangelog/mergeUpdateIntoChangelog.ts
```diff
@@ -0,0 +1,72 @@
...
+    # v1.5.0
+    ...
+ */
+export default function mergeUpdateIntoChangelog(
+  currChangelog: string,
+  changelogAddition: string,
+  tagName: string,
+) {
+  // split into all links and actual changelog content
+  const [currChangelogLinks, currChangelogContent] = currChangelog.split('------');
```
williaster, line 34: maybe should pull `------` into a constant since it's used below

kristw (approved):
LGTM. Thanks for all the research. A lot of customization here.

PR #1078, opened 2021-02-21
fix(network): string type in node/linkComponent

### LINKED ISSUE #939: bug(network): string type is invalid for node/linkComponent
Hi,

I started using @visx/network and tried to do customization for `nodeComponent `and `linkComponent`. But that's not working and the graph is unable to render. When I inspected and custom components are loaded in DOM.

Attached a sandbox demo [https://codesandbox.io/s/visxnetwork-customization-fuz8j?file=/Example.tsx](url)

Did I do anything wrong?. Please help me to render custom components

jraymakers:
Fixes #939 

#### :boom: Breaking Changes

- This changes the type signature of a public-facing prop, so it might be considered a breaking change. However, the type option that's being removed was broken anyway, so this doesn't break any functionality.

#### :memo: Documentation

- Extended `network` example to show use of `nodeComponent` and `linkComponent`, as well as `top` and `left` props. Adapted from demo in comment of https://github.com/airbnb/visx/issues/939

#### :bug: Bug Fix

- Remove string type from `nodeComponent` and `linkComponent` props in `network` package. This type was invalid in spite of being in the type signature. See https://github.com/airbnb/visx/issues/939

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-network/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-network/Example.tsx
@@ -33,11 +33,7 @@ export default function Example({ width, height }: NetworkProps) {
         graph={graph}
         top={50}
         left={100}
-        nodeComponent={() => (
-          <text dx="-0.5em" dy="0.5em" fontSize={28}>
-            💜
-          </text>
-        )}
+        nodeComponent={() => <rect x={-25} y={-15} width={50} height={30} fill="purple" />}
         linkComponent={({ link: { source, target } }) => (
           <>
             <line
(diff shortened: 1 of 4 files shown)


jraymakers:
I can't tell why Happo failed, because I can't follow the Detail links. But, it seems likely it's detecting the changes I made (intentionally) to the `network` gallery page.

williaster:
Hey @jraymakers sorry for the delay, thanks for the fix! I agree this shouldn't be considered breaking since it was broken :)

We recently added happo and there are a couple of demos that are flaky, apologies for making you think you did something wrong and that you can't access the report. This is what it's reporting for your change: 

image

It's a pretty big change aesthetically, I'm wondering if we could update the logic such that a single link is changed and a single node is changed. This would demonstrate the customizability and keep the aesthetics similar to before.

jraymakers:
Sure, that makes sense. I'll make the customizations more subtle.

jraymakers:
@williaster I changed the example to just have one custom node and one custom link, and used styling that should fit in better with the defaults and the other gallery tiles. I also fixed the clipping issue.

As an added bonus, the example now shows how to use custom node and link types with this generic component.

williaster (approved):
@jraymakers amazing, it looks great! thanks for being open to iterating on styles 😄 

image

Heads up I'm working on automating releases, I think this could be a good candidate for testing but it may take me a couple days to get it right.

PR #1094, opened 2021-03-04
fix(xychart/Axis): export AxisScale type, set default Scale generic

williaster:
#### :bug: Bug Fix

Somewhere between a bug fix and a feature, this just updates the `@visx/xychart` `Axis` component to export the `AxisScale` generic type so users don't need to import it from (or even install) `@visx/axis`. It also sets a default for the generic. 

I'd maybe consider it a :bug: in the sense that users can't reference the correct generic without `@visx/axis`.

@kristw @hshoff


kristw:
Deep export should not be a problem. 
Not sure if it will have conflict in the umbrella package if we also export from from top-level (`index`) of `@visx/xy-chart`?. I usually feel deep import is kinda hacky and users need to know where to find the type from.

williaster:
Good call @kristw , updated to export the type from `types/index` which is exported in the package `index`.

kristw (approved):

PR #1097, opened 2021-03-05
new(shape/BarRounded): export useBarRounded, support render functions

williaster:
#### :rocket: Enhancements

Title says it all, this is needed to leverage this shape in `@visx/xychart` (second PR coming)

@kristw @hshoff


packages/visx-shape/src/shapes/BarRounded.tsx
```diff
@@ -35,34 +35,50 @@ export type BarRoundedProps = {
...
+  children?: ({ path }: { path: string }) => React.ReactNode;
 };
 
-export default function BarRounded({
-  className,
-  innerRef,
-  x,
-  y,
-  width,
+/** Hook that returns a BarRounded path. */
```
kristw, line 42: The name `useXXX` misled me a bit. I thought this is a `react` hook but it just returns a path string.
Could go with `createBarRoundedPath()` for naming as well.

williaster: the idea is that it could be used as a hook, hence the `useXXX`, I'll update to `useBarRoundedPath` tho I agree that's clearer.

kristw (approved):

packages/visx-shape/test/BarRounded.test.tsx
```diff
@@ -31,63 +32,70 @@ describe('<BarRounded />', () => {
     const PathElement = container.querySelector('path');
     expect(fakeRef.current).toContainElement(PathElement);
   });
+
```
williaster, line 35:
```suggestion

```

PR #1098, opened 2021-03-05
new(xychart/BarSeries,BarGroup,BarStack): add support for bar radius

### LINKED ISSUE #1092: BarRounded with AnimatedBarSeries and XYChart
Is this possible with Bar Series or do we need to create our own shape?

williaster:
#### :rocket: Enhancements

This builds on #1097, and adds support for rounded `BarSeries` bars in `@visx/xychart` (including as children of `BarGroup`, `BarStack`) through the `radius`/`radiusAll`/`radiusTop`/`radiusBottom`/`radiusLeft`/`radiusRight` props. 

Note that when a `radius` is specified, the bar will be rendered with a `path` instead of a `rect`. I'm debating about **always** rendering a `BarRounded` (even if `radius=0`) for consistency, but I think 
- that would be a breaking change
- `path`s currently cannot animate on entrance `0 => value` (same for lines/areas right now) while `rect`s can. I think this is a limitation of `d3-interpolate-path`.

Closes #1092 and #1093.

Note that if users want full control over rendering bars (discussed in #1092), they can import `@visx/xychart/lib/components/series/private/BaseBarSeries` and specify a `BarsComponent`.

**BarSeries**
```tsx
<BarSeries radius={4} radiusRight {...} />
```

**BarStack**
```tsx
<BarStack {...}>
   <BarSeries radius={6} radiusTop {...} />
   <BarSeries {...} />
   <BarSeries radius={6} radiusBottom {...} />
</BarStack>
```

**BarGroup**
```tsx
<BarGroup {...}>
   <BarSeries radius={4} radiusTop {...} />
</BarGroup>
```

@kristw @hshoff 
cc @tonyneel923 @goldenivan

### DIFF
--- a/packages/visx-xychart/src/components/series/private/Bars.tsx
+++ b/packages/visx-xychart/src/components/series/private/Bars.tsx
@@ -1,4 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { AxisScale } from '@visx/axis';
+import { BarRounded } from '@visx/shape';
 import React from 'react';
 import { BarsProps } from '../../../types';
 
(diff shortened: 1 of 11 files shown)


packages/visx-xychart/src/components/series/private/Bars.tsx
```diff
@@ -7,21 +9,43 @@ export default function Bars({
...
+            tabIndex={isFocusable ? 0 : undefined}
+            {...barProps}
+            {...restProps}
+          />
+        ) : (
+          <BarRounded
+            key={key}
+            className="visx-bar"
+            tabIndex={isFocusable ? 0 : undefined}
+            radius={radius}
```
kristw, line 38: I forgot when `BarRounded` was added why `radiusTop` was called `top` instead of `radiusTop`.

kristw (commented):
What is the current behavior of the animation when transition from radius = 0 to radius != 0 (which probably cause a switch between `rect` and `path`)?

packages/visx-xychart/src/components/series/private/Bars.tsx
williaster: it was contributed by someone else. I think in the context of `BarRounded`, `top/right/bottom/left/topRight/etc` are more self evident (esp with types), but not as obvious here.

williaster:
The `rect => path` switch only occurs when `radius` changes from `null => number`, so `0 => 8` wouldn't cause that. There does seem to be some issue with how `d3-interpolate-path` handles the animation though, there is a flicker with the final state, and then the animation occurs. Not sure if this is because of how we define the `path` for `BarRounded`, I'll tinker with it.

Mar-08-2021 11-42-26

williaster:
In considering the requirements in #1093 more, specifically this requirement in a `BarStack`

> in our barStack, we have not every values on every bars. So the last bar is not always the same.

 I think this probably needs to be updated to be set-able per bar through an accessor. So I'll try to update to this type of API before merging.

tonyneel923:
Has there been any progress on this pr?

kristw (approved):

williaster:
Sorry I haven't had bandwidth to follow up on setting this per-bar 😱

kyythane:
@williaster I was wondering if it would be possible to merge the version without per-bar support? My team would really appreciate being able to use rounded bars. If not, would it be possible for me to take over this PR?

williaster:
thanks for the ping @kyythane 🙏  yeah we can just merge this since the per-bar feature hasn't come up again recently.

if we want to add that in the future and it needs a breaking change we can release a `3.0.0` or whatever it is at the time 👍  this PR depends on another one, so will work on getting them both in now.

kyythane:
@williaster Thank you!!

williaster:
verified this worked again since it's been so long. and looks like the animation twitching is fixed since the react-spring@9 upgrade 🎉

PR #1103, opened 2021-03-12
deps(lerna): bump to 3.22.0

williaster:
#### :house: Internal

Our last auto-release bumped **all** packages rather than just those which changed. Seeing if a `lerna` version bump fixes this.

@kristw @hshoff

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,4 @@
 {
-  "lerna": "3.0.0",
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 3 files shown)


kristw (approved):

lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "lerna": "3.0.0",
```
williaster, line 2: noting this field has been obsolete since `3.0.0`

PR #1105, opened 2021-03-12
fix(shape, xychart): default LinePath strokeLinecap to round

williaster:
#### :bug: Bug fix

Title says it all. This is a good default behavior in order to make sure that paths of single data points surrounded by `null` values are still visible for `Line` and `AreaSeries`.

~We could also consider setting this as the default in `@visx/shape`, but that's a very popular package so I worry about changing behavior~ 🤔 

EDIT: going to make this change in `@visx/shape` as well as `@visx/xychart` so that all packages / consumers get the fix.

**Before**

**After**

@kristw @etr2460 @hshoff

### DIFF
--- a/packages/visx-shape/src/shapes/LinePath.tsx
+++ b/packages/visx-shape/src/shapes/LinePath.tsx
@@ -38,6 +38,9 @@ export default function LinePath<Datum>({
       className={cx('visx-linepath', className)}
       d={path(data) || ''}
       fill={fill}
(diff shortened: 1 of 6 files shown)


kristw (approved):

williaster:
noting that I made the change in `@visx/shape`, but also still needed to make it in `@visx/xychart` for the animated components which ultimately use `<animated.path />`s

PR #1106, opened 2021-03-12
Export types in @visx/text

ConneXNL:
Currently @visx/text does not export its types. So the following does not work:
`import { TextProps } from '@visx/text';`

This PR adds the exporting of those types.

Ugly work-around until then:
```
import { Text } from '@visx/text';
Parameters<typeof Text>[0]>
```


williaster (approved):
Thanks for the addition @ConneXNL 🙏 Looks good to me, we'll get it merged and released by end of day.

FYI a _slightly_ less gross (but gross nonetheless) workaround for now is a deep import like `import { TextProps } from '@visx/text/lib/types';`

PR #1110, opened 2021-03-15
fix(brush/BrushHandle): prevent BrushHandle to trigger dragEnd onMouseLeave

Atralbus:
#### :bug: Bug Fix

- This PR tries to solve https://github.com/airbnb/visx/issues/1109

I'm not sure if this `onMouseLeave` event is crucial for some other use case, but it looks like this new behavior is a bit more similar to the behavior in `BaseBrush`.

### DIFF
--- a/packages/visx-brush/src/BrushHandle.tsx
+++ b/packages/visx-brush/src/BrushHandle.tsx
@@ -131,7 +131,6 @@ export default class BrushHandle extends React.Component<BrushHandleProps> {
                 style={{ cursor }}
                 onMouseMove={dragMove}
                 onMouseUp={dragEnd}
-                onMouseLeave={dragEnd}
               />
             )}
             <rect


williaster:
Thanks for the fix @Atralbus ! I think currently this might be considered a breaking change 💥 , which we'd like to avoid if possible. And as you said it's unclear if current lib consumers rely on this behavior.

I wonder if an alternative might be to **opt into** this feature at the `Brush` level, something like `ignoreResizeMouseOut` perhaps. Then users can do whatever makes sense for their specific application. wdyt?

williaster:
Actually in looking back at your original 🐛  report, I agree the UX as it exists now is pretty poor and ideally users wouldn't have to opt into something like this. 

I played a bit with this fix, tho, and while it fixes the original issue it's not ideal when you `mouseUp` **outside** of the handle as the brush stays in "edit" mode. It can then be pretty challenging to get it to `mouseUp` properly at the edge of the bounds.

Apr-12-2021 18-26-59

Seems like ideally we would detect the `mouseUp` event in that case and stop edit mode wherever the brush last was, but you'd probably need a window listener for that which will be more work to implement.

d3x42:
Oh, I see. I'll try to take a look at this case.

Atralbus:
@williaster

> I wonder if an alternative might be to **opt into** this feature at the `Brush` level, something like `ignoreResizeMouseOut` perhaps. Then users can do whatever makes sense for their specific application. wdyt?

I think it might be nice have an option to disable it.

> I played a bit with this fix, tho, and while it fixes the original issue it's not ideal when you mouseUp outside of the handle as the brush stays in "edit" mode. It can then be pretty challenging to get it to mouseUp properly at the edge of the bounds.

This problem also exists if we start brushing from an empty area(`Brush` component). That's why I thought it would be fine. But it would be nice to solve that.

> Seems like ideally we would detect the mouseUp event in that case and stop edit mode wherever the brush last was, but you'd probably need a window listener for that which will be more work to implement.

I didn't check the code in `d3` but they might be handling this brush events with window listener. It may also fix the fast drag problem I mentioned in the bug report.

d3x42:
@williaster @Atralbus Please check my PR for this issue: https://github.com/airbnb/visx/pull/1164

I suppose it covers most cases (for horizontal mode, didn't test it on vertical and did nothing for corners). But I could require some cleanup. I'll appreciate your review.

williaster:
@d3x42 is the idea that your PR would supersede this PR? it seems like you handle the `window` listeners which also handles avoiding calling `dragEnd` upon mouse leave. Just did a first review of that PR.

d3x42:
@williaster Yep, that what i planned

d3x42:
Could be closed. Fixed in [PR-1164](https://github.com/airbnb/visx/pull/1164)

williaster:
Superseded by #1164

PR #1116, opened 2021-03-18
fix(tooltip, annotation): bump react-use-measure to ^2.0.4. fixes #1091

### LINKED ISSUE #1091: [tooltip][annotation] update react-use-measure dep
There's an issue with `react-use-measure` causing problems with cjs exports as described here: https://github.com/grafana/grafana/issues/31598 and there's a PR to fix here: https://github.com/pmndrs/react-use-measure/pull/38

This issue is to track the `react-use-measure` change. If it blocks too long we should explore dropping the dep and going forward with an api compatible fix. 

cc @williaster

hshoff:
#### :bug: Bug Fix

- fix(tooltip, annotation): bump react-use-measure to ^2.0.4. fixes https://github.com/airbnb/visx/issues/1091

### DIFF
--- a/packages/visx-annotation/package.json
+++ b/packages/visx-annotation/package.json
@@ -40,7 +40,7 @@
     "@visx/text": "1.7.0",
     "classnames": "^2.2.5",
     "prop-types": "^15.5.10",
-    "react-use-measure": "^2.0.3"
+    "react-use-measure": "^2.0.4"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


williaster (approved):
Nice

williaster:
Heads up happo is using a stale baseline comparison commit  (i.e., not master) so there will likely be diffs. I'm working on it with the happo folks.

Issue #1047, opened 2021-02-04
[visx/visx] namespace all exports 


williaster:
#1043 name-spaced the `@visx/xychart` package from `@visx/visx` to an `XYChart` variable to avoid name conflicts between package exports.

For `2.0.0` we could make this consistent for all packages.

robsutcliffe:
@williaster I'd be willing to help as a beginner

To clarify: the plan is for [all packages exported here](https://github.com/airbnb/visx/blob/master/packages/visx-visx/src/index.ts) to be exported using the package name as a namespace. so for example Annotation would be imported like:
`import { Annotation } from '@visx/visx';`

and then used like:
`<Annotation.LineSubject />`

williaster:
hey @robsutcliffe 👋 , awesome! 

yes, you're correct about the plan and the API 💯  We did this initially with the `@visx/xychart` package, and as you can see in the `index` file you linked to we used `XYChartPackage` for the exported name. I think ideally we don't include `*Package` in the export name, your example of `Annotation` is could be used across all packages (e.g., `XYChartPackage` could be updated to `XYChart`).

robsutcliffe:
Great. Feel free to assign this to me and I'll add a PR soon

williaster:
closed by #1121

PR #1121, opened 2021-03-25
1047

robsutcliffe:
This change is based on issue [#1047](https://github.com/airbnb/visx/issues/1047) 

#### :boom: Breaking Changes

- Created a unique namespace for each package exported from @visx/visx that corresponds with the package name
- Updated unit test for @visx/visx to use the new namespaces

#### :rocket: Enhancements

- n/a

#### :memo: Documentation

- n/a

#### :bug: Bug Fix

- n/a

#### :house: Internal

- n/a

_*I couldn't find any documentation or examples that import anything from packages directly through @visx/visx but this is my first contribution to this project so please let me know if there is more that needs to be updates_ 

cc @williaster


packages/visx-visx/src/index.ts
```diff
@@ -1,33 +1,61 @@
...
-export * from '@visx/responsive';
-export * from '@visx/scale';
-export * from '@visx/shape';
-export * from '@visx/text';
-export * from '@visx/tooltip';
-export * from '@visx/voronoi';
-export * from '@visx/zoom';
-
-export const XYChart = XYChartPackage;
+export default {
```
williaster, line 31:
```suggestion
export {
```

I think we _don't_ want this to be the `default` export because it would prevent this style of importing:

```ts
// error
import { Shape } from '@visx/visx';

// would have to do
import visx from '@visx/visx';
const { Shape } = visx;
```

In the test file then you would need to fallback to importing everything via

```ts
import * as visx from '...';
```

packages/visx-visx/test/index.test.ts
```diff
@@ -1,113 +1,113 @@
...
-    expect(visx.Tooltip).toBeDefined();
+    expect(visx.Tooltip.Tooltip).toBeDefined();
   });
 
   it('should export @visx/voronoi', () => {
-    expect(visx.voronoi).toBeDefined();
+    expect(visx.Voronoi.voronoi).toBeDefined();
   });
 
   it('should export @visx/xychart', () => {
```
williaster, line 113: I can't comment on it because it's not part of the diff, but I think you missed one below

```tsx
// current
expect(visx.Zoom).toBeDefined();

// idea
expect(visx.Zoom.Zoom).toBeDefined();
```

williaster (commented):
Thanks for the contribution @robsutcliffe! Sorry for the delayed review, this looks great to me overall! I had two minor comments.

I'll figure out when we want to roll this out, essentially think about whether we want to bundle any other breaking changes in the `2.0` release or not.

williaster:
@robsutcliffe we're going to try to land this this week for the `2.0.0` release. lmk if you have interest in the suggested fixes above^ else I can take over.

thanks again for the addition, sorry it's taken bit to land.

robsutcliffe:
Sorry @williaster I didn't see your earlier comments. Those changes are in now and I've merged all the master changes too to make sure no conflicts have come up since this request was opened.

williaster (approved):
Amazing! Thanks for rebasing and updating @robsutcliffe 🙏  Fixing a release issue then will get this in 🎉

packages/visx-visx/test/index.test.ts
```diff
@@ -1,120 +1,120 @@
 /* eslint-disable import/namespace */
-import * as visx from '../src';
+import visx from '../src';
```
williaster, line 2: ah I think this is failing the build since there is no default export

```suggestion
import * as visx from '../src';
```

Issue #1123, opened 2021-03-27
[visx/text] Word-wrapping requires fontSize in style property


alexa-infra:
In the way `getStringWidth()` calculates width right now:

    Object.assign(textEl.style, style);
    textEl.textContent = str;
    return textEl.getComputedTextLength();

There are the following problems: 1. it doesn't take into account `fontSize` of `Text` component, 2. if `style` is not set, then it uses document font size, 3. or it uses `style` of previously rendered `Text` component (because temporary element is created once and `style` is not cleared after a call), 4. if `fontSize` is defined via classes, then it's not counted too, 5. and basically the same is true for all `font-` defining properties

To reproduce the problem:

      <Text x={180} y={180} width={360} textAnchor="middle" verticalAnchor="middle"
            fontSize="60pt">We don't need no education</Text>

![Screenshot from 2021-03-27 21-45-28](https://user-images.githubusercontent.com/464663/112734405-cecd2b80-8f45-11eb-8fff-8a8f73572c9f.png)

and workaround to fix that:

      <Text x={180} y={180} width={360} textAnchor="middle" verticalAnchor="middle"
            fontSize="60pt" style={{fontSize: '60pt'}}>We don't need no education</Text>

![Screenshot from 2021-03-27 21-44-37](https://user-images.githubusercontent.com/464663/112734419-e3a9bf00-8f45-11eb-9c68-ad478eabb668.png)

williaster:
Hey @alexa-infra 👋 , thanks for the bug report I agree there are some explicit issues, but some of the limitations you describe are pretty challenging to fix (happy to hear any ideas you or others have!). Seems like ideally we would

**Re 1 & 5** We should probably pull out some of the common `font` properties from `Text` and add them to `style`, allowing users to override whatever they want in their own `style` object. These should probably include
  - `fontSize`
  - `fontWeight`
  - `fontFamily`

**Re 3**
We should definitely clear the `style` property of the cached DOM element between renders so we don't get bleed-over. I think we still want to avoid recreating the node each invocation, though.

**Re 2 & 4**
If styles are defined in classes via `css-in-jss`, etc. I'm not sure there's much we can do. There are a million different frameworks for that, so `visx` generally has taken the approach of not assuming any of them, which often means falling back to old-school `style` objects + `className`s.

I don't have the bandwidth to make these fixes but would happily review a PR if someone else wants to take a shot 🙏

dreamyguy:
[...]

PR #1127, opened 2021-03-29
fix(visx/annotation) Improve label positioning, add maxWidth

vovakulikov:
#### :boom: Breaking Changes
- Possible breaking change (remove default value of width at label component from @visx/annotation)

#### :bug: Bug Fix
- Fixes https://github.com/airbnb/visx/issues/1126

This PR is trying to improve horizontal label positioning at @visx/annotation package.  

Info from https://github.com/airbnb/visx/issues/1126

> Label component from @visx/annotation is trying to calculate right position for text within component according x,y,dx and dy props. But this component does’t take into account real width of text element

TLDR; This fix has added real width measures by `useMeasures` hook and now label component does calculation according real width and paddings value as well.
**Examples**

Problem - Bad label positioning example
https://codesandbox.io/s/bold-antonelli-0v17h?file=/src/pie-chart.tsx
GitHub repo branch https://github.com/vovakulikov/chart-grid-sandbox/tree/bad-label-positioning-example

**Example with fix** label position  https://codesandbox.io/s/focused-https-mvvjn
Github repo branch https://github.com/vovakulikov/chart-grid-sandbox/tree/label-positioning-with-annotation-fix

### DIFF
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -101,10 +101,8 @@ export default function Label({
     padding.top + padding.bottom + (titleBounds.height ?? 0) + (subtitleBounds.height ?? 0),
   );
 
-  const measuredWidth = padding.right + padding.left + Math.max(
-    titleBounds.width ?? 0 ,
-    subtitleBounds.width ?? 0
-  );
+  const measuredWidth =
+    padding.right + padding.left + Math.max(titleBounds.width ?? 0, subtitleBounds.width ?? 0);
   const width = propWidth ?? measuredWidth;
   const innerWidth = (width ?? measuredWidth) - padding.left - padding.right;
 
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -128,16 +128,16 @@ export default function Label({
   });
 
   const titleMeasuredWidth = titleWordsByLine.reduce(
-    (maxWidth, line) => Math.max(maxWidth, line.width ?? 0),
+    (maxTitleWidth, line) => Math.max(maxTitleWidth, line.width ?? 0),
     0,
   );
 
   const subtitleMeasuredWidth = subtitleWordsByLine.reduce(
-    (maxWidth, line) => Math.max(maxWidth, line.width ?? 0),
+    (maxSubtitleWidth, line) => Math.max(maxSubtitleWidth, line.width ?? 0),
     0,
   );
 
-  const textMeasuredWidth = Math.max(titleMeasuredWidth, subtitleMeasuredWidth);
+  const textMeasuredWidth = Math.floor(Math.min(maxWidth, Math.max(titleMeasuredWidth, subtitleMeasuredWidth)));
   const measuredWidth = padding.right + padding.left + textMeasuredWidth;
   const width = propWidth ?? measuredWidth;
   const innerWidth = width - padding.left - padding.right;
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -137,7 +137,9 @@ export default function Label({
     0,
   );
 
-  const textMeasuredWidth = Math.floor(Math.min(maxWidth, Math.max(titleMeasuredWidth, subtitleMeasuredWidth)));
+  const textMeasuredWidth = Math.floor(
+    Math.min(maxWidth, Math.max(titleMeasuredWidth, subtitleMeasuredWidth)),
+  );
   const measuredWidth = padding.right + padding.left + textMeasuredWidth;
   const width = propWidth ?? measuredWidth;
   const innerWidth = width - padding.left - padding.right;
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -89,7 +89,7 @@ export default function Label({
   titleProps,
   verticalAnchor: propsVerticalAnchor,
   width: propWidth,
-  maxWidth = 120,
+  maxWidth = 125,
   x: propsX,
   y: propsY,
 }: LabelProps) {
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -2,6 +2,7 @@ import React, { useContext, useMemo } from 'react';
 import cx from 'classnames';
 import Group from '@visx/group/lib/Group';
 import Text, { TextProps } from '@visx/text/lib/Text';
+import { useText } from '@visx/text';
 import useMeasure, { Options as UseMeasureOptions } from 'react-use-measure';
 import AnnotationContext from '../context/AnnotationContext';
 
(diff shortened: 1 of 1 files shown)


packages/visx-annotation/src/components/Label.tsx
```diff
@@ -85,29 +85,35 @@ export default function Label({
   titleFontWeight = 600,
   titleProps,
   verticalAnchor: propsVerticalAnchor,
-  width = 125,
+  width: propWidth,
```
vovakulikov, line 88: Possible breaking change.

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -85,29 +85,35 @@ export default function Label({
...
     padding.top + padding.bottom + (titleBounds.height ?? 0) + (subtitleBounds.height ?? 0),
   );
-  const innerWidth = width - padding.left - padding.right;
+
+  const measuredWidth = padding.right + padding.left + Math.max(
+    titleBounds.width ?? 0 ,
+    subtitleBounds.width ?? 0
+  );
+  const width = propWidth ?? measuredWidth;
+  const innerWidth = (width ?? measuredWidth) - padding.left - padding.right;
```
vovakulikov, line 109: Fallback on real width of label in case if consumer did't pass label width

vovakulikov:
@williaster can you help me with Happo UI tests. I don't have access to these tests therefore can't see what is going on there. Maybe I can fix that. Thanks.

williaster (commented):
Hey @vovakulikov sorry for the delayed review, have been super busy. Thanks for another contribution! Happo has been a little flaky so don't worry about that / I just approved the diff. We're working on the issue, it basically has a stale base ref that it's comparing to for visual differences.

I think this approach works well for single-word labels as in your example. Otherwise there's a chicken-egg problem:
- the measured `width` is determined from whatever the `<Text />` component renders
- the `<Text />` component size is determined by `width` it is passed

So I think what happens if no `width` is specified initially
1) `measuredWidth = padding.left + padding.right ~= 24` (since measured = 0)
2) `measuredWidth=24` is passed to `Text`, which tries to render all of the text inside a `width` of `24`. `Text` won't break words, so a single word will render fine.
3) after `Text` is mounted/rendered, `measuredWidth` might update to include the new `measuredWidth` for whatever the (longest) word is

I pulled the branch to test it locally (the annotation example for happo specifies a width, so it wouldn't be affected by this change). This is what I get if I remove the `width`

image

If I dynamically update the text, you can see that the width updates to the longest word as described

Apr-08-2021 11-02-33

I think you **can** actually get the measured width to work *if you pass `undefined`* to `Text`, then it won't try to wrap based on a (non-existent) `width`. So in the case that `propsWidth` is undefined, I think we'd just pass `undefined` to the title + subtitle `Text` (and can probably remove all of the `measuredWidth` calculations since it'll be unused).

We've been trying to avoid breaking changes, I'll try to think on that a little bit. One thought I had was that even if you don't want to specify an explicit `width`, you might want auto-sizing with a `maxWidth`. So I wonder if we could do something like the following:
1) leave `width` as is, remove the default
2) add a `maxWidth` prop defaulted to the previous `width` default of `125`.

vovakulikov:
@williaster Thanks for really comprehensive explanation about what is going on with width. 

It's getting start a really interesting bug because my example actually has a double-word title. I've checked my example again and added some suffixes just to check how it works and... 

Screenshot 2021-04-10 at 17 06 33

*(Screenshot from origin bug example with title suffix, **Notice** width for annotation was calculated according to length of title)*

So initial value for title was **calculated right for first render** but then we passed this first-calculated `measuredWidth` to `<Text/>` and by that we kind of lost our ability to measure width in a future because no matter what length of content we have, we have already passed width from first render to text. 

Actually this is almost the same that you've described above), So I totally agree to remove lines about passing width to `<Text />`.  

Screenshot 2021-04-10 at 17 29 44

*(Example without passing width to `<Text />`. Notice that we don't have word wrapping at all)*

But I can't figure out how to add this `maxWidth` prop to avoid this long text problem without using `useText` hook. It seems like we have to know some value of width before passing this value further to `<Text />`.

So as a result we have this visual behaviour

Screenshot 2021-04-10 at 18 34 48

*Labels with default `maxWidth` value 120px*

And gif about how it works at gallery example 
Screen Recording 2021-04-10 at 18 48 49

*Example from visx demo package, Notice in this example we have a small jump while text updating is going. Need investigate this further, right now don't have clear understanding why this is happening*

**A few words about this**
>and can probably remove all of the measuredWidth calculations since it'll be unused.

I've tried to remove this `measuredWidth` but it seems like we still need to have a width value since we use this to calculate adjusted coordinate for the container.

https://github.com/airbnb/visx/blob/42d3213fe5286388fc456eade9b3874d096616aa/packages/visx-annotation/src/components/Label.tsx#L111-L121

vovakulikov:
Also https://github.com/airbnb/visx/issues/1123 probably related to useText calculation problem

williaster:
Hey @vovakulikov thanks so much for the thorough follow up! I didn't quite have time to get to this today but will review again tomorrow 🙏

felixfbecker:
@williaster reading the discussion/problem statements here, I wonder if using `<foreignObject>` with HTML elements would help/simplify in some way? Could that offer automatic text reflow?

williaster:
@felixfbecker that's an interesting idea, we haven't really used that in `visx` at all. For `@visx/legend` and `@visx/tooltip` we opted to make them HTML based to get layout for free, and we have `@visx/text` for cases where you really probably should use SVG `<text />`s.

Annotations were initially implemented with SVG to get something out and usable, but we discussed having an HTML based variant as well at some point. `<foreignObject>` could be interesting to consider instead, though. maybe we can create a new issue to track that, since I think this change is still a strict win 🏆 .

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -100,6 +104,44 @@ export default function Label({
...
+  const subtitleMeasuredWidth = subtitleWordsByLine.reduce(
+    (maxSubtitleWidth, line) => Math.max(maxSubtitleWidth, line.width ?? 0),
+    0,
+  );
+
+  const textMeasuredWidth = Math.floor(
+    Math.min(maxWidth, Math.max(titleMeasuredWidth, subtitleMeasuredWidth)),
+  );
+  const measuredWidth = padding.right + padding.left + textMeasuredWidth;
+  const width = propWidth ?? measuredWidth;
```
williaster, line 144: it's a bummer we can't skip the calculations above if `propWidth` **is** passed, but not easy to do with `hooks`. Theoretically could create a wrapper component explicitly to provide `width` if it's not passed, but I think we could consider that a future improvement if there are any perf issues. I didn't notice any lag in the annotation demo when quickly updating the title/subtitle text.

williaster (approved):
This looks great @vovakulikov, thanks again for the fix and addition of `maxWidth`! 

I pulled it locally to test it out as well and it seems pretty solid. I agree the edge cases in calculations likely come from the issue reported in #1123.

packages/visx-annotation/src/components/Label.tsx
vovakulikov: Totally with you about this. I like this idea about component wrapper with measure logic within.

PR #1155, opened 2021-04-10
fix(brush): use PointerEvents instead of MouseEvents 

RemaaBdair:
I opened this pr with the same changes as in this one https://github.com/airbnb/visx/pull/864, but with addressing the reviews as well in order to speed things up.
#### :boom: Breaking Changes

- none

#### :rocket: Enhancements

-

#### :memo: Documentation

-

#### :bug: Bug Fix

- #845 

#### :house: Internal

-

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -10,9 +10,10 @@ import { MarginShape, Point, BrushShape, ResizeTriggerAreas, PartialBrushStartEn
 
 const BRUSH_OVERLAY_STYLES = { cursor: 'crosshair' };
 
-type MouseHandlerEvent =
+type PointerHandlerEvent =
   | React.MouseEvent<SVGRectElement, MouseEvent>
-  | React.TouchEvent<SVGRectElement>;
+  | React.TouchEvent<SVGRectElement>
+  | React.PointerEvent<SVGRectElement>;
 
 export type BaseBrushProps = {
   brushDirection?: 'horizontal' | 'vertical' | 'both';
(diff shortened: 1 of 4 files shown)


packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -10,9 +10,10 @@ import { MarginShape, Point, BrushShape, ResizeTriggerAreas, PartialBrushStartEn
 
 const BRUSH_OVERLAY_STYLES = { cursor: 'crosshair' };
 
-type MouseHandlerEvent =
+type PointerHandlerEvent =
```
williaster, line 13: I think all of this can be simplified to `React.PointerEvent<SVGRectElement>`, which is already equivalent to `MouseEvent | TouchEvent`

packages/visx-brush/src/BrushSelection.tsx
```diff
@@ -5,8 +5,8 @@ import { BaseBrushState as BrushState, UpdateBrush } from './BaseBrush';
 
 const DRAGGING_OVERLAY_STYLES = { cursor: 'move' };
 
-type MouseHandler = (
-  event: React.MouseEvent<SVGRectElement, MouseEvent> | React.TouchEvent<SVGRectElement>,
+type PointerHandler = (
+  event: React.MouseEvent<SVGRectElement, MouseEvent> | React.TouchEvent<SVGRectElement> | React.PointerEvent<SVGRectElement>,
```
williaster, line 9: same comment here

```suggestion
  event: React.PointerEvent<SVGRectElement>,
```

williaster (commented):
Thanks for starting a new PR for this long-standing fix @RemaaBdair ! 🙏 

I had one suggestion to simplify but otherwise looks good to me. Note, too, that it is currently failing our `prettier` linting. I think after my suggested changes it may want the type definitions to each be on a single line. You should be able to run `yarn format` in your `visx` root directory to fix it after the suggested changes, and then commit any changes that come from running that.

williaster (approved):
Amazing, thanks for the fix @RemaaBdair !

PR #1156, opened 2021-04-11
new(shape/Polygon): add optional 'points' override

jakeisnt:
#### :rocket: Enhancements

- The Polygon component can now accept `points` as a field to override its functionality for generating the points based on the `size`, `sides` and `center` props. This allows users of the library to leverage the underlying `polygon`'s flexibility while still making use of visx's styles and utilities.

Note: If a `points` prop is provided to a Polygon component, it's now used by the Polygon rather than being passed to the children.

#### :memo: Documentation

- I've introduced an additional prop to the `Polygon` component and provided a comment as documentation for it.

### DIFF
--- a/packages/visx-shape/src/shapes/Polygon.tsx
+++ b/packages/visx-shape/src/shapes/Polygon.tsx
@@ -6,8 +6,8 @@ import { AddSVGProps } from '../types';
 const DEFAULT_CENTER = { x: 0, y: 0 };
 
 export const getPoint = ({
-  sides,
-  size,
+  sides = 4,
+  size = 25,
   center = DEFAULT_CENTER,
   rotate = 0,
   side,
@@ -26,7 +26,7 @@ export const getPoints = ({
   size,
   center,
   rotate,
-}: Pick<PolygonProps, 'sides' | 'size' | 'center' | 'rotate'>) =>
+}: NonNullable<Pick<PolygonProps, 'sides' | 'size' | 'center' | 'rotate'>>) =>
   new Array(sides).fill(0).map((_, side) =>
     getPoint({
       sides,
@@ -78,7 +78,7 @@ export default function Polygon({
   }).map(({ x, y }) => [x, y]);
 
   // eslint-disable-next-line react/jsx-no-useless-fragment
-  if (children) return <>{children({ points })}</>;
+  if (children) return <>{children({ points: pointsToRender })}</>;
 
   return (
     <polygon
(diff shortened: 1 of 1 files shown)


packages/visx-shape/src/shapes/Polygon.tsx
```diff
@@ -58,16 +60,17 @@ export type PolygonProps = {
...
 export default function Polygon({
-  sides,
+  sides = 4,
   size = 25,
   center = DEFAULT_CENTER,
   rotate = 0,
   className,
   children,
   innerRef,
+  points = null,
```
williaster, line 70: Removing the `null` because of the TS error

`Type 'null' is not assignable to type '[number, number][]'`

```suggestion
  points,
```

packages/visx-shape/src/shapes/Polygon.tsx
```diff
@@ -39,10 +39,12 @@ export const getPoints = ({
 
 export type PolygonProps = {
   /** Number of polygon sides. */
-  sides: number;
+  sides?: number;
```
williaster, line 42: Above we are getting TS errors because `sides` and `size` are now optional in `PolygonProps`. It may work if you update the `getPoint` signature type to something like this 

```tsx
... & NonNullable<Pick<PolygonProps, 'sides' | 'size' | 'center' | 'rotate'>>
```

otherwise you could just define them as numbers and not try to `Pick` them

packages/visx-shape/src/shapes/Polygon.tsx
```diff
@@ -39,10 +39,12 @@ export const getPoints = ({
...
   /** Size of the shape. */
-  size: number;
+  size?: number;
   /** className to apply to polygon element. */
+  /** Points to use to render the polygon. If this is defined, `sides`, `size`, `rotate`, and `center` are ignored. */
```
williaster, line 46: looks like this new prop + doc string got inserted between the `className` doc string and the `className` prop, can you co-locate those again?

williaster (commented):
Hey @jakeisnt thanks for the contribution! 🙏  This seems like a useful override. 

There are a few issues with the resulting TypeScript which shouldn't be too bad to fix.

packages/visx-shape/src/shapes/Polygon.tsx
jakeisnt: Yes!

packages/visx-shape/src/shapes/Polygon.tsx
jakeisnt: This worked locally, thanks!

jakeisnt:
Should be good to go! Let me know if you'd like me to restructure the commits.

williaster (approved):
lgtm, thanks for the TS fixes! 

Note I removed the comment about a `breaking change`, and moved that note under `enhancements`. I wouldn't consider this breaking since the `points` prop is new / shouldn't affect current usage.

Issue #1109, opened 2021-03-15
[brush] Drag ends when mouse leaves the BrushHandle


Atralbus:
When we create a new brush `rect`, we can move the mouse outside the brush area while dragging. If we don't trigger `mouseup` we can move the mouse back in to the brush area to continue dragging. But when we trigger a drag from `BrushHandle` and move mouse outside the brush area it immediately triggers `dragEnd` which prevents us to continue brushing by moving mouse inside the brush area.

1. Go to https://airbnb.io/visx/brush.
2. Drag the brush from left or right handle and move mouse outside the brush area.
3. Move back mouse inside the brush area while holding left mouse button.
4. See that brush no longer moves.

I think we should be able to move brush even if the mouse leaves the brush area or handle.

Correct behavior in `BaseBrush`
![Screen Recording 2021-03-15 at 15 40 35](https://user-images.githubusercontent.com/4368872/111155433-82790780-85a5-11eb-94b5-9b45ee9ded7a.gif)

The problem with `BrushHandle`
![Screen Recording 2021-03-15 at 15 40 20](https://user-images.githubusercontent.com/4368872/111155441-84db6180-85a5-11eb-805c-f00d9ff04fff.gif)

This is a minor issue with the user experience of `Brush` component. But there seems to be a bigger problem. Currently, the component cannot catch up with fast mouse move events, if you move mouse outside the brush area too fast, it will leave the brush a bit far from the edge of the brush area. If you check this d3 [demo](https://bl.ocks.org/bumbeishvili/6c54d3f0e202aa7004a669a768369c5d) you can see that no matter how fast you move mouse outside, it stops on the edge of the brush area. This greatly improves the user experience with the component. Here's the fast drag issue;

![Screen Recording 2021-03-15 at 16 06 43](https://user-images.githubusercontent.com/4368872/111158008-9b36ec80-85a8-11eb-815b-fe4e133c3447.gif)

Thank you for this great library. It makes it a lot easier to create complex charts in React. Hoping to see more.

williaster:
closed by #1164

PR #1164, opened 2021-04-14
feat(brush): add using widows move events with brush

d3x42:
The new param isUseWindowMoveEvents will allow window event listeners to control dragging move and end. 

#### :rocket: Enhancements

It's possible to move the cursor away from the brush area (and finish brushing there).

Related PR and issues:
PR:  [PR-1110](https://github.com/airbnb/visx/pull/1110) 
Issue: [#1109](https://github.com/airbnb/visx/issues/1109)

My goal here - in the "isUseWindowMoveEvents" mode handle all mouse movements and mouse up events on the Base Brush component level. If there are examples for BrushCorner in the gallery I could try to handle it too.


d3x42:
Oh, it seems that my linter didn't get all issues here. I'll provide lint fixes tomorrow.

d3x42:
@williaster Ready for review. Thanks!

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -33,11 +34,17 @@ export type BaseBrushProps = {
   clickSensitivity: number;
   disableDraggingSelection: boolean;
   resetOnEnd?: boolean;
+  isUseWindowMoveEvents?: boolean;
 };
 
 export type BaseBrushState = BrushShape & {
   activeHandle: ResizeTriggerAreas | null;
   isBrushing: boolean;
+  isDragInProgress: boolean;
```
williaster, line 43: would it be possible to use the existing `isBrushing` flag instead of introducing `isDragInProgress` and `isMovingBrushSelection`? it's not clear readability-wise why we have 3 (will keep reading but that's my initial impression)

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +127,145 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     }
   }
 
+  componentDidMount() {
+    window.addEventListener('mouseup', this.handleMouseUp, false);
```
williaster, line 131: should the listeners be added / removed conditional on whether `props.useWindowMoveEvents === true`?

packages/visx-brush/src/Brush.tsx
```diff
@@ -58,6 +58,8 @@ export type BrushProps = {
   handleSize: number;
   /** Reference to the BaseBrush component. */
   innerRef?: React.MutableRefObject<BaseBrush | null>;
+  /** Prevent drag end on mouse leave from brush */
+  isUseWindowMoveEvents?: boolean;
```
williaster, line 62: same comment throughout

```suggestion
  useWindowMoveEvents?: boolean;
```

packages/visx-brush/src/BrushHandle.tsx
```diff
@@ -12,12 +12,23 @@ export type BrushHandleProps = {
   onBrushEnd?: (brush: BrushState) => void;
   type: ResizeTriggerAreas;
   handle: { x: number; y: number; width: number; height: number };
+  isUseWindowMoveEvents?: boolean;
```
williaster, line 15: Same suggestion here 
```suggestion
  useWindowMoveEvents?: boolean;
```

packages/visx-brush/src/BrushSelection.tsx
```diff
@@ -14,13 +14,16 @@ export type BrushSelectionProps = {
...
   updateBrush: (update: UpdateBrush) => void;
+  onMoveSelectionChange?: (value: boolean) => void;
   onBrushEnd?: (brush: BrushState) => void;
   disableDraggingSelection: boolean;
   onMouseLeave: PointerHandler;
   onMouseMove: PointerHandler;
   onMouseUp: PointerHandler;
   onClick: PointerHandler;
   selectedBoxStyle: React.SVGProps<SVGRectElement>;
+  isUseWindowMoveEvents?: boolean;
```
williaster, line 25:
```suggestion
  useWindowMoveEvents?: boolean;
```

packages/visx-drag/src/useDrag.ts
```diff
@@ -26,6 +26,8 @@ export type UseDragOptions = {
   dx?: number;
   /** Optionally set the initial drag dy, or override the current drag dy. */
   dy?: number;
+  /** If defined - parent controls dragging state  */
```
williaster, line 29:
```suggestion
  /** If defined, parent controls dragging state. */
```

packages/visx-drag/src/Drag.tsx
```diff
@@ -13,6 +13,8 @@ export type DragProps = UseDragOptions & {
   height: number;
   /** Whether to render an invisible rect below children to capture the drag area as defined by width and height. */
   captureDragArea?: boolean;
+  /** move drag state control to parent */
```
williaster, line 16:
```suggestion
  /** If defined, parent controls dragging state. */

```

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -442,8 +632,13 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
                     stageWidth={stageWidth}
                     stageHeight={stageHeight}
                     updateBrush={this.updateBrush}
-                    brush={this.state}
+                    brush={{ ...this.state }}
```
williaster, line 635: why does this need to be copied now?

williaster (commented):
Hey @d3x42 thanks for getting this started! 

I had a few specific comments to start, and my main high-level question is whether we need to introduce `isMovingBrushSelection` and `brushHandleChange`, or whether we can simply use the existing `isBrushing` flag. If you could explain this design choice more it'd be helpful.

packages/visx-brush/src/BaseBrush.tsx
d3x42: I'll roll back it. It was added during  searching for some bug in this PR

d3x42:
> Hey @d3x42 thanks for getting this started!
> 
> I had a few specific comments to start, and my main high-level question is whether we need to introduce `isMovingBrushSelection` and `brushHandleChange`, or whether we can simply use the existing `isBrushing` flag. If you could explain this design choice more it'd be helpful.

It was separated because I need to determine which event would be handled by the mouse move event listener (logic from move handlers in the child components slightly different). I think that I could use one enum variable instead of multiple boolean. Something like brushingType: Moving|Brushing|Resize_left|Resize_Right etc.

packages/visx-brush/src/BaseBrush.tsx
d3x42: Next week I'll try solution with one enum variable

d3x42: Now it should be much easier to understand. I replaced all new boolean variables with bushing type which help to understand what kind of action in progress. It helps me in moveMove event listener to understand how the move should be handled (for example it slightly different for moving brushing area and selecting a new one).

d3x42:
@williaster It seems that I resolved all of your comments. Please check when you have time.

williaster:
Hey @d3x42 sorry for such a long delay, have been super slammed at work and out a bit on vacation. Will try to get this reviewed by end of week 🙏

williaster:
Hey @d3x42 thanks for the ping. I just pulled this branch locally to play with it functionally (by setting `useWindowMoveEvents=true` on the `<Brush />` in the demo package) and there seem to be several issues still.

**Cannot resize the brush**
no-resize

**Jumpy-ness when moving the brush on the track**
It always jumps to the right side of the brush when you first move it
jumpy

**Jumpy-ness when moving the brush outside the track**
jumpy-window

How were you testing this functionally? I think these will need to be fixed before we can merge it, but thanks again for all the work on improving this 🙏

d3x42:
It's strange. Resize works for me:
resize

d3x42:
@williaster Sorry for the bug with jumping on move starts, it should be fine now. 

Resize works fine for me. I checked it in chrome, firefox and safari on macOS. Could you add some details on your os, browser, and browser version to reproduce it?

williaster:
Hey @d3x42 I pulled it again and it's no longer jumpy/works well when dragging outside of the svg 🎉.

Still can't resize tho, I tried it on chrome/firefox/safari and same behavior for each. I'm on `macOS BigSur 11.3.1`.  I wanted to confirm that you were rebuilding both `@visx/brush` and `@visx/drag` locally when testing? It's a little confusing because the `@visx/demo` site seems to use the `esm` build for direct imports (`brush`) and non-`esm` builds for dependencies of those imports (`drag`).

So locally I was running `yarn dev` in the demo package, then rebuilding with two commands
`yarn build:workspaces --workspaces=@visx/brush --esm`
`yarn build:workspaces --workspaces=@visx/drag`

To verify, when `yarn dev` is running for `@visx/demo` I usually check that the next server re-compiles when running either of those.

d3x42:
@williaster Thanks! After rebuilding @visx/drag without esm I was able to reproduce the issue. Fix already here.

williaster:
Hey @d3x42 I can resize now, but noticed a jumping issue during the transition between the `svg`/`window`. I wonder if any of the learnings from https://github.com/airbnb/visx/pull/1180 would be useful for this

May-17-2021 12-16-27

d3x42:
@williaster I fixed jumpings on resizing and selection

d3x42:
@weibohe Hi! Any news on the review for this PR?

d3x42:
Merge master changes to branch. 

@williaster could you please take a look on this pr?

d3x42:
@williaster gentle ping

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +125,148 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+                    ? Math.min(Math.max(y1 + offsetY, prevBrush.bounds.y0), prevBrush.bounds.y1)
+                    : y1,
+              },
+            ),
+          },
+        };
+      });
+    }
+
+    if (['move'].includes(brushingType ?? '')) {
```
williaster, line 214: Could simplify a bit/avoid instantiating arrays

```suggestion
    if (brushingType === 'move') {
```

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +125,148 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+            x0: x0 + validDx,
+            y0: y0 + validDy,
+            x1: x1 + validDx,
+            y1: y1 + validDy,
+          },
+        };
+      });
+    }
+
+    if (['select'].includes(brushingType ?? '')) {
```
williaster, line 242: Could simplify a bit/avoid instantiating arrays

```suggestion
    if (brushingType === 'select') {
```

packages/visx-brush/src/Brush.tsx
```diff
@@ -58,6 +58,8 @@ export type BrushProps = {
   handleSize: number;
   /** Reference to the BaseBrush component. */
   innerRef?: React.MutableRefObject<BaseBrush | null>;
+  /** Prevent drag end on mouse leave from brush */
```
williaster, line 61:
```suggestion
  /** Prevent drag end on mouse leaving from brush stage. */
```

packages/visx-brush/src/BrushHandle.tsx
```diff
@@ -12,13 +13,24 @@ export type BrushHandleProps = {
   onBrushEnd?: (brush: BrushState) => void;
   type: ResizeTriggerAreas;
   handle: { x: number; y: number; width: number; height: number };
+  useWindowMoveEvents?: boolean;
```
williaster, line 16: similarly for this would it make sense to be called `isControlled`?

williaster (commented):
Hey @d3x42 sorry for the delay on my end, I have been super busy but I appreciate the pings. I think this is close, just a couple more comments on my end for readability.

williaster:
don't worry about happo failing, we are having some auth issues with it.

williaster:
also played with it functionally and looks good, do you want to set `useWindowEvents` on the `@visx/demo` example so this is the default behavior there? 😄

packages/visx-brush/src/BrushSelection.tsx
```diff
@@ -14,13 +17,16 @@ export type BrushSelectionProps = {
...
   updateBrush: (update: UpdateBrush) => void;
+  onMoveSelectionChange?: (type?: BrushingType, options?: BrushingOptions) => void;
   onBrushEnd?: (brush: BrushState) => void;
   disableDraggingSelection: boolean;
   onMouseLeave: PointerHandler;
   onMouseMove: PointerHandler;
   onMouseUp: PointerHandler;
   onClick: PointerHandler;
   selectedBoxStyle: React.SVGProps<SVGRectElement>;
+  useWindowMoveEvents?: boolean;
```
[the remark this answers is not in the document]
d3x42: It makes sense. I'll rename it

d3x42:
> also played with it functionally and looks good, do you want to set `useWindowEvents` on the `@visx/demo` example so this is the default behavior there? 😄

Sure! I'll update the demo. :)

d3x42:
@d3x42 PR updated with your suggestions.) "isControlled" definitely way better for child components in Brush.

packages/visx-brush/src/types.ts
```diff
@@ -44,6 +44,12 @@ export type ResizeTriggerAreas =
   | 'bottomLeft'
   | 'bottomRight';
 
+export type BrushingType = 'move' | 'select' | ResizeTriggerAreas;
+export type BrushingOptions = {
```
williaster, line 48: `BrushingOptions` is pretty generic and so is a little hard to understand what it means. Could we update to something like `BrushPageOffset`?

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +125,148 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+  }
+
+  componentWillUnmount() {
+    if (this.props.useWindowMoveEvents) {
+      window.removeEventListener('mouseup', this.handleMouseUp);
+      window.removeEventListener('mousemove', this.handleMouseMove);
+    }
+  }
+
+  handleMouseUp = () => {
```
williaster, line 142: could we make it clear that these are window handlers? (and could generalize to pointer events)

```suggestion
  handleWindowPointerUp = () => {
```

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +125,148 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+          ...prevBrush,
+          activeHandle: null,
+          isBrushing: false,
+          brushingType: undefined,
+        };
+      });
+    }
+  };
+
+  handleMouseMove = (event: MouseEvent) => {
```
williaster, line 168: similar comment here about scoping to window handlers, and I think ideally we'd use `MouseTouchOrPointerEvent` (see example of this type being imported from `@visx/drag` in the brush utils file).  Being careful about mouse-specifics because of #845. 

```suggestion
  handleWindowPointerMove = (event: MouseTouchOrPointerEvent) => {
```

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -111,6 +125,148 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
...
+    const { useWindowMoveEvents, onBrushEnd, resetOnEnd } = this.props;
+    const { brushingType, isBrushing, brushingOptions, start } = this.state;
+
+    if (!useWindowMoveEvents || !isBrushing) return;
+
+    const offsetX = event.pageX - (brushingOptions?.pageX || 0);
```
williaster, line 174: just for readability/for future devs, could we add a comment for why these offsets are needed?

williaster (commented):
Okay I think this is probably the final review. Sorry for multiple review rounds but this one is definitely more complex than some.

Thank you for all the hard work on this, it's a significant improvement to the current UX! 👏  I think after these changes we can merge this puppy! 🚢

d3x42:
@williaster PR updated. Multiple reviews rounds fine for me, they really make this PR better.)

williaster (approved):
Sweet, lgtm! 😍  Thank you again 🎉

sakulstra:
Hey, i saw this pr superseded #864 , but i'm pretty sure it doesn't.
as far as i know  `mouseMove` will not fire on `touchMove` so on touch devices the brush is still not usable.
I think you can verify by visiting: https://airbnb.io/visx/brush

williaster:
@sakulstra that's correct, this is being tracked in #845 . Fix is out now in #1286 .

PR #1175, opened 2021-04-18
fix(@visx/text) Bad size measurements in Firefox

vovakulikov:
#### :bug: Bug Fix
- Fixes: https://github.com/airbnb/visx/issues/1111

I don't have a clear understanding of this but it seems like `react-use-measure` hook doesn't work well with inner `svg` element within <Text />. 

I consider this problem came from how Firefox treats empty svg elements without width and height within another svg element with height and width - They usually have `width=100%` `height=100%` of parent size even if we put size value from CSS `width: 0px; height: 0px;` explicitly.

In the first render we're getting a really big value of Text svg height, so we're just stuck with this big initial height value here and therefore have this visual behaviour here.

Possible solution in this PR this is replace `innerRef` target from svg `<Text />` root element to `<text />` element which can be measured without this problem.

Only one concern here this is kind of breaking change in public API types (before we had i`nnerRef: Ref<SVGSvgElement>`, now we have `Ref<SVGTextElement>`). I see here two option how to avoid breaking change in types 
- Leave public types API as it is and add casting to text ref element `<text ref={innerRef as Ref<SVGTextElement>} />`
- Add new prop - `textInnerRef` with`Ref<SVGTextElement>` and use this prop to measure size at annotation `<Label />` instead of use old innerRef 
https://github.com/airbnb/visx/blob/d5fb82e42af56da4ee325ac5d2028fde52a48b01/packages/visx-annotation/src/components/Label.tsx#L225-L235

### DIFF
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -223,7 +223,7 @@ export default function Label({
       )}
       {title && (
         <Text
-          innerRef={titleRef}
+          innerTextRef={titleRef}
           fill={fontColor}
           verticalAnchor="start"
           x={padding.left + (titleProps?.textAnchor === 'middle' ? innerWidth / 2 : 0)}
@@ -238,7 +238,7 @@ export default function Label({
       )}
       {subtitle && (
         <Text
-          innerRef={subtitleRef}
+          innerTextRef={subtitleRef}
           fill={fontColor}
           verticalAnchor="start"
           x={padding.left + (subtitleProps?.textAnchor === 'middle' ? innerWidth / 2 : 0)}
--- a/packages/visx-text/src/Text.tsx
+++ b/packages/visx-text/src/Text.tsx
@@ -21,7 +21,7 @@ export default function Text(props: TextProps) {
     width,
     ...textProps
   } = props;
-  
+
   const { x = 0, fontSize } = textProps;
   const { wordsByLines, startDy, transform } = useText(props);
 
--- a/packages/visx-text/src/Text.tsx
+++ b/packages/visx-text/src/Text.tsx
@@ -21,14 +21,14 @@ export default function Text(props: TextProps) {
     width,
     ...textProps
   } = props;
-
+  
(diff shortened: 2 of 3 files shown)


williaster:
Hey @vovakulikov thanks for really digging into the cause of the issue and the PR 🙏  One issue with this implementation is that there would be `n` refs, one for each line of text. I think that would likely deviate from the expected behavior on the consumer side since they'd only pass in one `innerRef` function.

If this is truly a bug with `react-use-measure` I'm wondering if we can file an issue there since otherwise we'd probably need to hack around it here. I made a simple sandbox to confirm what you described separate from `visx`

https://codesandbox.io/s/priceless-margulis-nctit?file=/index.tsx

chrome


firefox

vovakulikov:
@williaster Thanks for making this demo about `svg` and Firefox, this is exactly what I meant. 

> One issue with this implementation is that there would be n refs, one for each line of text

But we render just one `text` element within `<Text />` and use `tspan` for text lines, no? So it wouldn't be a problem since we still have one ref for `<Text />`

https://github.com/airbnb/visx/blob/d5fb82e42af56da4ee325ac5d2028fde52a48b01/packages/visx-text/src/Text.tsx#L31-L37

williaster:
> But we render just one text element within <Text /> and use tspan for text lines, no? 
Ah yes I totally misread the condition on that line, sorry. In this case maybe this approach will work fine! 🎉  I might still post an issue in `react-use-measure` to ask about it.

as far as avoiding breaking changes for this change, I think I'd probably vote for this option rather than casting to the wrong type
> Add new prop - textInnerRef withRef<SVGTextElement> and use this prop to measure size at annotation <Label /> instead of use old innerRef

Sorry it's so challenging to avoid breaking changes, have to be creative sometimes 😅

vovakulikov:
@williaster Totally with you about create issue in `react-use-measure`. For now I've added new prop to get ref of text element which was discussed above. Thanks for reviewing this btw.

williaster (approved):
Thanks again @vovakulikov 🙏 This looks great ❤️

PR #1180, opened 2021-04-20
new(@visx/zoom): adjust drag move to take optional offset

### LINKED ISSUE #1172: [@visx/zoom, @visx/event] implementing pan/zoom when mouse moves out of the element's bounds / browser bounds
So I asked this recently in #1169 and I've managed to find a solution.

So my problem is, I don't want pan/zoom to end when the mouse leaves the `<Zoom />` Elements bounds. I want panning to continue until mouseup is fired, no matter where the mouse moves in the meantime. We can see this happening in the example: https://airbnb.io/visx/docs/zoom

To do this, in my mousedown, I'm attaching the event listeners for mousemove and mouseup to the window instead of the dragged element. During mouse down i'm setting ` document.body.style["pointerEvents"] = "none"` to avoid events interfering with pan and on mouseup, i set pointer events back to auto and remove the event listeners.

This works great, expect for one problem. Whenever it goes from mousedown to mousemove, there is an annoying jump which is very bad for UX.

I found out that this is because in `zoom.dragMove`, `localPoint` is used which uses `event.target.getBoundingClientRect()`. This is a problem because the `event.target` of `mousedown` is going to be whatever Element is the target of the PanZoom mousedown, and since my mousemove/mouseup event listeners are attached to window and now I've set `pointer-events` to none, `event.target` for mousemove and mouseup will always be the html. This causes `event.target.getBoundingClientRect()` to give different `top` and `left` values, so `localPoint` will return an offset point.

Now I'm really stuck because with current implementation of zoom, I would have to provide the same `event.target` to `dragStart` and `dragMove`.

I don't see a way to do this since `event.target` of mousedown/dragStart will always be the panning surface, while `event.target` during mousemove will always be the document (`html`).

Especially considering most people probably don't want their pan to end as soon as the mouse leaves element bounds, and target of `mousemove` outside the browser bounds will always be `html` makes this a big caveat for anyone implementing this.

Unless I'm missing something and its actually easy to get the `event.target` of mousedown to be `html`, I would be happy to contribute a solution for this. I'm not sure what's the best way to do this though. 

I can see in `localPointGeneric`, there was a change to be explicit about the `node` where `getBoundingClientRect` is called instead of getting `event.target`. Maybe we can move this up to be able to explicitely pass the target?

boy51:
#### :rocket: Enhancements

- can provide optional offsetX and offsetY to `zoom.dragMove` 

#### :bug: Bug Fix

- Fixes #1172 / #1169 

I was trying to implement panning after the mouse leaves element/browser bounds. To achieve this I attached the mousemove listener directly to `window`. This resulted in the problem that `mousedown` has a different `event.target` than `mousemove`. As a result, there was a jump whenever it goes from `mousedown` to `mousemove`, resulting in a very bad ux.

I tried every solution I could think of and this is the only way I got it to work. If using `ParentSize`, I set the move offset to be `-x` and `-y` of `parent.ref.getBoundingClientRect`, which is the position of the zoom container relative to window.

~~This doesn't fix my problem 100%, because on the very first mousemove there is still a wrong coordinate. I'm not sure exactly why this happens though and I'm happy if someone knows more. Its not a huge dealbreaker for ux but definitely noticeable.~~

I realize this may not be a great solution, but its the only way I got it to work in my app, so I would be happy for any feedback and adjusting this so we can merge.

### DIFF
--- a/packages/visx-zoom/src/Zoom.tsx
+++ b/packages/visx-zoom/src/Zoom.tsx
@@ -212,14 +212,22 @@ class Zoom extends React.Component<ZoomProps, ZoomState> {
     this.setState({ isDragging: true });
   };
 
-  dragMove = (event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent) => {
+  dragMove = (
+    event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent,
+    options?: { offsetX?: number; offsetY?: number },
+  ) => {
     if (!this.state.isDragging || !this.startPoint || !this.startTranslate) return;
     const currentPoint = localPoint(event);
     const dx = currentPoint ? -(this.startPoint.x - currentPoint.x) : -this.startPoint.x;
     const dy = currentPoint ? -(this.startPoint.y - currentPoint.y) : -this.startPoint.y;
+
+    let translateX = this.startTranslate.translateX + dx;
+    if (options?.offsetX) translateX += options?.offsetX;
+    let translateY = this.startTranslate.translateY + dy;
+    if (options?.offsetY) translateY += options?.offsetY;
     this.setTranslate({
-      translateX: this.startTranslate.translateX + dx,
-      translateY: this.startTranslate.translateY + dy,
+      translateX,
+      translateY,
     });
   };
 


boy51:
I just fixed the problem where the first coordinate is wrong, it was caused within my app. 

So, I don't see any other way to do this if anyone wants to implement their listeners in the window.

williaster (approved):
Hey @boy51 sorry for the delayed review on this, but wow 🏆  thanks for the contribution and going super deep on making this work with `window` event listeners 🙏   

This seems like a creative solution to it and a reasonable addition. also @d3x42 I think some of the learnings here could help make brush work with `window` events / address the issues I just mentioned here https://github.com/airbnb/visx/pull/1164#issuecomment-840173688

PR #1208, opened 2021-05-12
fix(responsive): don't call resize.cancel if we never initialised

ahixon:
#### :bug: Bug Fix

- If you mount visx within react-testing-library, you might've encountered the error:
```Error: Uncaught [TypeError: resize.cancel is not a function]```
  
  We no longer reference the resize callback if we haven't initialised it yet.


williaster (approved):
Thanks for the fix @ahixon 🙏  lgtm!

PR #1218, opened 2021-05-15
deps: upgrade react-spring to v9

garfieldduck:
#### :house: Internal

`@visx/react-spring` is currently dependent on `react-spring` v8. 
While `react-spring` introduces several breaking changes in v9, and thus making `@visx/react-spring` incompatible with projects using the latest version of `react-spring`.

This PR upgrades the `react-spring` to v9.

##### Note

There are still [some typing issues](https://github.com/pmndrs/react-spring/issues/1483) in the formal release of v9 which prevents inferring the type of `useTransition` correctly.
To avoid this issue, this PR uses `9.0.0-rc.3` as the peer dependency instead of `9.1.2`.


packages/visx-demo/yarn.lock
```diff
@@ -0,0 +1,7675 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
williaster, line 1: since `visx` uses yarn workspaces, we have only committed the `yarn.lock` file in the repo root directory. so I think if you remove this file we can just rely on the updated root `yarn.lock` you've updated.

packages/visx-react-spring/src/axis/AnimatedTicks.tsx
```diff
@@ -20,63 +20,56 @@ export default function AnimatedTicks<Scale extends AxisScale>({
...
-    tick => `${tick.value}`,
-    useLineTransitionConfig({ scale, animateXOrY: horizontal ? 'x' : 'y', animationTrajectory }),
-  );
+  const animatedTicks = useTransition(ticks, {
+    ...useLineTransitionConfig({
+      scale,
+      animateXOrY: horizontal ? 'x' : 'y',
+      animationTrajectory,
+    }),
+    keys: (tick: { value: any }) => `${tick.value}`,
```
williaster, line 29: wonder if we could do something like this instead of using `any`

```tsx
// add this to the import above
import { ComputedTick, ... } from '@visx/axis/lib/types';

// in hook
keys: (tick: ComputedTick<Scale>) => `${tick.value}`,
```

williaster (commented):
Hey @garfieldduck thanks so much for the contribution! I had a couple super minor suggestions, and will also try to pull the branch locally since our visual diffing in CI doesn't capture animation. Otherwise this lgtm!

packages/visx-react-spring/src/axis/AnimatedTicks.tsx
```diff
@@ -20,63 +20,56 @@ export default function AnimatedTicks<Scale extends AxisScale>({
...
-            item,
-            key,
-            // @ts-ignore react-spring types only include CSSProperties
-            props: { fromX, toX, fromY, toY, opacity },
-          },
-          index,
-        ) => {
-          const tickLabelProps = allTickLabelProps[index] ?? allTickLabelProps[0] ?? {};
-          return item == null || key == null ? null : (
+      {animatedTicks(({ fromX, toX, fromY, toY, opacity }, item, { key }, index) => {
```
gazcn007, line 35: nit: im getting a ts warning -> `Property 'fromX' does not exist on type '{}'.ts(2339)` for { fromX, toX, fromY, toY, opacity }, 
might be good to keep the @ts-ignore if no explicit type is defined
```
// @ts-ignore react-spring types only include CSSProperties
```

packages/visx-demo/yarn.lock
garfieldduck: Yes, definitely! I didn't notice that I added this file accidentally.

packages/visx-react-spring/src/axis/AnimatedTicks.tsx
garfieldduck: Great point! I'll change it.

packages/visx-react-spring/src/axis/AnimatedTicks.tsx
garfieldduck: It may be caused by the type inference issue of `9.1.2`.
After running yarn, it picks up`9.1.2` automatically instead of `9.0.0-rc.3`.
But I think it's not nice to stick to `9.0.0-rc.3` instead of using `^9.0.0-rc.3` in peer dependencies, so I think putting `@ts-ignore` here is indeed a good idea.

williaster:
Hey @garfieldduck sorry for the delay on this PR, we've been trying to figure out how to stage this alongside the `react-17` `peerDependency` update. We've started in on that (#1268) and are thinking we should try to land this first because [this discussion](https://github.com/pmndrs/react-spring/issues/1478#issuecomment-830772961) suggests that `react@17` support was added in `react-spring@9.2.0`. 

Do you want to try bumping up to `^9.2.0` then we can try and get this merged / get a `@visx/*@2.0.0-alpha` out to build off of in the `react-17` work? If you don't have bandwidth in the next couple of days I can help try to bump to the latest and get it in. Thanks again for this BIG contribution! 🙏

packages/visx-react-spring/src/axis/AnimatedTicks.tsx
williaster: looks like this is still broken in `9.2.0` so maybe `@ts-ignore` is the way to go 😢

williaster:
Superseded by #1277

garfieldduck:
Hey @williaster, thanks so much for the upgrade 🎉!!!
Sorry I was on vacation and did not notice your mention.

PR #1219, opened 2021-05-18
fix(scale): upgrade d3-time to v2.4 for d3-scale compatibility

Pringels:
The latest minor version `3.3.0` of `d3-scale` makes use of utility functions exposed by `d3-time` version `2.1.0` (see: https://github.com/d3/d3-scale/commit/80ff9b202907968ad40ac84fa6347c6768f503c3#commitcomment-50668715)

This makes `d3-scale` `3.3.*` incompatible with `d3-time` `1.x` or even `2.0.0`

### DIFF
--- a/packages/visx-scale/package.json
+++ b/packages/visx-scale/package.json
@@ -33,10 +33,10 @@
   "dependencies": {
     "@types/d3-interpolate": "^1.3.1",
     "@types/d3-scale": "^3.2.1",
-    "@types/d3-time": "^1.0.10",
+    "@types/d3-time": "^2.0.0",
     "d3-interpolate": "^1.4.0",
     "d3-scale": "^3.2.3",
-    "d3-time": "^1.1.0"
+    "d3-time": "^2.1.1",
   },
   "publishConfig": {
     "access": "public"
--- a/packages/visx-scale/package.json
+++ b/packages/visx-scale/package.json
@@ -36,7 +36,7 @@
     "@types/d3-time": "^2.0.0",
     "d3-interpolate": "^1.4.0",
     "d3-scale": "^3.2.3",
-    "d3-time": "^2.1.1",
+    "d3-time": "^2.1.1"
   },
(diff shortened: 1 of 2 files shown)


packages/visx-scale/package.json
```diff
@@ -33,10 +33,10 @@
   "dependencies": {
     "@types/d3-interpolate": "^1.3.1",
     "@types/d3-scale": "^3.2.1",
-    "@types/d3-time": "^1.0.10",
+    "@types/d3-time": "^2.0.0",
     "d3-interpolate": "^1.4.0",
     "d3-scale": "^3.2.3",
-    "d3-time": "^1.1.0"
+    "d3-time": "^2.1.1",
```
Pringels, line 39:
```suggestion
    "d3-time": "^2.1.1"
```

packages/visx-scale/package.json
```diff
@@ -33,10 +33,10 @@
   "dependencies": {
     "@types/d3-interpolate": "^1.3.1",
     "@types/d3-scale": "^3.2.1",
-    "@types/d3-time": "^1.0.10",
+    "@types/d3-time": "^2.0.0",
     "d3-interpolate": "^1.4.0",
     "d3-scale": "^3.2.3",
-    "d3-time": "^1.1.0"
+    "d3-time": "^2.1.1"
```
williaster, line 39: I think with the dependency change, for CI we'll need to update the `yarn.lock` file in the monorepo root by running `yarn` there and committing the diff.

williaster (commented):
Thanks for the fix @Pringels 🙏  Bummer about the breaking minor version bump. I had one comment that hopefully will get the build working.

packages/visx-scale/package.json
Pringels: Fixed - thanks :)

williaster (approved):
Thanks again @Pringels! 💯

Issue #1193, opened 2021-05-03
Pie chart slice color?


omasback:
It seems not possible to assign different colors to different pie slices using the `<Pie>` props. Is there any way to do this without resorting to manually coding the children slice elements? Seems like an easy feature to provide that could save a lot of boilerplate.

goldenivan:
Hi @omasback,

You can specify the color of the pie in the path, especially the "fill" value. If you set "blue", all pies will be blue. If you define in your dataset a color property, you can use it like this : 
```
 <path d={pie.path(arc)} fill={data.color} />
```

An example of that : https://codesandbox.io/s/tf4ed?file=/Example.tsx:1783-1837

Regards,

williaster:
Yes this is a limitation discussed in this discussion thread https://github.com/airbnb/visx/discussions/1177 where I also suggested a simple feature implementation.

We can track with this issue, I'm happy to review a PR if anyone is interested!

omasback:
@williaster that perfectly answers my question, thanks!

iampueroo:
Hello @williaster . If no one is tackling this I can take this, but let me first try to communicate what I think the problem and fix actually is:

The core problem seems to be that the `Pie` [component](https://github.com/airbnb/visx/blob/master/packages/visx-shape/src/shapes/Pie.tsx#L85) does not allow _dynamic_ slice colors (the `fill` property) by default, and thus the user is required to render child `<path>` components by hand to pass through dynamic colors.

To fix this, the Pie component can have the `fill` prop be both a `string` (as it is today), but also a string accessor function. The function will be called and passed each arc datum, and returns the string for that datum's color. This way, a pie-chart with different colors can be built out-of-the-box (without custom children).

-----
So, in the end the example code will look like this (I took this from #1177):
```js
 <svg width={width} height={height}>
        <Group top={centerY + margin.top} left={centerX + margin.left}>
          <Pie
            data={letters}
            pieValue={frequency}
            pieSortValues={() => -1}
            outerRadius={radius}
            innerRadius={radius - donutThickness}
            fill={(datum) => getLetterFrequencyColor(datum.data.frequency)} // This would be new
          />
        </Group>
</svg>
```
and Pie's component will look like this:
```js
// Other stuff...
const fillAccessor = generateFillAccessor(fill); // Casts a `string` value to the identity function
// More other stuff
return (
    <Group className="visx-pie-arcs-group" top={top} left={left}>
      {arcs.map((arc, i) => (
[...]

PR #1225, opened 2021-05-26
new(shape): support dynamic fill directly in Pie

iampueroo:
This is a PR for issue #1193 

#### :boom: Breaking Changes

- None. Previously the `fill` property supported only `string` value. It still accepts the `string` value (but can also accept a string accessor function). I added a test case for both.

#### :rocket: Enhancements

- This PR enables developers to create a Pie chart with custom colors per wedge by just using the `<Pie />` component. Prior, they had to add the `<path>` elements as children of the `<Pie>` in order to dynamic colors.

#### :memo: Documentation

- The `fill` function can now be a `string` or a function that receives an `PieArcDatum<Datum>` and returns a `string` fill value: `(arc: PieArcDatum<Datum>) => string`.

#### :bug: Bug Fix

- Wouldn't call this a bug fix....

#### :house: Internal

- Um... ?

### DIFF
--- a/packages/visx-shape/src/shapes/Pie.tsx
+++ b/packages/visx-shape/src/shapes/Pie.tsx
@@ -35,7 +35,7 @@ export type PieProps<Datum> = {
   pieSortValues?: PiePathConfig<Datum>['sortValues'];
   /** Render function override which is passed the configured arc generator as input. */
   children?: (provided: ProvidedProps<Datum>) => React.ReactNode;
-  /** Optional accessor function to return the fill string value of a given arc **/
+  /** Optional accessor function to return the fill string value of a given arc. **/
   fill?: string | StringAccessor<Datum>;
 } & Pick<PiePathConfig<Datum>, 'startAngle' | 'endAngle' | 'padAngle'> &
   Pick<
@@ -99,4 +99,3 @@ export default function Pie<Datum>({
     </Group>
   );
 }
-
--- a/packages/visx-shape/src/shapes/Pie.tsx
+++ b/packages/visx-shape/src/shapes/Pie.tsx
@@ -7,6 +7,8 @@ import { arc as arcPath, pie as piePath } from '../util/D3ShapeFactories';
 
 export type PieArcDatum<Datum> = PieArcDatumType<Datum>;
 
+type StringAccessor<Datum> = (arc: PieArcDatum<Datum>) => string;
+
 export type ProvidedProps<Datum> = {
   path: ArcType<$TSFIXME, PieArcDatum<Datum>>;
   arcs: PieArcDatum<Datum>[];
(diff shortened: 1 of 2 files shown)


packages/visx-shape/src/shapes/Pie.tsx
```diff
@@ -7,6 +7,8 @@ import { arc as arcPath, pie as piePath } from '../util/D3ShapeFactories';
 
 export type PieArcDatum<Datum> = PieArcDatumType<Datum>;
 
+type StringAccessor<Datum> = (arc: PieArcDatum<Datum>) => string;
```
iampueroo, line 10: **Question**: I chose to provide `PieArcDatum` since it also includes the `index`, but it could also be `(datum: Datum, index: number)`. I'd like to get your thoughts on this - `PieArcDatum` may be unnecessary, especially if the pattern in other components is to merely provide the `Datum` type.

If not, I think the name shouldn't be `arc` but instead something like `arcDatum` or just `pieArcDatum` - I can change

packages/visx-shape/src/shapes/Pie.tsx
```diff
@@ -78,14 +83,28 @@ export default function Pie<Datum>({
...
+            {...restProps}
+          />
           {centroid?.(path.centroid(arc), arc)}
         </g>
       ))}
     </Group>
   );
 }
+
+function getFillGenerator<Datum>(fill: string | StringAccessor<Datum>): StringAccessor<Datum> {
```
iampueroo, line 104: Not sure where this should go so I threw it at the bottom. Please let me know if there is a better place for this.

packages/visx-shape/src/shapes/Pie.tsx
williaster: Good question, I think that `PieArcDatum` makes sense. Since it's strictly more info, including the computed arc size, it seems like a win to me.

packages/visx-shape/test/Pie.test.tsx
```diff
@@ -102,6 +105,22 @@ describe('<Pie />', () => {
     expect(fn).toHaveBeenCalled();
   });
 
+  test('it should accept a custom fill function', () => {
```
williaster, line 108: thanks for adding tests!

packages/visx-shape/src/shapes/Pie.tsx
williaster: See my comment above about potentially removing this/simplifying the logic, but if you'd like to keep it I think inside the file is fine :)

williaster (commented):
Thanks for the great addition @iampueroo! This looks good to me overall, I had one potential simplification (see what you think) but otherwise we can get this in!

iampueroo:
@williaster thanks for the review. I just pushed [this commit](https://github.com/airbnb/visx/pull/1225/commits/a0abe40139ae07c65bbad50564fdccb962f3017d) which removes the fill generator function and changes a variable name to be clearer. Happy to keep tweaking things as needed.

packages/visx-shape/src/shapes/Pie.tsx
```diff
@@ -33,6 +35,8 @@ export type PieProps<Datum> = {
   pieSortValues?: PiePathConfig<Datum>['sortValues'];
   /** Render function override which is passed the configured arc generator as input. */
   children?: (provided: ProvidedProps<Datum>) => React.ReactNode;
+  /** Optional accessor function to return the fill string value of a given arc **/
```
williaster, line 38: Super nit but these are transformed into docs so adding the period :)

```suggestion
  /** Optional accessor function to return the fill string value of a given arc. **/
```

williaster (approved):
Thanks for iterating! this looks good to me, I had one minor comment but I can commit it and merge 🚀

Issue #921, opened 2020-11-06
Adding hooks to @visx/geo


nschnierer:
Hey folks,

long time no see (added the geo package long time ago).
Nice to see the progress of this project 💯 

Regarding to #880 it's very complicated to get access to the `GeoProjection` interface of `d3-geo`.
Maybe this could be solved with a hook even better, like the following idea:
```
  const { features, projection } = useProjection({
    projection: geoMercator,
    data: usa.features,
    scale: 1.0 // defaultScale
  });
```

Right now, I'm not sure how to handle projection updates (changing scale etc.).
So any thoughts? Are there already similar approaches in other packages?

BTW: This hook could used therefore by `<Mercator />` etc., I think these components will be there for easier usage.

Inspiration: https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect

williaster:
Hey @nschnierer 👋  sorry this slipped through without a quick reply. 

Adding a `useProjection` hook sounds like a great addition, as hooks have become more popular we've been adding them as needed, often refactoring the component implementation to use the hook as you mentioned. Definitely happy to review a PR for this if you (or someone else) is interested in the addition!

Regrading projection updates, if you're thinking about animation, @techniq implemented an animated scale in his repo [here](https://github.com/techniq/vx-hierarchy-examples/blob/master/src/scales/AnimatedScale.tsx) (demo [here](https://codesandbox.io/s/github/techniq/vx-hierarchy-examples?file=/src/examples/IcicleHorizontal.tsx)), a similar approach might work for updating a projection. We've mostly still tried to stay un-opinionated on animations so app consumers can use what they like, but added a `@visx/react-spring` package with a few things in it because the animated primitives aren't always straight forward to implement / there's utility in abstracting them (and `react-spring` is as good as it gets perf-wise).

Curious to hear your thoughts.

williaster:
I don't have the bandwidth to add this at the moment but think it's a great request and would happily review a PR if someone wants to take a shot!

danielimmke:
I can take it. I will add it to my TODO list.

PR #1236, opened 2021-06-11
new(geo): pass projection to child override, add U.S.A. demo

danielimmke:
#### :boom: Breaking Changes

- None

#### :rocket: Enhancements

- Discussed on #921. I did not add a hook. Instead I made the constructed projection method available to child elements. In the future if a hook is created it shouldn't be difficult to replace it under the hood without changing the API surface.
- Exposing the projection function makes it much easier to add things like text labels and custom points on a map - while retaining things like custom scaling/translation.
- I added a new demo of a US map that makes use of this new functionality.

#### :memo: Documentation

- I added in documentation covering these changes.

#### :bug: Bug Fix

- While I was working on this, I noticed that not all of D3's projections expose all of the same methods. So I added a checker to the methods in the Projection component. AlbersUsa does not have `rotate` or `center` methods and if you passed those props on the `<AlbersUsa />` component it would brick.

I'm open to any and all suggestions, this is a first cut I know it may need some refining.


danielimmke:
@williaster

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+  properties: { name: string };
+}
+
+// @ts-ignore
+const { features: unitedStates } = topojson.feature(topology, topology.objects.states) as {
+  type: 'FeatureCollection';
+  features: FeatureShape[];
+};
+
+export const colors: Array<string> = ['#744DCA', '#3D009C', '#9020FF', '#C630FD']
```
williaster, line 29: nit – just more consistent with syntax we use elsewhere

```suggestion
export const colors: string[] = ['#744DCA', '#3D009C', '#9020FF', '#C630FD']
```

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+  'KY': [10, 0],
+  'TN': [0, 4]
+}
+
+const GeoAlbersUsa = ({ width, height, fullSize = true }: GeoAlbersUsaProps) => {
+  const centerX = width / 2;
+  const centerY = height / 2;
+  const scale = (width + height) / 1.55;
+
+  const [ displayLabels, setDisplayLabels ] = useState(fullSize)
```
williaster, line 50: this may fail CI, you might want to try running the following locally as I noticed a few missing semi-colons, etc. (from the monorepo root)

```bash
yarn format
yarn lint:fix
```

```suggestion
  const [displayLabels, setDisplayLabels] = useState(fullSize);
```

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+              fill: '#FFF',
+              fontFamily: 'sans-serif',
+              cursor: 'default'
+            }
+
+            if(abbr === 'HI') {
+              stylesObj.fill = '#3C019C'
+            }
+
+            const ignore = ['VT', 'NH', 'MA', 'RI', 'CT', 'NJ', 'DE', 'MD']
```
williaster, line 79: could move this outside the function since it's a constant, maybe near `nudges`?

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  d={path || ''}
+                  fill={(i % 2 === 0) ? colors[0] : colors[1]}
+                  stroke={background}
+                  strokeWidth={0.5}
+                />
+              )
+            }
+
+            return (
+              <>
```
williaster, line 94: the key should be on the top-level entity so will have to use `React.Fragment` instead of the shorthand `<>`

```suggestion
              <React.Fragment key={`map-feature-${i}`}>
```

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </svg>
+      {fullSize && <label style={{position: 'relative', left: '20px', top: '-60px', fontSize: '14px', display: 'flex', alignItems: 'center'}}>
+        <input type="checkbox" checked={displayLabels} onChange={() => { setDisplayLabels(!displayLabels) }} />
+        &nbsp;Display labels
+      </label>}
+    </>
+  );
+};
+
+export default GeoAlbersUsa;
```
williaster, line 121: can we add a newline here?

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+// @ts-ignore
+const { features: unitedStates } = topojson.feature(topology, topology.objects.states) as {
+  type: 'FeatureCollection';
+  features: FeatureShape[];
+};
+
+export const colors: Array<string> = ['#744DCA', '#3D009C', '#9020FF', '#C630FD']
+
+const nudges = {
```
williaster, line 31: nit – could we make this name a little more descriptive like `coordOffsets` or similar? took a couple passes to understand it

packages/visx-geo/src/projections/Projection.tsx
```diff
@@ -135,15 +136,15 @@ export default function Projection<Datum extends GeoPermissibleObjects>({
...
-  if (clipAngle) currProjection.clipAngle(clipAngle);
-  if (clipExtent) currProjection.clipExtent(clipExtent);
-  if (scale) currProjection.scale(scale);
-  if (translate) currProjection.translate(translate);
-  if (center) currProjection.center(center);
-  if (rotate) currProjection.rotate(rotate);
-  if (precision) currProjection.precision(precision);
-  if (fitExtent) currProjection.fitExtent(...fitExtent);
-  if (fitSize) currProjection.fitSize(...fitSize);
+  if (clipAngle && currProjection.clipAngle) currProjection.clipAngle(clipAngle);
```
williaster, line 139: thanks for catching this, great find! surprised TS didn't flag it 😅

williaster (commented):
Thanks for the quick addition @danielimmke! I had a few minor comments that should be a quick pass to fix, and great find on the bug ❌  🐛 

Could you also add a screenshot of the new example?

danielimmke:
Sure - my only question - where do you want the screenshot? Just here on the PR to show you what it looks like? If so, here they are:

Geo page

Standalone

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <AlbersUsa<FeatureShape>
+          data={unitedStates}
+          scale={scale}
+          translate={[centerX, centerY - 25]}
+        >
+          {({ features }) => features.map(({ feature, path, projection }, i) => {
+            let coords = projection(geoCentroid(feature))
+            let abbr:string = stateAbbrs[feature.id]
+
+            if(nudges[abbr]) {
```
williaster, line 64: looks like TS doesn't like this, if you declare nudges as `const coordOffsets: Record<string, [number, number]> = ...` it might solve it. I think now TS interprets it as a mutable object with no type. even doing `const coordOffsets = { ... } as const` might be enough

williaster:
looks great! 🤩  yeah I just meant on the PR, we don't spin up a docs site for PRs but after merging it'll be live in the docs. I was just curious beforehand.

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/yarn.lock
```diff
@@ -0,0 +1,8910 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
williaster, line 1: we only commit the lock file in the root of the monorepo, so if possible can we remove this file?

other examples in `@visx/demo` have the `d3-geo` dependency I believe, so shouldn't need to update the root `yarn.lock` file either. just to explain further: the sandbox examples have a `package.json` so they work on codesandbox. if we were actually adding a new dep here we would need to add it both in the sandbox `package.json` **and** in the `@visx/demo` `package.json`, and then install the dep by running `yarn` in the monorepo root. 

all that to say that if you remove this file, you shouldn't need to do anything else since you added the appropriate dependencies in the sandbox `package.json` 😅

packages/visx-demo/src/pages/geo-albers-usa.tsx
```diff
@@ -8,6 +8,7 @@ const GeoAlbersUsaPage = () => {
   return (
     <Show
       events
+      /* @ts-ignore */
```
williaster, line 11: hmm, might be able to remove this if we make the component type more specific. if this doesn't work no worries.

```tsx
// current
const GeoAlbersUsa = ({ width, height, fullSize = true }: GeoAlbersUsaProps) => { ... }

// proposed
const GeoAlbersUsa: React.FC<GeoAlbersUsaProps> = ({ width, height, fullSize = true }) => { ... }
```

packages/visx-demo/src/sandboxes/visx-geo-albers-usa/Example.tsx
```diff
@@ -26,96 +26,125 @@ const { features: unitedStates } = topojson.feature(topology, topology.objects.s
...
           data={unitedStates}
           scale={scale}
           translate={[centerX, centerY - 25]}
         >
-          {({ features }) => features.map(({ feature, path, projection }, i) => {
-            let coords = projection(geoCentroid(feature))
-            let abbr:string = stateAbbrs[feature.id]
+          {({ features }) =>
+            features.map(({ feature, path, projection }, i) => {
+              const coords:[number,number] | null = projection(geoCentroid(feature));
```
williaster, line 70: nit – might fail prettier/linting
```suggestion
              const coords:[number, number] | null = projection(geoCentroid(feature));
```

williaster (commented):
Looking really great! couple more small comments (sorry 🙄 ), the most important one is removing the `yarn.lock` then we should be good to go! 🎉 

Thanks for iterating on this 🙏

danielimmke:
@williaster Thanks for the weekend review - I have to admit, I'm not very strong with TypeScript and I find a lot of the error messages to be difficult to understand.

It might be quicker if you could work directly on it to resolve the remaining TS issues. I tried that suggestion you had and it didn't work.

danielimmke:
I don't know why the build keeps failing - something about "not authorized" I think I got all the lint issues done

williaster (approved):
Hey @danielimmke sorry for the build issues/delay in getting this merged. Github actions recently changed such that they now require repo admins to approve the CI flow to run per-commit, and this has caused issues with how our screenshot testing auth works.

The tests/linting/typescript stuff all passed tho so I think we can just merge it. Thanks again for the great contribution!

danielimmke:
No worries @williaster - you were very quick to review and it is much appreciated. We are using Visx at the CDC as the foundation for a lot of stuff and I'm switching over our mapping to it. There's a chance I will have additional PRs going forward so I'm glad to know that there's someone committed to maintaining from an open source perspective.

Issue #1201, opened 2021-05-05
change examples to have passed-in data


ianstormtaylor:
Right now most of the examples I've seen use static data supplied by `import` statements. But I think the examples themselves should take the data in as a prop instead, because that is how most people will need to write their components.

This has downstream effects because most of the uses of scales and data transformations are defined outside the React components, but this doesn't fly in real-world use cases, where `useMemo` should be used instead.

Overall it just makes it more taxing to translate from examples when trying to build new charts.

williaster:
Hey @ianstormtaylor thanks for checking out `visx` and for this request. It's a lot of work to refactor 30+ examples in this way/we don't currently have the bandwidth to do that, but I'm happy to review any PRs refactoring examples in this way.

dennisja:
@ianstormtaylor Are you working on this?
If not, I guess it would be something good for me to work on.
@williaster Would you prefer all of the examples refactored in 1 pull request?

williaster:
Hey @dennisja 👋 , sweet! I think it'd be easiest to review these as separate PRs because there's ~30 examples which would be a huge PR.

csdiehl:
Hey I'm happy to refactor some of these examples in a separate PR as well, if more help is needed

PR #1240, opened 2021-06-16
refactor/bar and bar stack examples

dennisja:
#### :memo: Documentation

Hello @williaster 
This is a follow-up PR of the discussion we had on #1201 
I have made changes to two examples (bar and bar stack). The feedback I get on this will help me refactor the other examples.

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-bars/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-bars/Example.tsx
@@ -11,19 +11,24 @@ const verticalMargin = 120;
 const getLetter = (d: LetterFrequency) => d.letter;
 const getLetterFrequency = (d: LetterFrequency) => Number(d.frequency) * 100;
 
-export type BarsProps = {
-  width: number;
+export type BarsProps<Datum> = {
+  data: readonly Datum[];
   height: number;
+  width: number;
+  xAccessor: (d: Datum) => string;
+  yAccessor: (d: Datum) => number;
+
   events?: boolean;
-  data?: readonly LetterFrequency[];
 };
 
-export default function Example({
+function Bars<Datum>({
   width,
   height,
   events = false,
-  data = letterFrequency.slice(5),
-}: BarsProps) {
+  data,
+  xAccessor,
+  yAccessor,
+}: BarsProps<Datum>) {
   // bounds
   const xMax = width;
   const yMax = height - verticalMargin;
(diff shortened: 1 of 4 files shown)


packages/visx-demo/src/sandboxes/visx-bars/Example.tsx
```diff
@@ -16,9 +15,15 @@ export type BarsProps = {
   width: number;
   height: number;
   events?: boolean;
+  data?: readonly LetterFrequency[];
```
williaster, line 18: when reading #1201 , I interpreted it as refactoring a given example to take in fully arbitrary data, not a specific dataset.

so for `Bars` I think that would require TypeScript generics for data and for users to pass in the relevant x/y accessors, for example

```tsx
type BarsProps<Datum> = {
  ...
  data: Datum[];
  xAccessor: (d: Datum) => string; 
  yAccessor: (d: Datum) => number;
}

// usage
<BarsExample
  data={letterFrequency}
  xAccessor={d => d.letter}
  yAccessor={d => d.frequency}
  {...}
/>
```

The `BarStack` example would need to do similar except it would need to have accessors for x/y/stack, and probably also a TypeScript generic for the possible stack keys.

williaster (commented):
Hey @dennisja ! thanks for starting on this! I had one major comment about the approach here. let me know if it doesn't make sense / you want to discuss further.

packages/visx-demo/src/sandboxes/visx-bars/Example.tsx
dennisja: Thanks @williaster this makes more sense. I will work on it this week. I wish I saw it over the weekend

dennisja:
Hey @williaster 
Sorry that I took long to work on this. I had a few weeks off work and wasn't writing code during that time. But the good news is I would like to work on all the examples if no one else is willing to, so I will dedicate time for 1 or two examples every week until we are done. I'm just coming back to data visualizations so I might miss a few things but I believe it will be easier when working on the other examples.

There is an alternative I used besides the Generics alternative.
It would involve defining the data structure the graph will take based on the standardized graph names i.e for a bar stark
we would have something like
```ts
type StackCategory {
 name: string; // the city name in CityTemperature
 value: number; // the temperature in CityTemperature
}

type BarData = {
   name: string // the date for CityTemeperature
   categories: StackCategory[]
}

type BarStackProps = {
  data: BarData[];
  getXValue: (d: BarData)=>string
   ....
}
```
To use this with data from any source, you would have to transform the incoming data structure to a data structure similar to `BarData[]`.

What do you think of this alternative in comparison to the generics approach?
Also, Happo, is failing. I'm not sure why because I don't have access to the logs.

williaster:
hey @dennisja ! sorry for the delay on my end. that's really exciting about the other examples! that sounds great to me if you are interested.

regarding your proposal for the standardized graph names, I think that makes sense/the code looks good. I think purely relying on generics might not work for some of the more complex examples (such as BarStack) so I think that is a good solution.

also sorry about the happo diff, we aren't able to grant non-admins access to it as far as I know (I can double check). for now I wouldn't worry about it and I can relay anything that's an issue. currently it's just picking up on extra bars being added to the bars example (which is fine)

image

williaster (approved):

dennisja:
@williaster 
Thanks for looking at this.
I think if everything is okay with it we can land it. 
I can use the alternative I described to implement the Horizontal Barstack

PR #1251, opened 2021-06-24
build: skip release workflow when in forked repo

kristw:
#### :house: Internal

- pushing to `master` in forked repo triggers the `push.yml` github action workflow which include npm release. The release won't be successful (because fork doesn't have npm credentials) but it may bump the versions of `package.json` in the fork.

This PR add condition in the workflow to only run `commit size` and `release` steps from `airbnb/visx` but not from `some-fork/visx`.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -35,22 +35,24 @@ jobs:
         run: yarn build
 
       - name: 📐 Commit package sizes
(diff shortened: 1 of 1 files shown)


williaster (approved):
Ooo nice 🙏  lgtm. thanks for finding / fixing this!

PR #1252, opened 2021-06-24
feat(scale): support rounding for symlog

### LINKED ISSUE #1095: [scale] `round` parameter missing for `scaleSymlog`
It is not possible to create Symlog scales with rounding from a config: https://codesandbox.io/s/exciting-buck-bnq3q (while it works for other scales such as Linear or Log)

At first glance, adding `round` to the list of accepted parameters in https://github.com/airbnb/visx/blob/master/packages/visx-scale/src/scales/symlog.ts#L13 should fix it

kristw:
#### :bug: Bug Fix

Fix #1095

### DIFF
--- a/packages/visx-scale/package.json
+++ b/packages/visx-scale/package.json
@@ -32,10 +32,10 @@
   "homepage": "https://github.com/airbnb/visx#readme",
   "dependencies": {
     "@types/d3-interpolate": "^1.3.1",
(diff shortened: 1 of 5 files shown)


kristw:
This one should be good to go aside from happo.

williaster (approved):
LGTM. Sorry about happo, I recently found out that because we don't run happo in the push workflow (just opened #1254 to hopefully fix)

PR #1255, opened 2021-06-25
new(xychart/tooltip): add renderGlyph property

Janpot:
#### :rocket: Enhancements

- Implement https://github.com/airbnb/visx/issues/988


packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -18,26 +19,24 @@ const TOOLTIP_NO_STYLE: React.CSSProperties = {
...
+};
+
 export type TooltipProps<Datum extends object> = {
   /**
    * When TooltipContext.tooltipOpen=true, this function is invoked and if the
    * return value is non-null, its content is rendered inside the tooltip container.
    * Content will be rendered in an HTML parent.
    */
   renderTooltip: (params: RenderTooltipParams<Datum>) => React.ReactNode;
+  renderGlyph?: (params: RenderGlyphProps<Datum>) => React.ReactNode;
```
williaster, line 39: could we add a prop comment here so it'll appear in the docs?

packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -178,11 +178,30 @@ function TooltipInner<Datum extends object>({
   }
 
   // collect positions + styles for glyphs; glyphs always snap to Datum, not event coords
-  const glyphProps: GlyphProps[] = [];
+  const glyphs: React.ReactNode[] = [];
+
+  const renderGlyph =
+    renderGlyphProp ||
+    (<Datum extends object>(props: RenderGlyphProps<Datum>) => {
```
williaster, line 185: might be able to pull this outside the render function as `defaultRenderGlyph` since it should be a constant?

packages/visx-xychart/src/components/Tooltip.tsx
```diff
@@ -275,34 +302,9 @@ function TooltipInner<Datum extends object>({
...
-                    strokeWidth={strokeWidth}
-                    paintOrder="fill"
-                    {...glyphStyle}
-                  />
-                </svg>
-              </TooltipInPortal>
-            ),
-          )}
+          {glyphs.map((glyph, i) => (
+            <React.Fragment key={i}>{glyph}</React.Fragment>
```
williaster, line 306: any reason to not include the `TooltipInPortal`? I think this means the glyph(s) would be rendered inside the svg container and the rest of the tooltip elements would be in portals 🤔

williaster (commented):
Excited for this! Left a few initial comments.

packages/visx-xychart/src/components/Tooltip.tsx
Janpot: Had it in there for now because it accesses the theme for the `stroke`

packages/visx-xychart/src/components/Tooltip.tsx
Janpot: Because it needs to know the dimensions of the glyph in order to position itself correctly

Janpot: @williaster I think we can either do a full width/height portal
```tsx

          {glyphs.map((glyph, i) => (
            <TooltipInPortal
              key={i}
              className="visx-tooltip-glyph"
              left={0}
              top={0}
              offsetLeft={0}
              offsetTop={0}
              detectBounds={false}
              style={TOOLTIP_NO_STYLE}
            >
                {glyph}
              </svg>
            </TooltipInPortal>
          ))}
```

or just

```
          <svg>
            {glyphs.map((glyph, i) => (
              <React.Fragment key={i}>{glyph}</React.Fragment>
            ))}
          </svg>
```

packages/visx-xychart/src/components/Tooltip.tsx
williaster: could potentially still move it out and use context to access the theme but nbd

packages/visx-xychart/src/components/Tooltip.tsx
williaster: makes sense. I agree the crosshairs could potentially be moved out of the portals as well as it's not as important if they get clipped (that was the main impetus for rendering the tooltip in a portal).

williaster:
@Janpot this looks solid overall, lmk if you want to convert it from a draft/when it's ready to merge on your end (I'll probably pull it to play with it locally a bit)

packages/visx-xychart/src/components/Tooltip.tsx
Janpot: Actually, I can make it work using `overflow` on the `svg`. Ironically I was the one who also came up with [that solution](https://github.com/airbnb/visx/issues/990#issue-779509740) for the crosshairs 🤦.

Janpot:
@williaster I have no clue what this happo is unhappy about, is there any way for me to check out these diffs?

williaster:
hey @Janpot sorry for the happo diffs, unfortunately I don't think we're able to open up access to it (I will double check tho, it's been a while since I asked). Sometimes it's a little too sensitive and gives false positives which was true in this case, so I accepted the diffs/no need to worry about it 😅 

image

williaster (approved):
Woot! this looks great to me, thanks for the addition and iterating on the implementation!

Is there anything else you want to tweak? If you are happy we can merge / release it!

Janpot:
Nope, this can be merged as far as I'm concerned

elliot-at-looker:
Can someone provide an example of using the renderGlyph prop? I am having a hard time seeing anything besides a circle for the tooltip glyph

PR #1257, opened 2021-06-29
fix(zoom): fix provided zoom interface

boy51:
#### :bug: Bug Fix

- Added my changes from #1180 to types/provided zoom interface

As a side note, I would really appreciate if we can add react@17 to the peer dependencies... Currently 17 is not included. I'm running production on 17 for a while and it works fine. With the latest changes in npm 7, there are many errors when trying to use with react 17 even though it works fine. Which is kinda sad considering most people trying to get into this lib are probably just installing react 17 thinking it doesn't work. If you guys are cool with it I can make PRs for this too.

### DIFF
--- a/packages/visx-zoom/src/types.ts
+++ b/packages/visx-zoom/src/types.ts
@@ -47,7 +47,10 @@ export interface ProvidedZoom {
   /** Callback for dragEnd, sets isDragging to false. */
   dragEnd: () => void;
   /** Callback for dragMove, results in a scale transform. */
-  dragMove: (event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent) => void;
+  dragMove: (
+    event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent,
+    options?: { offsetX?: number; offsetY?: number },
+  ) => void;
   /** Callback for dragStart, sets isDragging to true.  */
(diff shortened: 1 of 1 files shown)


packages/visx-zoom/src/types.ts
```diff
@@ -47,7 +47,7 @@ export interface ProvidedZoom {
   /** Callback for dragEnd, sets isDragging to false. */
   dragEnd: () => void;
   /** Callback for dragMove, results in a scale transform. */
-  dragMove: (event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent) => void;
+  dragMove: (event: React.MouseEvent | React.TouchEvent | MouseEvent | TouchEvent, options?: { offsetX?: number; offsetY?: number }) => void;
```
williaster, line 50: lgtm but I think `prettier` is unhappy with this being on a single line. `yarn format` should work, but may also affect other files (fine to just not commit those, or feel free to!)

williaster (approved):
Thanks @boy51 ! This lgtm but had one comment about prettier.

Thanks for the nudge on React 17. We're planning internally for how to do this. I'll tag you on the open issue for that to coordinate if you'd like to help 🙏  Thanks again!

packages/visx-zoom/src/types.ts
boy51: Fixed :)

williaster:
thanks @boy51!

PR #1261, opened 2021-06-30
feat(shape): add more segmentation for SplitLinePath

### LINKED ISSUE #920: SplitLinePath: The segments are not split correctly
I am using SplitLinePath for a chart and when I am giving the segments array, the generated SVG points are not correct,  meaning that they are not split correctly and the first segment has points from the second segment and so on, so the segments are shifted a little bit. Did someone else had this issue or can help me?

kristw:
#### :rocket: Enhancements

`SplitLinePath` can split line by multiple algorithms

```tsx
<SplitLinePath segmentation="x" />
```

- `x`: Split based on x-position, assuming x values increase only (`segment[i].x > segment[i-1].x`) or decrease only (`segment[i].x < segment[i-1].x`).
- `y`: Split based on y-position, assuming y values increase only (`segment[i].y > segment[i-1].y`) or decrease only (`segment[i].y < segment[i-1].y`).
- `length`: Assuming the path length between consecutive points are equal. (This was the original algorithm that led to #920.)

Default is set to `x`.

#### :memo: Documentation

- Update `SplitLinePath` example

#### :bug: Bug Fix

- Fix #920 

image

### DIFF
--- a/packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/Example.tsx
+++ b/packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/Example.tsx
@@ -1,10 +1,9 @@
 import React, { useMemo } from 'react';
-import { scaleLinear } from '@visx/scale';
 import { curveCardinal } from '@visx/curve';
 import { LinePath, SplitLinePath } from '@visx/shape';
 import { LinearGradient } from '@visx/gradient';
-
-import generateSinPoints from './generateSinPoints';
+import { SplitLinePathChildren } from '@visx/shape/lib/shapes/SplitLinePath';
+import generateSinSegments from './generateSinSegments';
 
 type Point = { x: number; y: number };
 const getX = (d: Point) => d.x;
@@ -13,7 +12,7 @@ export const background = '#045275';
 export const backgroundLight = '#089099';
 export const foreground = '#b7e6a5';
 
-export type SplitLinePathProps = {
+export type SplitLinePathExampleProps = {
   width: number;
   height: number;
   margin?: { top: number; right: number; bottom: number; left: number };
@@ -22,40 +21,53 @@ export type SplitLinePathProps = {
   numberOfSegments?: number;
 };
 
-export default function SplitPath({
+const CustomSegment: SplitLinePathChildren = ({ segment, styles }) => (
+  <g>
+    {segment.map(({ x, y }, i) =>
+      i % 8 === 0 ? (
+        <circle
+          key={i}
+          cx={x}
+          cy={y}
+          r={10 * (i / segment.length)}
+          stroke={styles?.stroke}
+          fill="transparent"
+          strokeWidth={1}
+        />
+      ) : null,
+    )}
+  </g>
+);
+
+export default function SplitLinePathExample({
   width,
   height,
   numberOfWaves = 10,
   pointsPerWave = 100,
-  numberOfSegments = 8,
-}: SplitLinePathProps) {
-  const data = useMemo(() => generateSinPoints({ width, height, numberOfWaves, pointsPerWave }), [
-    width,
-    height,
-    numberOfWaves,
-    pointsPerWave,
-  ]);
-
-  const dividedData = useMemo(() => {
-    const segmentLength = Math.floor(data.length / numberOfSegments);
-    return new Array(numberOfSegments)
-      .fill(null)
-      .map((_, i) => data.slice(i * segmentLength, (i + 1) * segmentLength));
-  }, [numberOfSegments, data]);
-
-  const getScaledX = useMemo(() => {
-    const xScale = scaleLinear({ range: [0, width], domain: [0, width] });
-    return (d: Point) => xScale(getX(d)) ?? 0;
-  }, [width]);
-
-  const getScaledY = useMemo(() => {
-    const yScale = scaleLinear({ range: [0, height], domain: [height, 0] });
-    return (d: Point) => yScale(getY(d)) ?? 0;
-  }, [height]);
+}: SplitLinePathExampleProps) {
+  const data = useMemo(
+    () => ({
+      leftToRight: generateSinSegments({
+        width: width / 2,
+        height: height / 2,
+        numberOfWaves,
+        pointsPerWave,
+      }),
+      rightToLeft: generateSinSegments({
+        width: width / 2,
+        height: height / 2,
+        numberOfWaves,
+        pointsPerWave,
+        direction: 'right-to-left',
+      }),
+    }),
+    [width, height, numberOfWaves, pointsPerWave],
+  );
 
   return width < 10 ? null : (
     <div>
       <svg width={width} height={height}>
+        {/* Background */}
         <LinearGradient
           id="visx-shape-splitlinepath-gradient"
           from={background}
@@ -72,11 +84,12 @@ export default function SplitPath({
           rx={14}
         />
 
-        <g transform={`rotate(${0})translate(${-0}, ${-height * 0.5})`}>
+        {/* left to right */}
+        <g transform={`translate(${width / 2}, ${height / 4})`}>
           <LinePath
-            data={data}
-            x={getScaledX}
-            y={getScaledY}
+            data={data.leftToRight.flat()}
+            x={getX}
+            y={getY}
             strokeWidth={8}
             stroke="#fff"
             strokeOpacity={0.15}
@@ -84,9 +97,10 @@ export default function SplitPath({
           />
 
           <SplitLinePath
-            segments={dividedData}
-            x={getScaledX}
-            y={getScaledY}
+            sampleRate={2}
+            segments={data.leftToRight}
+            x={getX}
+            y={getY}
             curve={curveCardinal}
             styles={[
               { stroke: foreground, strokeWidth: 3 },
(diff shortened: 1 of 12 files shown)


packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
+      }
+    }
 
     return lineSegments;
   } catch (e) {
     return [];
   }
 }
-
-export default memoize(
```
kristw, line 76: The rewritten `SplitLinePath` let us use `useMemo` there so no more need to `memoize` here.

packages/visx-shape/src/shapes/SplitLinePath.tsx
```diff
@@ -1,73 +1,86 @@
...
+        path: pathString,
+        segmentation,
+        pointsInSegments,
+        sampleRate,
+      }),
+    [pathString, segmentation, pointsInSegments, sampleRate],
   );
 
   return (
-    <LinePath data={combinedSegments} defined={defined} curve={curve} x={x} y={y}>
```
kristw, line 46: Using the `line` factory directly simplifies the code and also let us use `useMemo` to memoize the pathString.

packages/visx-shape/src/util/getOrCreateMeasurementElement.ts
```diff
@@ -0,0 +1,26 @@
+const SVG_NAMESPACE_URL = 'http://www.w3.org/2000/svg';
```
kristw, line 1: Extracted from `getSplitLineSegments.ts`

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/generateSnakePath.ts
```diff
@@ -0,0 +1,58 @@
...
+  y: number;
+}
+
+const pointKey = ({ x, y }: Point) => [x, y].join('_');
+
+function distance(a: Point, b: Point) {
+  return Math.sqrt((b.x - a.x) ** 2 + (b.y - a.y) ** 2);
+}
+
+export default function generateSnakePath({
```
kristw, line 12: generate a continuous path that fill rectangular space, similar to the old Nokia snake game.

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/generateSinPoints.ts
```diff
@@ -14,7 +14,7 @@ export default function generateSinPoints({
   const distanceBetweenPoints = waveLength / pointsPerWave;
   const sinPoints: { x: number; y: number }[] = [];
 
-  for (let waveIndex = 0; waveIndex <= numberOfWaves; waveIndex += 1) {
```
kristw, line 17: off by one error

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/package.json
```diff
@@ -12,6 +12,7 @@
     "@visx/responsive": "latest",
     "@visx/scale": "latest",
     "@visx/shape": "latest",
+    "lodash": "^4.17.21",
```
williaster, line 15: looks like we don't have `lodash` as a `@visx/demo` dep. It's a little annoying / fragile, but sandboxes need all of their deps so they work as an independent app, but adding them here doesn't add them to the demo package. so can we do the following?

- add them to the demo `package.json`
- run `yarn` in the root to update the `yarn.lock` file

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/generateSnakePath.ts
williaster: could we commit that? 😆 

```suggestion
// generate a continuous path that fill rectangular space, similar to the classic Nokia snake game.
export default function generateSnakePath({
```

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/generateSinSegments.ts
```diff
@@ -0,0 +1,51 @@
+import generateSinPoints from './generateSinPoints';
+
+type Point = { x: number; y: number };
+
+export default function generateSinSegments({
```
williaster, line 5: this is great 🔥

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/generateSinPoints.ts
williaster: 😱  🤦

packages/visx-demo/src/sandboxes/visx-shape-splitlinepath/Example.tsx
```diff
@@ -72,21 +119,142 @@ export default function SplitPath({
           rx={14}
         />
 
-        <g transform={`rotate(${0})translate(${-0}, ${-height * 0.5})`}>
+        {/* left to right */}
+        <g transform={`translate(${PADDING}, ${height / 4})`}>
+          {/* Render all segments as a single line for comparison */}
+          <LinePath
+            data={data.leftToRight.flat()}
```
williaster, line 126: til `flat()`

packages/visx-shape/src/shapes/SplitLinePath.tsx
williaster: good call

packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
+        while (current < numSegments - 1 && isBeyondSegments[current + 1](position)) {
+          current += 1;
+        }
+        lineSegments[current].push(sample);
       }
+    } else {
+      // segmentation === "length"
+      const numPointsInSegment = pointsInSegments.map(points => points.length);
+      const numPoints = numPointsInSegment.reduce((sum, curr) => sum + curr, 0);
+      const lengthBetweenPoints = totalLength / (numPoints - 1);
```
williaster, line 85: Not sure if this matters/might be an unlikely edge case

```suggestion
      const lengthBetweenPoints = totalLength / Math.max(1, (numPoints - 1));
```

packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
 
-    const totalPathLength = pathElement.getTotalLength();
-    const totalPieces = segments.reduce((sum, curr) => sum + curr.length, 0);
-    const pieceSize = totalPathLength / totalPieces;
+    const numSegments = pointsInSegments.length;
+    const lineSegments: LineSegments = pointsInSegments.map(() => []);
 
-    let cumulativeSize = 0;
+    if (segmentation === 'x' || segmentation === 'y') {
+      const segmentBegins = pointsInSegments.map(
```
williaster, line 57: wdyt? starts is a little more clearly the place where each segment starts

```suggestion
      const segmentStarts = pointsInSegments.map(
```

packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
+        points => points.find(p => typeof p[segmentation] === 'number')?.[segmentation],
+      );
 
-    const lineSegments = segments.map(segment => {
-      const segmentPointCount = segment.length;
-      const coords: { x: number; y: number }[] = [];
+      const first = pathElement.getPointAtLength(0);
+      const last = pathElement.getPointAtLength(totalLength);
+      const isIncreasing = last[segmentation] > first[segmentation];
+      const isBeyondSegments = isIncreasing
```
williaster, line 64: what about 

```suggestion
      const isBeyondSegmentStart = isIncreasing
```

packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
+          )
+        : segmentBegins.map(begin =>
+            typeof begin === 'undefined' ? TRUE : (xOrY: number) => xOrY <= begin,
+          );
 
-      for (let i = 0; i < segmentPointCount + sampleRate; i += sampleRate) {
-        const distance = (cumulativeSize + i) * pieceSize;
-        const point = pathElement!.getPointAtLength(distance);
-        coords.push(point);
+      let current = 0;
```
williaster, line 72:
```suggestion
      let currentSegment = 0;
```

packages/visx-shape/src/util/getSplitLineSegments.ts
```diff
@@ -1,80 +1,110 @@
...
+      for (let distance = 0; distance <= totalLength; distance += sampleRate) {
+        const sample = pathElement.getPointAtLength(distance);
+        const position = sample[segmentation];
+        while (current < numSegments - 1 && isBeyondSegments[current + 1](position)) {
```
williaster, line 76: variables are pretty well-named so this is decently easy to follow! wonder if a comment or two for this part could be useful tho to make it a little more readable, what about something like?

> find the current segment to which this sample belongs

williaster (commented):
Wow thanks for making this kick ass! love the new demo and the additional algorithms 🙌 

I had a couple minor comments, some just about readability.

kristw:
Thanks for all the comments. Addressed all.

williaster (approved):
woo!

Issue #1265, opened 2021-07-06
[@visx/xychart] Missing react-spring dependency


schillerk:
Minimal reproduction
`npx create-react-app visx-demo`

Replace App.jsx with the demo here:
https://airbnb.io/visx/docs/xychart

`Module not found: Can't resolve 'react-spring' in '/Users/kyleschiller/src/xychart-demo/node_modules/@visx/react-spring/lib/axis'`

after `npm i react-spring`, issue is resolved.

williaster:
Hey @schillerk , thanks for filing the issue 🙏  I don't think this is actually a problem, though, since we declare `react-spring` as a `peerDependency` in the `@visx/xychart` `package.json`:

https://github.com/airbnb/visx/blob/fbbaf42ca8cd56f8fd8ef53ed8799e31fbffb0b9/packages/visx-xychart/package.json#L40

The `peerDependency` indicates that package consumers need to install that dependency themselves. If you do that it should work as you noted, and here's the working sandbox that the `xychart` demo links to https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-xychart?file=/package.json:535-566

schillerk:
Got it, thanks for the clarification!

TrySpace:
I first had `'to' is not exported from react-spring`. Tried removing nodes_modules & reinstalling, no avail.

Then updated to `9.3.2`, then the error went away. However, now `react-spring/web.cjs` is not available, but lulckily I don't care about IE11

PR #1266, opened 2021-07-06
fix(xychart): Add react-spring to xychart dependencies

schillerk:
🐛 Bug Fix
Fixes [1265](https://github.com/airbnb/visx/issues/1265)

We already have `@visx/react-spring`, but some files import the actual `react-spring`, e.g.
https://github.com/airbnb/visx/blob/master/packages/visx-xychart/src/components/series/private/AnimatedBars.tsx

A better fix might be to replace `react-spring` with `@visx/react-spring` everywhere?


williaster:
#1265 is closed

PR #1268, opened 2021-07-08
[deps] Upgrade to react v17 and react-testing-library

gazcn007:
#### :boom: Breaking Changes

- Add react-testing-lib
- Add React v17:
   - Which enzyme doesn't have a good support of, mount will fail. [Thread Here](https://github.com/enzymejs/enzyme/issues/2462)
- Remove React v15

#### :rocket: Enhancements

- Improve unit tests by rewriting some of them to RTL (react-testing-library)

##### Tests to fix:
- [x] visx-annotation : Upgrade and Rewriting tests to RTL
- [x] visx-axis : Pass without fixes
- [x] visx-bounds : Rewriting tests to RTL
- [x] visx-brush : Pass without fixes
- [x] visx-chord : Pass without fixes
- [x] visx-clip-path : Pass without fixes
- [x] visx-curve : No React Dependencies, No Upgrade needed
- [x] visx-demo : Pass without fixes (no tests)
- [x] visx-drag : Rewriting tests to RTL
- [x] visx-event : No React Dependencies, No Upgrade needed
- [x] visx-geo : Pass without fixes (mount is not used, only shallow is used for enzyme, no failures)
- [x] visx-glyph : Pass without fixes (mount is not used, only shallow is used for enzyme, no failures)
- [x] visx-gradient : Rewriting tests to RTL
- [x] visx-grid : Pass without fixes
- [x] visx-group : Pass without fixes
- [x] visx-heatmap : Pass without fixes
- [x] visx-hierarchy : Pass without fixes
- [x] visx-legend : Pass without fixes
- [x] visx-marker : Pass without fixes
- [x] visx-mock-data : No React Dependencies, No Upgrade needed
- [x] visx-network : Pass without fixes
- [x] visx-pattern : Pass without fixes
- [x] visx-point : No React Dependencies, No Upgrade needed
- [x] visx-react-spring : Pass without fixes
- [x] visx-responsive : Upgrade and Rewriting tests to RTL
- [x] visx-scale :  No React Dependencies, No Upgrade needed
- [x] visx-shape : Upgrade and Rewriting tests to RTL
- [x] visx-stats : Pass without fixes
- [x] visx-text : Upgrade and Rewriting tests to RTL
- [x] visx-threshold : Pass without fixes
- [x] visx-tooltip : Pass without fixes
- [x] visx-visx : Pass without fixes (no unit tests in this package)
- [x] visx-voronoi : Pass without fixes
- [x] visx-xychart : Upgrade and Rewriting tests to RTL
- [x] visx-zoom : Pass without fixes

#### :bug: Bug Fix

-

#### :house: Internal

- TBD


packages/visx-bounds/test/withBoundingRects.test.tsx
```diff
@@ -40,41 +68,48 @@ describe('withBoundingRects()', () => {
...
+
+    // getBoundingClientRect should be called twice, once for the component, and once for its parent
+    await waitFor(() => expect(Element.prototype.getBoundingClientRect).toHaveBeenCalledTimes(2));
+
+    const RenderedComponent = await getByTestId('BoundingRectsComponent');
+    const RenderedComponentParent = await getByTestId('BoundingRectsComponentParent');
 
-    expect(Element.prototype.getBoundingClientRect).toHaveBeenCalled();
-    expect(RenderedComponent.prop('rect')).toEqual(expectedRectShape);
-    expect(RenderedComponent.prop('parentRect')).toEqual(expectedRectShape);
```
gazcn007, line 51: @williaster This is where RTL's philosophy of testing differs from Enzyme. [https://testing-library.com/docs/#what-you-should-avoid-with-testing-library](Directly checking internal props & states is a NO-NO in RTL)

williaster: ^I agree the RTL philosophy is better for things like this 💯 this is why enzyme struggles with adapters for each react release 😅

packages/visx-glyph/test/Cross.test.tsx
```diff
@@ -29,7 +29,7 @@ describe('<GlyphCross />', () => {
     shallow(<GlyphCross>{fn}</GlyphCross>);
     const args = fn.mock.calls[0][0];
     const keys = Object.keys(args);
-    expect(keys.includes('path')).toEqual(true);
+    expect(keys).toContain('path');
```
gazcn007, line 32: more idiomatic

packages/visx-grid/test/GridPolar.test.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import { scaleLinear } from '@visx/scale';
 import { GridPolar, GridAngle, GridRadial } from '../src';
-import { scaleLinear } from '../../visx-scale';
```
gazcn007, line 5: this will cause tests to fail when visx-scale in the upper folder is not built

packages/visx-responsive/test/ScaleSVG.test.tsx
```diff
@@ -9,14 +9,9 @@ describe('<ScaleSVG />', () => {
...
-        expect(n.tagName).toEqual('svg');
-        done();
-      };
-
-      mount(<ScaleSVG innerRef={refCallback} />);
-    });
+    const fakeRef = React.createRef<SVGSVGElement>();
+    const { container } = render(<ScaleSVG innerRef={fakeRef} />);
+    const SVGElement = container.querySelector('svg');
+    expect(fakeRef.current).toContainElement(SVGElement);
```
gazcn007, line 15: I think this is an improvement over the previous test in enzyme :)

packages/visx-text/test/Text.test.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
```
keithyang, line 2:
```suggestion
import { shallow } from 'enzyme';
```

packages/visx-bounds/test/withBoundingRects.test.tsx
```diff
@@ -40,41 +86,48 @@ describe('withBoundingRects()', () => {
...
+
+    // getBoundingClientRect should be called twice, once for the component, and once for its parent
+    await waitFor(() => expect(Element.prototype.getBoundingClientRect).toHaveBeenCalledTimes(2));
 
-    expect(Element.prototype.getBoundingClientRect).toHaveBeenCalled();
-    expect(RenderedComponent.prop('rect')).toEqual(expectedRectShape);
-    expect(RenderedComponent.prop('parentRect')).toEqual(expectedRectShape);
-    expect(typeof RenderedComponent.prop('getRects')).toBe('function');
+    const RenderedComponent = getByTestId('BoundingRectsComponent');
+    const RenderedComponentParent = getByTestId('BoundingRectsComponentParent');
```
williaster, line 98: TIL `getByTestId`

packages/visx-demo/package.json
```diff
@@ -87,7 +87,7 @@
     "nprogress": "^0.2.0",
     "prismjs": "^1.19.0",
     "raw-loader": "^4.0.0",
-    "react": "^16.9.0",
+    "react": "^16.9.0 || ^17.0.0-0",
```
williaster, line 90: since this one is a `dependency` (and not really intended to be installed) we could probably just use `^17.0.0` here? this would also let us test and make sure everything works with `17` in the demo site.

packages/visx-grid/test/GridPolar.test.tsx
williaster: great catch. this used to happen in several packages and was mostly cleaned up in the TS re-write but clearly missed some 🙉

packages/visx-responsive/test/ScaleSVG.test.tsx
williaster: +1

packages/visx-responsive/test/withParentSize.test.tsx
```diff
@@ -1,39 +1,27 @@
...
+describe('withParentSize', () => {
   test('it should be defined', () => {
     expect(withParentSize).toBeDefined();
   });
 
   test('it chould pass parentWidth and parentHeight props to its child', () => {
-    const Component = () => <div />;
     const HOC = withParentSize(Component);
-    const wrapper = mount(<HOC />);
+    const { getByTestId } = render(<HOC initialWidth={200} initialHeight={200} />);
```
williaster, line 22: nice way to avoid `getBoundingClientRect`, I think when we initially wrote this test it didn't take the `initialWidth/Height`

packages/visx-shape/test/Arc.test.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import mockConsole from 'jest-mock-console';
-import { shallow, mount } from 'enzyme';
-
+import { render } from '@testing-library/react';
 import { Arc } from '../src';
 import { ArcProps } from '../src/shapes/Arc';
+import '@testing-library/jest-dom';
```
williaster, line 6: not sure if this will throw a lint error for being after local imports 🤔  fine if it doesn't throw!

packages/visx-shape/test/Area.test.tsx
```diff
@@ -40,18 +42,14 @@ describe('<Area />', () => {
...
-      );
-    });
+    const fakeRef = React.createRef<SVGPathElement>();
+    const { container } = render(
+      <svg>
+        <Area data={fakeData} x={x} y={y} innerRef={fakeRef} />
+      </svg>,
+    );
+    const PathElement = container.querySelector('path');
+    expect(fakeRef.current).toContainElement(PathElement);
```
williaster, line 52: also TIL about the `@testing-library/jest-dom` selectors, nice 😎

williaster (commented):
Nice! this looks great to me, I think I only had one real suggestion which was to set `@visx/demo` to `^17.0.0`.

For the `react-spring@9` issues, do you think we should merge that in and then rebase? Looks like it has a few remaining issues but I'll try to get it to pass CI today.

williaster:
ah, one other thing we should bump is the `react` + `react-dom` deps inside all of the `visx-demo` sandboxes (e.g., [here](https://github.com/airbnb/visx/blob/master/packages/visx-demo/src/sandboxes/visx-bars/package.json#L16))

gazcn007:
visx-xychart has a huge amount of tests using mount and checking on implementation details with enzyme. I opened a separate [PR](https://github.com/gazcn007/visx/pull/1/files) that merges into this branch

packages/visx-text/test/Text.test.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { shallow } from 'enzyme';
```
williaster, line 2: I can't comment on it since it wasn't changed, but wonder if we should just remove the two `shallow` tests here since they only assert that there are not errors. then we'd be fully on RTL as well 👍

williaster (approved):
Woot! 🔥  This is awesome, thanks for doing such a great and thorough job. There are a lot of packages, and going through them all is big task, so nicely done 🙌 

I had one small remaining comment but otherwise think we could get this in! (we could also address that comment in a small followup / it's very minor)

As far as releases, there may be one more breaking change in #1305 and #1121 . So I think we should do another `2.0.0-alpha` pre-release (I think we need a `patch-release` to force a change based on how our CI works) and then merge those two and graduate to `2.0.0` asap.

packages/visx-text/test/Text.test.tsx
gazcn007: yea sure, these tests are trivial. For later iterations, I would like to convert the shallow tests to RTL as well (these are relatively easier, most of them are just renaming work), so we can fully remove enzyme out of a devDependencies.

keithyang (approved):

PR #1281, opened 2021-07-18
fix: remove @types/classnames

Janpot:
#### :rocket: Enhancements

noticed warnings during install

```
warning workspace-aggregator-d56295b5-e1bb-4287-aa1e-e72f5692b6cf > docs > @visx/xychart > @types/classnames@2.3.1: This is a stub types definition. classnames provides its own type definitions, so you do not need this installed.
```

Looks like `classnames@>=2.3` comes with types


gazcn007 (approved):
👍

williaster:
thanks for the fix @Janpot ! 🙏

PR #1286, opened 2021-07-21
fix(brush, drag): call start callbacks consistently, support mobile

williaster:
#### :bug: Bug Fix

This should fix both 
- https://github.com/airbnb/visx/issues/1283 and
- https://github.com/airbnb/visx/issues/845

Demo of the start/end callback invocations for move/resize/fresh brush:

start-end-brush

@gazcn007 @kristw 
cc @JoeVanGundy

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -140,13 +140,10 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
   }
 
   handleWindowPointerUp = () => {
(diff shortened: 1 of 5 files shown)


packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -256,14 +263,6 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
           extent,
         };
 
-        if (onBrushEnd) {
```
williaster, line 259: these shouldn't have been called in the move handler

kristw (approved):

PR #1294, opened 2021-07-25
new(xychart): distinguish nearestDatum in renderGlyph

Janpot:
#### :rocket: Enhancements

Tooltip: Make it possible to distinguish between `nearestDatum` and other series in `renderGlyph`. Also correct the type of `renderGlyph`.

Alternatively we could go for different mechanisms, like passing the `nearestDatum` or `nearestDatumKey` so consumers can compare themselves with the `key` prop, or something like a `type: 'datum' | 'series'`. Also not sure if the naming of `isNearestDatum` is in the spirit of the library. Suggestions welcome.

_edit: Actually, looks like users are already able to achieve this behaviour by using the `TooltipContext` in their custom tooltip glyph, so I'm fine with discarding this PR altogether as well._


williaster (approved):
this all seems reasonable to me. while I think you're right that consumers could do this themselves using `Context`, this would make it significantly easier so I'm cool with merging it. 

re naming, for the general tooltip data we use `nearestDatum`, so I think `isNearestDatum` is reasonable for this case. we're in a bit of a limbo with releases as we make some breaking changes (react-spring@9 + react@17), so this may come out in an alpha for the time being.

williaster:
gonna go ahead and merge this to get out with a couple other xychart enhancements

PR #1301, opened 2021-08-02
deps: bump all nimbus config

williaster:
#### :house: Internal

WIP

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,12 +54,12 @@
     "type:workspaces": "nimbus typescript --build"
   },
   "devDependencies": {
-    "@airbnb/config-babel": "^2.1.3",
-    "@airbnb/config-eslint": "^2.1.3",
-    "@airbnb/config-jest": "^2.1.3",
-    "@airbnb/config-prettier": "^2.0.4",
-    "@airbnb/config-typescript": "^2.1.2",
-    "@airbnb/nimbus": "^2.1.3",
+    "@airbnb/config-babel": "^3.1.0",
+    "@airbnb/config-eslint": "^3.1.0",
+    "@airbnb/config-jest": "^3.0.2",
+    "@airbnb/config-prettier": "^3.1.0",
+    "@airbnb/config-typescript": "^3.0.1",
+    "@airbnb/nimbus": "^3.1.4",
     "@octokit/rest": "18.1.0",
     "@types/enzyme": "^3.10.3",
     "@types/jest": "^24.0.18",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,102 +2,103 @@
 # yarn lockfile v1
 
 
-"@airbnb/config-babel@^2.1.3":
-  version "2.2.4"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -74,9 +74,10 @@
     "@testing-library/user-event": "^11.0.1",
     "chalk": "4.1.0",
     "coveralls": "^3.0.6",
-    "enzyme": "^3.10.0",
```
gazcn007, line 77: nit: the ending space is unnecessary.
I will open a PR soon to convert the rest of enzyme tests into RTL. So we can get rid of enzyme from dev dependency : )

williaster: sweeeeeeet re enzyme! that will be amazing 🦾

packages/visx-chord/src/Ribbon.tsx
```diff
@@ -43,13 +43,13 @@ export default function Ribbon({
...
 }: Omit<React.SVGProps<SVGPathElement>, keyof RibbonProps> & RibbonProps) {
-  const ribbon = d3ribbon<any, Chord, ChordSubgroup>();
+  const ribbon = d3ribbon<unknown, Chord, ChordSubgroup>();
   if (source) ribbon.source(source);
   if (target) ribbon.target(target);
   if (radius) setNumberOrNumberAccessor(ribbon.radius, radius);
   if (startAngle) setNumberOrNumberAccessor(ribbon.startAngle, startAngle);
   if (endAngle) setNumberOrNumberAccessor(ribbon.endAngle, endAngle);
-  const path = (ribbon(chord) as unknown) as string | null;
+  const path = ribbon(chord) as unknown as string | null;
```
etr2460, line 52: do we still need `as unknown`?

packages/visx-demo/src/pages/linktypes.tsx
```diff
@@ -1,26 +1,26 @@
...
-        bottom: 40,
-      }}
-      packageJson={packageJson}
-    >
-      {LinkTypesSource}
-    </Show>
-  );
-};
-export default LinkTypesPage;
+import React from 'react';
```
etr2460, line 1: i don't understand this diff
image

packages/visx-scale/test/scaleRadial.test.ts
```diff
@@ -17,7 +17,8 @@ describe('sclaeRadial()', () => {
 
   it('set unknown', () => {
     const scale = scaleRadial({ domain: [0, 10], unknown: 'green' });
-    expect(scale('sandwich' as any)).toEqual('green');
+    // coerce to make TS happy
+    expect(scale('sandwich' as unknown as number)).toEqual('green');
```
etr2460, line 21: wait, what?

`scale` seems to only expect a number. you're passing in `sandwich` and forcing typescript to be ok with it. Shouldn't this test be deleted since TS makes sure this can never happen?

Or, if you want to test the unknown values, can we pass in a number that's not in the domain to test that? either `-1` or `11`?

etr2460 (approved):
a few comments, but stamping to unblock.

one thing i've found useful for these types of PRs in the future is to have 3 different commits added to the PR. The first one is all the config/package.json changes. The second is after running yarn install and autofixing everything. Then the third commit contains all manual changes. That way a reviewer can look at the first and third commits only for review, and ignore the second. food for thought!

PR #1305, opened 2021-08-04
new(zoom): add pinch and zoom support

tonyneel923:
#### :boom: Breaking Changes
- zoom.containerRef needs to be added to whatever element needs zoom functionality.
- props.transformMatrix was changed to => initialTransformMatrix
- passive/style/className were removed

#### :rocket: Enhancements
Zoom now supports mobile pinch.

#### :memo: Documentation
I consolidated the handlers all into react-use-gesture. A user can opt out by not adding the containerRef so it seems like passive was no longer necessary.

In order for mobile zoom to work fully users must add ```style={{ touchAction: 'none' }}``` to the element in order for the browser to allow js to handle pinch and zoom.

Also I had to use the newest version of use-gesture which is technically in beta because of bugs in version 9. By the author's own admission it is ready for production so I do not see an issue with it.

Discussion which prompted this PR: https://github.com/airbnb/visx/discussions/1292

### DIFF
--- a/packages/visx-zoom/src/Zoom.tsx
+++ b/packages/visx-zoom/src/Zoom.tsx
@@ -39,7 +39,7 @@ export type ZoomProps<ElementType> = {
     params: Parameters<
       Handler<'pinch', TouchEvent | PointerEvent | WheelEvent | WebKitGestureEvent>
     >[0],
-  ) => ScaleSignature;
+  ) => Scale;
   /** Minimum x scale value for transform. */
   scaleXMin?: number;
   /** Maximum x scale value for transform. */
@@ -95,7 +95,10 @@ function Zoom<ElementType extends Element>({
   },
   wheelDelta = (event: React.WheelEvent | WheelEvent) =>
     -event.deltaY > 0 ? { scaleX: 1.1, scaleY: 1.1 } : { scaleX: 0.9, scaleY: 0.9 },
-  pinchDelta,
+  pinchDelta = ({ offset: [s], lastOffset: [lastS] }) => ({
+    scaleX: s - lastS < 0 ? 0.9 : 1.1,
+    scaleY: s - lastS < 0 ? 0.9 : 1.1,
+  }),
   width,
   height,
   constrain,
@@ -257,20 +260,15 @@ function Zoom<ElementType extends Element>({
 
   const handlePinch: UserHandlers['onPinch'] = useCallback(
     state => {
-      if (pinchDelta) {
-        return scale(pinchDelta(state));
-      }
-
       const {
         origin: [ox, oy],
-        offset: [s],
-        lastOffset: [lastS],
       } = state;
       if (containerRef.current) {
         const { top, left } = containerRef.current.getBoundingClientRect();
+        const { scaleX, scaleY } = pinchDelta(state);
         scale({
-          scaleX: s - lastS < 0 ? 0.9 : 1.1,
-          scaleY: s - lastS < 0 ? 0.9 : 1.1,
+          scaleX,
+          scaleY,
           point: { x: ox - left, y: oy - top },
         });
       }
--- a/packages/visx-zoom/src/Zoom.tsx
+++ b/packages/visx-zoom/src/Zoom.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { useState, useRef } from 'react';
 import { localPoint } from '@visx/event';
+import { useGesture, WebKitGestureEvent, Handler } from '@use-gesture/react';
 import {
   composeMatrices,
   inverseMatrix,
@@ -11,7 +12,7 @@ import {
 } from './util/matrix';
 import { TransformMatrix, Point, Translate, Scale, ScaleSignature, ProvidedZoom } from './types';
 
-export type ZoomProps = {
+export type ZoomProps<ElementType> = {
   /** Width of the zoom container. */
   width: number;
   /** Height of the zoom container. */
@@ -25,6 +26,20 @@ export type ZoomProps = {
    * Scale factors greater than 1 will increase (zoom in), less than 1 will descrease (zoom out).
    */
   wheelDelta?: (event: React.WheelEvent | WheelEvent) => Scale;
+  /**
+   * ```js
+   *  pinchDelta(state)
+   * ```
+   *
+   * A function that returns { scaleX, scaleY, point } factors to scale the matrix by.
+   * Scale factors greater than 1 will increase (zoom in), less than 1 will descrease (zoom out), the point is used to find where to zoom.
+   * The state parameter is from react-use-gestures onPinch handler
+   */
+  pinchDelta?: (
+    params: Parameters<
+      Handler<'pinch', TouchEvent | PointerEvent | WheelEvent | WebKitGestureEvent>
+    >[0],
+  ) => ScaleSignature;
   /** Minimum x scale value for transform. */
   scaleXMin?: number;
   /** Maximum x scale value for transform. */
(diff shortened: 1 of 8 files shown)


tonyneel923:
It seems the build fails because it tries to download from my companies mirror registry. I only have a work laptop so I cannot fix that unless I remove the yarn lock changes completely but that would make it not in sync.

Is it possible for someone to fix that minor issue for me? Anything else I can do.

williaster:
Wow thanks for the great contribution @tonyneel923 !  ❤️ 

I will try to take a look at this tomorrow (sorry it's a bigger change so want to look at it more closely!). I just pulled your branch to update the lock file and it allowed me to push to your branch. Sorry if that's not what you meant, my only other thought was to add a `.npmrc` file in the monorepo root which specified the npm or yarn registry so this wouldn't be a problem in the future (thanks for pointing it out! I may add that separately in another PR)

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,238 @@ type ZoomState = {
...
+    event.preventDefault();
     const point = localPoint(event) || undefined;
     const { scaleX, scaleY } = wheelDelta!(event);
-    this.scale({ scaleX, scaleY, point });
+    scale({ scaleX, scaleY, point });
   };
 
-  toString = () => {
-    const { transformMatrix } = this.state;
+  const handlePinch: Parameters<typeof useGesture>[0]['onPinch'] = state => {
```
williaster, line 230: same thought on possibly using `UserHandlers['onPinch']` here

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,238 @@ type ZoomState = {
...
+          dragEnd();
+        } else if (!(event instanceof KeyboardEvent)) {
+          dragMove(event);
+        }
+      },
+      onDragEnd: dragEnd,
+      onPinch: handlePinch,
+      onWheel: ({ event }) => handleWheel(event),
+    },
+    { target: containerRef, eventOptions: { passive: false }, drag: { filterTaps: true } },
```
williaster, line 285: options seem legit to me

williaster (commented):
Overall this is looking killer! Just did a quick first pass and will check it out again tomorrow. 

I think in the breaking changes (in addition to `zoom.containerRef` that you mentioned) we should also note that 
- `props.transformMatrix` was changed to => `initialTransformMatrix`
- `passive`/`style`/`className` were removed

tonyneel923:
> Wow thanks for the great contribution @tonyneel923 ! ❤️
> 
> I will try to take a look at this tomorrow (sorry it's a bigger change so want to look at it more closely!). I just pulled your branch to update the lock file and it allowed me to push to your branch. Sorry if that's not what you meant, my only other thought was to add a `.npmrc` file in the monorepo root which specified the npm or yarn registry so this wouldn't be a problem in the future (thanks for pointing it out! I may add that separately in another PR)

I am not sure why I did not think of a .npmrc file. We will go with I was tired. Thanks for the feedback on the pr. I will address it today.

tonyneel923:
I have updated the pr with your requested changes. I also changed pinchDelta to match wheelDelta as I realized the way it was would not work since you would have access to zoom in the callback. I tested on example and it worked perfectly.

tonyneel923:
Final note is I changed the test in zoom to not fail but I am not sure why it failed at finding one div because it definitely renders the contents. I do not have much experience with enzyme though and generally use rtl.

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,267 @@ type ZoomState = {
...
-
-  containerRef: HTMLDivElement | null = null;
-
-  startPoint: Point | undefined = undefined;
+function Zoom<ElementType extends Element>({
+  scaleXMin = 0,
+  scaleXMax = Infinity,
+  scaleYMin = 0,
+  scaleYMax = Infinity,
+  initialTransformMatrix = {
```
williaster, line 88: just looking for optimization opportunities, could we pull this out into a constant `defaultInitialTransformMatrix`? I think this way of defining it will be a new object each render (which doesn't matter for `useRef`/`useState` but I think will break memoization of a callback below)

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,267 @@ type ZoomState = {
...
+    scaleX: 1,
+    scaleY: 1,
+    translateX: 0,
+    translateY: 0,
+    skewX: 0,
+    skewY: 0,
+  },
+  wheelDelta = (event: React.WheelEvent | WheelEvent) =>
```
williaster, line 96: ditto on pulling out this as a constant

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,267 @@ type ZoomState = {
...
+    -event.deltaY > 0 ? { scaleX: 1.1, scaleY: 1.1 } : { scaleX: 0.9, scaleY: 0.9 },
+  pinchDelta = ({ offset: [s], lastOffset: [lastS] }) => ({
```
williaster, line 98: ditto on pulling out this as a constant, sorry I didn't catch these earlier. works okay as a class component defaultProps but less well for a function component.

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -79,229 +80,267 @@ type ZoomState = {
...
-    const { scaleX, scaleY } = wheelDelta!(event);
-    this.scale({ scaleX, scaleY, point });
-  };
+  const handlePinch: UserHandlers['onPinch'] = useCallback(
+    state => {
+      const {
+        origin: [ox, oy],
+      } = state;
+      if (containerRef.current) {
+        const { top, left } = containerRef.current.getBoundingClientRect();
```
williaster, line 267: not sure how this call will be for perf or if it's really avoidable, but we can iterate in the future if need be.

packages/visx-zoom/test/Zoom.test.tsx
```diff
@@ -33,8 +33,7 @@ describe('<Zoom />', () => {
       </Zoom>,
     );
 
-    expect(wrapper.find('div')).toHaveLength(1);
```
williaster, line 36: I think previously when `passive=false` (default), `Zoom` wrapped `children` in a `div` so that's why this is no longer there 👍

williaster (commented):
Woot! @tonyneel923 this looks great to me 🔥  I pulled it locally and tested on my phone 💯 

I had a couple super minor comments but then I think it's ready to land! Thanks again for the great addition 🙏  We are coordinating some breaking changes with `react@17`, so this may land first as an alpha but will be out soon.

tonyneel923:
I added in the changes you requested and I found a way to keep from running getBoundingClientRect all the time using the built in memo parameter in onPinch.

williaster (approved):
Nice approach on the memoization! working on a release issue then will merge this 🎉

boy51:
Hey I appreciate this change! I'm using zoom in a very tight component and it would be really helpful we could be more explicit about breaking changes. For now, I don't feel like I can implement this for fear of introducing bugs.

For example, I used `passive` because we are handling mousewheel manually. How am I supposed to migrate? I see a brief mention of omitting the `containerRef`. What does it do for me in the first place?

Also why the change to `initialTransformMatrix`? Is it purely semantic? Should I just rename `transformMatrix`, or does it change the behaviour? If not, why is a breaking change needed for this?`

Also there are undocumented breakign changes, the interface `ProvidedZoom` now requires generic.

If there is documentation surrounding this please let me know!

PR #1310, opened 2021-08-11
fix(scripts/performRelease): fix alpha release logic

williaster:
#### :house: Internal

In trying to stack some breaking changes into the `2.0.0-alpha.0` release, CI released a `3.0.0-alpha` (npm `@ next`) instead of a `2.0.1-alpha`. So this PR:

- Updates packages versions/dependencies from `@visx/*@3.0.0-alpha.0 => @visx/*@2.0.0-alpha.1` (done with a manual `lerna version`)
  - Note: we **cannot** `unpublish` the  `3.0.0-alpha.0` packages because people have dependencies on it. I may deprecate it, and I deleted the `v3.0.0-alpha.0` github tag.
- Updates an issue with the `performRelease` script. Here's an overview of the issue:
  - When we merged the breaking `react-spring@9` with tags `alpha-release` and `major-release`, it updated from `1.XX.X` to `2.0.0-alpha.0` as expected 
  - When we then merged the breaking `react@17` with tags `alpha-release` and `patch-release` I expected it to publish `2.0.1-alpha.0`. Instead it gathered all PRs since the last **`non-alpha`** release (i.e., since `1.XX.X`) and inspected the tags to determine what the **next** version would be. Because it found the `react-spring@9` `major-release` tag, it then published another major version bump to `3.0.0-alpha`
  - The fix then is to only consider the PRs since the last release (regardless of alpha/non-alpha) to determine what the next release should be. If we don't use this updated logic, we would have to go **remove** `major-release` from previous breaking change PRs so they didn't yield a second major version bump.
 
If the planned `2.0.1` non-alpha release fails, we can alway go to `3.0.0` but the script fix was needed anyway. 

@kristw @gazcn007


williaster:
^adding release tags to this PR to correct npm `@next` and test that another `2.0.1-alpha.0` release will work after the `3.0.0-alpha.0` was published 😬

kristw (approved):

PR #1311, opened 2021-08-13
new(wordcloud): add @visx/wordcloud

craciuncezar:
#### :rocket: Enhancements

- Adding a new word cloud visualization
- Started from this issue [wordcloud-proposal](https://github.com/airbnb/visx/issues/630) and extracted the layout logic into a custom hook [useWordcloud](https://github.com/airbnb/visx/blob/a0d36e6d1292a9181f847fc1be02278332a5288e/packages/visx-wordcloud/src/useWordcloud.ts)

achimedean
rectangular and with rotation


packages/visx-wordcloud/src/useWordcloud.ts
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * @default 0
+   */
+  height?: number;
+  /**
+   * Sets the words array.
+   *
+   * @default []
+   */
+  words?: Datum[];
```
kristw, line 26: Should `undefined` be allowed for this field?

packages/visx-wordcloud/src/useWordcloud.ts
```diff
@@ -0,0 +1,132 @@
+import { useEffect, useState } from 'react';
+import d3Cloud from 'd3-cloud';
+
+export interface BaseDatum {
```
kristw, line 4: We could extract `BaseDatum` and `WordcloudConfig` types into `./types.ts` so this file can export `useWordcloud` as default export, making it consistent with `Wordcloud.tsx`.

packages/visx-wordcloud/src/Wordcloud.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { BaseDatum, useWordcloud, WordcloudConfig } from './useWordcloud';
+
+export interface WordcloudProps<Datum extends BaseDatum> extends WordcloudConfig<Datum> {
+  children: (words: d3Cloud.Word[]) => React.ReactNode;
+}
+
+export default function Wordcloud<Datum extends BaseDatum>(props: WordcloudProps<Datum>) {
+  const { children, ...wordcloudConfig } = props;
+  const words = useWordcloud(wordcloudConfig);
+
```
kristw, line 13: Can make it a bit more readable with.

```ts
const { width, height } = wordcloudConfig;
```

and refer to `width` and `height` afterwards.

packages/visx-wordcloud/src/useWordcloud.ts
```diff
@@ -0,0 +1,132 @@
...
+  padding,
+  random,
+  rotate,
+  spiral,
+  words,
+}: WordcloudConfig<Datum>) {
+  const [cloudWords, setCloudWords] = useState<d3Cloud.Word[]>([]);
+
+  useEffect(() => {
+    const layout = d3Cloud<Datum>();
```
kristw, line 98: Maybe we don't have to do any of these layout thing at all if `width === 0 || height === 0`?

```ts
if (width === 0 || height === 0) {
  return;
}
```

packages/visx-wordcloud/src/useWordcloud.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (width !== undefined && height !== undefined) layout.size([width, height]);
```
kristw, line 100: a safer check for `x !== undefined`
is `typeof x !== 'undefined'`
because `undefined` can be overridden.

kristw (commented):
Thanks for the contribution. I left some comments.
Please make sure the unit tests are passed as well.

packages/visx-wordcloud/src/useWordcloud.ts
craciuncezar: Wow, I had no idea that undefined can be overridden 🤯. There is even a [no-undefined](https://eslint.org/docs/rules/no-undefined) eslint rule that prevents the usage of undefined because of this 😅, will fix it, It's better to be safe than sorry 👍

packages/visx-wordcloud/.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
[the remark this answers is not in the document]
craciuncezar: All the other packages have the `.npmrc` file with `package-lock=false` and I just assumed that is a tooling/ci requirement.

packages/visx-wordcloud/src/useWordcloud.ts
craciuncezar: Yeah having this field `undefined` makes no sense, I just kept the API parity with `d3-cloud`. I'll make this required, and on the same note, I think the width and height should be required as well, WDYT?

packages/visx-wordcloud/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "svg"
+  ],
+  "author": "Cezar Craciun @_cezarcraciun",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/airbnb/visx/issues"
+  },
+  "homepage": "https://github.com/airbnb/visx#readme",
+  "peerDependencies": {
+    "react": "^16.8.0-0",
```
williaster, line 32: we **just** landed `react@17` support, so maybe can open this up to the following? 

```suggestion
    "react": "^16.8.0-0 || ^17.0.0-0",
```

packages/visx-wordcloud/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "react-dom": "^16.8.0-0"
```
williaster, line 33: I think this isn't necessary because there aren't any calls to actually render to the DOM (just in the demo so only the demo needs it)

packages/visx-wordcloud/src/useWordcloud.ts
williaster: requiring `width/height` makes sense to me since you'd get no output otherwise. some lib consumers aren't familiar with the `d3` intricacies and have been confused by things like that in the past)

packages/visx-wordcloud/src/useWordcloud.ts
williaster: oh the intricacies of JS 😅

packages/visx-wordcloud/test/Wordcloud.test.tsx
```diff
@@ -0,0 +1,91 @@
+import { mount } from 'enzyme';
```
williaster, line 1: thanks for writing tests!

just noting that we should likely test this with `react@17`. in that upgrade (#1268) we had to swap a few tests for `react-testing-library` that broke since `enzyme` still doesn't support `react@17`.

cc @gazcn007 do you see anything here you think would be problematic with `react@17`?

williaster (commented):
wow thanks for the great contribution @craciuncezar 🙌 !!!

this looks really great, I had just a few comments which hopefully won't be too tricky to address. excited for this to land 🔜 !

packages/visx-wordcloud/test/Wordcloud.test.tsx
williaster: I think this actually is already maybe failing CI, might be easiest to use `react-testing-library` – hopefully just swapping `mount` calls `render`

packages/visx-wordcloud/src/useWordcloud.ts
```diff
@@ -0,0 +1,132 @@
...
+  fontWeight,
+  padding,
+  random,
+  rotate,
+  spiral,
+  words,
+}: WordcloudConfig<Datum>) {
+  const [cloudWords, setCloudWords] = useState<d3Cloud.Word[]>([]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
williaster: ^I had a similar question, I think it would for sure have called the wordcloud layout for each render with the inline function. In the example specifically I think that would only have happened with a size change of the parent, but probably best to implement the example with perf in mind.

packages/visx-wordcloud/src/Wordcloud.tsx
```diff
@@ -0,0 +1,25 @@
...
+export default function Wordcloud<Datum extends BaseDatum>(props: WordcloudProps<Datum>) {
+  const { children, ...wordcloudConfig } = props;
+  const { width, height } = wordcloudConfig;
+  const words = useWordcloud(wordcloudConfig);
+
+  if (width === 0 || height === 0) return null;
+
+  return (
+    <svg width={wordcloudConfig.width} height={wordcloudConfig.height}>
```
williaster, line 19: super minor but can update to `width`/`height` since you've already pulled them out

```suggestion
    <svg width={width} height=.height}>
```

packages/visx-wordcloud/src/Wordcloud.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <Group left={wordcloudConfig.width / 2} top={wordcloudConfig.height / 2}>
```
williaster, line 20: same here 

```suggestion
      <Group left={width / 2} top={height / 2}>
```

williaster (commented):
Woohoo looking great @craciuncezar ! Thanks for all of the iteration on this. 

I had 2 more minor comments but otherwise lgtm!

williaster:
New happo visual example looks good 📈  ✨  
image

packages/visx-wordcloud/src/useWordcloud.ts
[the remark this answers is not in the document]
mitchellwarr: purely for curiousity sake, do you think theres a way to change it inside the hook not to be concerned about constant rerender? i think this is unique across most of the visx packages right?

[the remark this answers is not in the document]
craciuncezar: I'm open to suggestions however I don't see how we can avoid this 🤔, I know it's not ideal but, as usual in React, it's up to the consumer to memoize or wrap fn in `useCallback`.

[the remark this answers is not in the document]
williaster: Yeah I think unfortunately there's probably not a clean way to remove this, as @craciuncezar said it's common in React for memoization logic to be pushed to the consumer. We have this same requirement in other layout code in `visx`, I would guess that many consumers don't optimize their usage, but if they don't see perf issues it's probably okay (wordcloud perf may present more of an issue since it's a more complex algorithm)

williaster (approved):
Thanks again for all the work here @craciuncezar 🙌

williaster:
I don't think we've released _**new**_ packages with CI before so it might fail but ... yolo

craciuncezar:
> Thanks again for all the work here @craciuncezar 🙌

I'm glad I could contribute to this awesome lib, thanks for all the work 🚀

PR #1314, opened 2021-08-16
new(demo/geo-albers/usa): use all 4 colors in the map

kenfehling:
#### :memo: Documentation
Improve the USA map demo by making use of all 4 colors already defined in the example, rather than only the first two.


williaster (approved):
thanks for the contribution @kenfehling! lgtm 🙌

williaster:
posting the happo diff 
image

PR #1346, opened 2021-09-20
Allow specifying zIndex for tooltip portals

kangaechigai:
#### :rocket: Enhancements

- `@visx/tooltip`:
  - `useTooltipInPortal` accepts a `zIndex` option to pass through to the `Portal`. The `TooltipInPortal` component it returns also accepts a `zIndex` prop to override the option.
  - `Portal` accepts `string` for `zIndex` to allow for the use of [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*).
- `@visx/xychart`:
  - `Tooltip` accepts a `zIndex` prop to pass through to the `Portal`. Fixes https://github.com/airbnb/visx/issues/1333.


packages/visx-tooltip/test/useTooltipInPortal.test.tsx
```diff
@@ -1,7 +1,46 @@
...
+    const wrapper = shallow(<TooltipWithZIndex zIndexOption={1} />, {
+      disableLifecycleMethods: true,
+    }).dive();
+    const zIndex = wrapper.find("Portal").prop("zIndex");
+    expect(zIndex).toEqual(1);
+  });
+
+  it("should pass zIndex prop from component to Portal", () => {
+    const wrapper = shallow(
+      <TooltipWithZIndex zIndexOption={1} zIndexProp="var(--tooltip-zindex)" />,
```
Blakesalot, line 38: @kangaechigai , is the setting for `zIndexOption` needed here since you're setting the `zIndexProp`?

kangaechigai: Not _needed_, but I wanted to illustrate that the prop overrides the option.

kangaechigai:
@williaster Thanks for the review (and for setting a good example of _not_ reviewing while you were on vacation)! I resolved the merge conflict. I probably won't have bandwidth anytime soon to refactor tests, so hopefully the current enzyme ones are okay for now. (I might open another PR in the next couple weeks with some potential improvements to `TooltipWithBounds` viewport detection & will try react testing library there.)

PR #1369, opened 2021-11-08
Update XYChart to handle partial dimensions

kyythane:
Our charts require that we make updates to the chart's margin after the initial layout. Currently, that can be a little awkward as `XYChart` resets the margin on each render. This change 1) allows disabling `XYChart` from updating the `margin` by passing in `null` and 2) updates `useDimensions` so it can handle partial objects, so you can update only 1 dimension or margin without needing to keep track of the rest.

#### :boom: Breaking Changes

-

#### :rocket: Enhancements

- DataProvider now takes a Partial Dimensions object.

#### :memo: Documentation

-

#### :bug: Bug Fix

-

#### :house: Internal

-


kyythane:
If the merge commits pose an issue, I can reopen this PR off of a clean copy of the repo.

packages/visx-xychart/src/components/XYChart.tsx
```diff
@@ -168,7 +167,23 @@ export default function XYChart<
...
+}
+
+function XYChartInternal(props: {
+  captureEvents: boolean;
+  accessibilityLabel: string;
+  children: React.ReactNode;
+}) {
+  const { captureEvents, accessibilityLabel, children } = props;
+  const eventEmitters = useEventEmitters({ source: XYCHART_EVENT_SOURCE });
+  const { width, height, margin } = useContext(DataContext);
```
kyythane, line 184: Created an internal component so I could read `width`, `height`, and `margin` off of the `DataContext`. Doing that in one component felt a little convoluted? 

This also moved `useEventEmitters`. Is there an issue if `useEventEmitters` is called after `useEventHandlers`?

kyythane: I can't check, but I _believe_ this extra component is what is causing happo to fail?

packages/visx-xychart/src/hooks/useDimensions.ts
```diff
@@ -7,29 +7,45 @@ const DEFAULT_DIMS = {
...
-): [Dimensions, (dims: Dimensions) => void] {
+  initialDims: PartialDimensions = {},
+): [Dimensions, (dims: PartialDimensions) => void] {
   const [dimensions, privateSetDimensions] = useState<Dimensions>({
-    width: initialDims?.width == null ? DEFAULT_DIMS.width : initialDims.width,
-    height: initialDims?.height == null ? DEFAULT_DIMS.height : initialDims.height,
-    margin: initialDims?.margin == null ? DEFAULT_DIMS.margin : initialDims.margin,
+    // using || instead of ?? to include 0 as falsy
+    width: initialDims.width || DEFAULT_DIMS.width,
+    height: initialDims.height || DEFAULT_DIMS.height,
```
kyythane, line 20: With this change, I wonder if `DEFAULT_DIMS` should be updated to match the default in XYChart? If so, I think I can remove the `null` from the margin type and maintain current behavior.

kyythane:
@williaster quick ping, sorry if you're busy. Curious about your thoughts on this PR

packages/visx-xychart/src/components/XYChart.tsx
```diff
@@ -34,7 +34,7 @@ export type XYChartProps<
   /** Total height of the desired chart svg, including margin. */
   height?: number;
   /** Margin to apply around the outside. */
-  margin?: Margin;
+  margin?: Margin | null;
```
gazcn007, line 37: I don't think you need the null here.
You can consider converting it to `Partial<Margin>` to make it consistent?

PR #1370, opened 2021-11-09
fix(responsive): fix shallow equality check on default array value for ParentSize

dylanmoz:
#### :boom: Breaking Changes

-

#### :rocket: Enhancements

-

#### :memo: Documentation

-

#### :bug: Bug Fix

I noticed that `debounceTime` wasn't being respected in ParentSize; the children function was being called many times a second. It ended up being because of the default value for `ignoreDimensions` triggering `useMemo` to reinstantiate the `resize` function, which makes the `useEffect` execute.

I also noticed that `children` always get's called initially with `width` 0. I changed that so a layout does not render until an initial width has been measured. We can remove this part of the PR if need be, but it will require users to add a `width === 0 ? null : renderContent()` type check anytime they use ParentSize to avoid rendering with that "incorrect" initial width.

#### :house: Internal

-

### DIFF
--- a/packages/visx-responsive/src/components/ParentSize.tsx
+++ b/packages/visx-responsive/src/components/ParentSize.tsx
@@ -45,20 +45,19 @@ export default function ParentSize({
   const target = useRef<HTMLDivElement | null>(null);
   const animationFrameID = useRef(0);
 
-  const [state, setState] = useState<ParentSizeState | null>(null);
+  const [state, setState] = useState<ParentSizeState>({
+    width: 0,
+    height: 0,
+    top: 0,
+    left: 0,
+  });
 
   const resize = useMemo(() => {
     const normalized = Array.isArray(ignoreDimensions) ? ignoreDimensions : [ignoreDimensions];
 
     return debounce(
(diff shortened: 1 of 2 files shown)


dylanmoz:
@williaster is it possible for me to see the Happo changes? If not just let me know what this PR is changing

williaster:
Hey @dylanmoz thanks for digging into the debounce 🙌  hook triggers can be tricky, I think the fix for pulling out the array as a constant makes sense 👍 

We can't open up the happo diffs unfortunately but this is what it's showing (one is good and one is bad 😅 )
image

I thinkkkk we should probably **not** make the change to prevent rendering when `width === 0` as it would be a breaking change and might explain the `xychart` happo diff 😞 We do have many checks throughout the demo code/examples currently to make sure `width > 0` which reflects the need to do that, but I agree it'd be great **not** to have to do that. I think it might be simpler to fix the debounce issue here and we could pursue a separate fix for that if needbe.

dylanmoz:
@williaster updated :)

williaster (approved):
lgtm, thanks again @dylanmoz ! 🙏

PR #1371, opened 2021-11-10
Update marker.html

AryanBeezadhur:


williaster:
Hey @AryanBeezadhur 👋  thanks for the PR! turns out we don't modify the html for the project site directly. Could you make these changes in the Readme.md files [here](https://github.com/airbnb/visx/blob/42d3213fe5286388fc456eade9b3874d096616aa/packages/visx-marker/Readme.md) instead? Then when we merge it will deploy to html :)

AryanBeezadhur:
@williaster reopened #1372

PR #1379, opened 2021-11-23
new(drag): Add restrictToPath as a Drag parameter

valtism:
#### :rocket: Enhancements

Allow the user to pass an SVGGeometryElement to restrict the drag area
to following the path of an SVG.

Useful for constraining the drag of an object to a curved line.


valtism:
@williaster Hardly had the last drag enhancement merged when I realised I needed another way to constrain my drag. This PR adds a way to constrain drag to an SVGGeometry like `path`. Simply constraining on `onDragMove` leads to jitter like this:

https://user-images.githubusercontent.com/1286001/142977510-c870dbe3-caed-4fe0-88a9-a2ca2df1a0b4.mov

With the constraint applied, the drag becomes much smoother.

https://user-images.githubusercontent.com/1286001/142977548-2366d189-ceff-4d57-8cb5-191468f15f82.mov

This should work even if the path is nested deeply within the SVG, as it applies a transform based on the parent SVG it is relative to.

valtism:
Unsure of what the issues with Happo are.

Also, only now realised that I totally don't need to even use these `restrict` properties. As long as I am updating the underlying data in React, I can just use `<Drag>` without applying `dx` and `dy` to the transform and not get any of the jitter from React and the transform fighting. 🤦

Oh well, I think this is still a useful feature, especially if only making changes with `onDragEnd`.

packages/visx-drag/src/util/useSamplesAlongPath.ts
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from 'react';
+
+function getSamples(restrictToPath: SVGGeometryElement, transform?: DOMMatrix, precision = 1) {
+  if (!restrictToPath) return [];
+  const samples = [];
+  const pathLength = restrictToPath.getTotalLength();
+  for (let sampleLength = 0; sampleLength <= pathLength; sampleLength += precision) {
+    const sample = restrictToPath.getPointAtLength(sampleLength);
```
williaster, line 8: hmm, looks like the returned `DOMPoint` is [not supported in IE](https://developer.mozilla.org/en-US/docs/Web/API/DOMPointReadOnly), tho `.getPointAtLength` supposedly is [for `SVGPathElement`s](https://developer.mozilla.org/en-US/docs/Web/API/SVGGeometryElement/getPointAtLength) (not `SVGGeometryElement` generally) and I'm pretty sure we use it [elsewhere](https://github.com/airbnb/visx/blob/1ee35ff40b8e5f1de45bc613ce8e58770e5ff6aa/packages/visx-shape/src/util/getSplitLineSegments.ts#L61). also IE end of life is in June 2022 so maybe no need to worry about it 🎉

williaster (approved):
Hey @valtism 👋  sorry for such a long time getting this in. Glad you found another workaround so weren't blocked by it, but agree it's useful so happy to land it. I had one small docs nit but can commit it my self with the UI if you're over it!

As always, thanks for the great contribution ❤️

williaster:
also, no happo problems – sometimes with chrome version changes we get some spurious diffs because font rendering looks different or something.

williaster:
thanks again!

PR #1383, opened 2021-11-26
Annotation label <foreignobject> render

valtism:
#### :rocket: Enhancements

Add option to pass in an element to render as a label as a child.
Uses <foreignObject> to allow user to pass in an HTML element.

Related to this discussion: https://github.com/airbnb/visx/issues/1173

This has the benefit of letting the element handle text reflow, with
the drawback being that the user need to manage min and max width of
the container element they render.

This approach seems much better imo, favouring composability over
configuration which is much more in line with the philosophy of d3.

### DIFF
--- a/packages/visx-annotation/src/components/HtmlLabel.tsx
+++ b/packages/visx-annotation/src/components/HtmlLabel.tsx
@@ -1,0 +1,87 @@
+import React, { useContext, useMemo } from 'react';
+import cx from 'classnames';
+import useMeasure from 'react-use-measure';
+import Group from '@visx/group/lib/Group';
+import AnnotationContext from '../context/AnnotationContext';
+import AnchorLine from './LabelAnchorLine';
+import { LabelProps } from './Label';
+
+const wrapperStyle = { display: 'inline-block' };
+
+export type HtmlLabelProps = Pick<
+  LabelProps,
+  | 'anchorLineStroke'
+  | 'className'
+  | 'horizontalAnchor'
+  | 'resizeObserverPolyfill'
+  | 'showAnchorLine'
+  | 'verticalAnchor'
+  | 'x'
+  | 'y'
+> & {
+  /** Pass in a custom element as the label to style as you like. Renders inside a <foreignObject>, be aware that most non-browser SVG renderers will not render HTML <foreignObject>s. See: https://github.com/airbnb/visx/issues/1173#issuecomment-1014380545.  */
+  children?: React.ReactNode;
+};
+export default function HtmlLabel({
+  anchorLineStroke = '#222',
+  children,
+  className,
+  horizontalAnchor: propsHorizontalAnchor,
+  resizeObserverPolyfill,
+  showAnchorLine = true,
+  verticalAnchor: propsVerticalAnchor,
+  x: propsX,
+  y: propsY,
+}: HtmlLabelProps) {
+  // we must measure the rendered title + subtitle to compute container height
+  const [labelRef, titleBounds] = useMeasure({
+    polyfill: resizeObserverPolyfill,
+  });
+  const { width, height } = titleBounds;
+
+  // if props are provided, they take precedence over context
+  const { x = 0, y = 0, dx = 0, dy = 0 } = useContext(AnnotationContext);
+
+  // offset container position based on horizontal + vertical anchor
+  const horizontalAnchor =
+    propsHorizontalAnchor || (Math.abs(dx) < Math.abs(dy) ? 'middle' : dx > 0 ? 'start' : 'end');
+  const verticalAnchor =
+    propsVerticalAnchor || (Math.abs(dx) > Math.abs(dy) ? 'middle' : dy > 0 ? 'start' : 'end');
+
+  const containerCoords = useMemo(() => {
+    let adjustedX: number = propsX == null ? x + dx : propsX;
+    let adjustedY: number = propsY == null ? y + dy : propsY;
+
+    if (horizontalAnchor === 'middle') adjustedX -= width / 2;
+    if (horizontalAnchor === 'end') adjustedX -= width;
+    if (verticalAnchor === 'middle') adjustedY -= height / 2;
+    if (verticalAnchor === 'end') adjustedY -= height;
+
+    return { x: adjustedX, y: adjustedY };
+  }, [propsX, x, dx, propsY, y, dy, horizontalAnchor, verticalAnchor, width, height]);
+
+  return (
+    <Group
+      top={containerCoords.y}
+      left={containerCoords.x}
+      pointerEvents="none"
+      className={cx('visx-annotationlabel', className)}
+    >
+      {showAnchorLine && (
+        <AnchorLine
+          anchorLineOrientation={Math.abs(dx) > Math.abs(dy) ? 'vertical' : 'horizontal'}
+          anchorLineStroke={anchorLineStroke}
+          verticalAnchor={verticalAnchor}
+          horizontalAnchor={horizontalAnchor}
+          width={width}
+          height={height}
+        />
+      )}
+      <foreignObject width={width} height={height} overflow="visible">
+        <div ref={labelRef} style={wrapperStyle}>
+          {children}
+        </div>
+      </foreignObject>
+    </Group>
+  );
+}
--- a/packages/visx-annotation/src/components/Label.tsx
+++ b/packages/visx-annotation/src/components/Label.tsx
@@ -132,7 +132,7 @@ function ForeignObjectLabel({
           height={height}
         />
       )}
-      <foreignObject overflow="visible">
+      <foreignObject width={width} height={height} overflow="visible">
         <div ref={labelRef} style={{ display: 'inline-block' }}>
           {children}
         </div>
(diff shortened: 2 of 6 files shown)


valtism:
@williaster Not sure how I can see the issues with Happo. I don't think there should be any visual regressions from this PR thought.

valtism:
Example of the annotation in action. Notice the `<div>` handles text reflow much better. The label is also much easier to style how you want.

https://user-images.githubusercontent.com/1286001/143527056-b6de800a-a564-4976-9f12-cf655568765c.mov

gazcn007 (approved):
LGTM

williaster:
hey @valtism , sorry I was away for all of Dec and am just getting back into the swing of things. will try to review/land this asap! (next day or two)

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -15,6 +15,8 @@ export type LabelProps = {
   backgroundPadding?: number | { top?: number; right?: number; bottom?: number; left?: number };
   /** Additional props to be passed to background SVGRectElement. */
   backgroundProps?: React.SVGProps<SVGRectElement>;
+  /** Pass in a custom element as the label to style as you like. Renders inside a <foreignObject> */
```
williaster, line 18: nit – period for docs.

```suggestion
  /** Pass in a custom element as the label to style as you like. Renders inside a <foreignObject>. */
```

I wonder also if it'd be worth noting from the [original issue](https://github.com/airbnb/visx/issues/1173#issuecomment-1014380545) something like 

> Note that copy/pasting content from a browser may not work if you use this option because not all non-browser SVG renderers render HTML `<foreignObject>`s

I'm actually not sure if this is true since the DOM tree in the implementation has a div `<foreignObject><div>...` so maybe don't have to worry about it for now. cc @richsilv

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -67,7 +69,78 @@ function getCompletePadding(padding: LabelProps['backgroundPadding']) {
   return { ...DEFAULT_PADDING, ...padding };
 }
 
-export default function Label({
+export default function Label({ ...props }: LabelProps) {
```
williaster, line 72: I think this has a unnecessary clone / spread

```suggestion
export default function Label(props: LabelProps) {
```

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -67,7 +69,78 @@ function getCompletePadding(padding: LabelProps['backgroundPadding']) {
...
+  if (props.children) {
+    return <ForeignObjectLabel {...props} />;
+  }
+  return <SvgLabel {...props} />;
+}
+
+function ForeignObjectLabel({
```
williaster, line 79: wondering if we could pull this out into a separate file since this one is getting pretty large.

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -67,7 +69,78 @@ function getCompletePadding(padding: LabelProps['backgroundPadding']) {
...
+          anchorLineOrientation={Math.abs(dx) > Math.abs(dy) ? 'vertical' : 'horizontal'}
+          anchorLineStroke={anchorLineStroke}
+          verticalAnchor={verticalAnchor}
+          horizontalAnchor={horizontalAnchor}
+          width={width}
+          height={height}
+        />
+      )}
+      <foreignObject width={width} height={height} overflow="visible">
+        <div ref={labelRef} style={{ display: 'inline-block' }}>
```
williaster, line 135: can we pull the `{ display: 'inline-block' }` style into a `const` outside the function since it doesn't change?

packages/visx-annotation/src/components/Label.tsx
williaster: Couple more thoughts:
- What do you think about calling it `HtmlLabel` instead? that might be more understandable by folks who don't know what `foreignObject` means. And `foreignObject` will still be documented via props.
- I think especially because this label type is so simplified / supports a subset of props, users might want to import and use it directly, so a separate file would be good. 
- For documentation purposes, it would probably also be ideal to explicitly define the subset of props it supports with `Pick<LabelProps, ...>` because otherwise users may wonder why some don't have an effect. 
- If separated into another file, for this component to show up in the docs we'll need to add it [here](https://github.com/airbnb/visx/blob/master/packages/visx-demo/src/pages/docs/annotation.tsx#L19) and export it from the package [`index.ts`](https://github.com/airbnb/visx/blob/master/packages/visx-annotation/src/index.ts#L2)

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -67,7 +69,78 @@ function getCompletePadding(padding: LabelProps['backgroundPadding']) {
...
+  children,
+  className,
+  horizontalAnchor: propsHorizontalAnchor,
+  resizeObserverPolyfill,
+  showAnchorLine = true,
+  verticalAnchor: propsVerticalAnchor,
+  x: propsX,
+  y: propsY,
+}: LabelProps) {
+  // we must measure the rendered title + subtitle to compute container height
```
williaster, line 90:
```suggestion
  // we must measure the rendered html content to compute container height
```

packages/visx-annotation/src/components/Label.tsx
```diff
@@ -255,3 +322,39 @@ export default function Label({
...
+
+interface AnchorLineProps {
+  anchorLineOrientation: 'horizontal' | 'vertical';
+  verticalAnchor: TextProps['verticalAnchor'];
+  horizontalAnchor: TextProps['textAnchor'];
+  anchorLineStroke: string;
+  width: number;
+  height: number;
+}
+function AnchorLine({
```
williaster, line 334: thanks for pulling this out 🙏  . if we pull `HtmlLabel` into a separate component, might have to do the same with this so it can be imported by `HtmlLabel` (don't think we need to add this to the docs/export from the package index, tho)

williaster (commented):
Woweee @valtism this one looks great 💯  thanks for adding this, it's a huge improvement for additional flexibility! 🔥 Can't wait to use it. 

I had a couple suggestions, mostly about moving out the html-based label into its own file to support easier import and documentation.

valtism:
@williaster thanks for the great feedback! I've incorporated everything you mentioned, but I'm not entirely sure how to test to see if the new `HtmlLabel` displays properly in the docs. Let me know if I missed anything.

packages/visx-annotation/src/components/HtmlLabel.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }, [propsX, x, dx, propsY, y, dy, horizontalAnchor, verticalAnchor, width, height]);
+
+  return (
+    <Group
+      top={containerCoords.y}
+      left={containerCoords.x}
+      pointerEvents="none"
+      className={cx('visx-annotationlabel', className)}
+    >
+      {showAnchorLine && (
```
williaster, line 70: one thing I noticed while playing with this is that the line renders _behind_ the HTML, do you think we should flip this and render the anchor line after `foreignObject`?

packages/visx-annotation/src/components/HtmlLabel.tsx
```diff
@@ -0,0 +1,87 @@
...
+          anchorLineOrientation={Math.abs(dx) > Math.abs(dy) ? 'vertical' : 'horizontal'}
+          anchorLineStroke={anchorLineStroke}
+          verticalAnchor={verticalAnchor}
+          horizontalAnchor={horizontalAnchor}
+          width={width}
+          height={height}
+        />
+      )}
+      <foreignObject width={width} height={height} overflow="visible">
+        <div ref={labelRef} style={wrapperStyle}>
```
williaster, line 81: when I was playing with this I found myself adding a bunch of styles to a `div` in `children`, I wonder if we should just expose this so users don't need an extra DOM element? something like `containerStyle?: React.CSSProperties`

packages/visx-annotation/test/Label.test.tsx
```diff
@@ -33,16 +35,18 @@ describe('<Label />', () => {
   });
   it('should render a background', () => {
     expect(
-      shallow(
-        <Label title="title test" showBackground resizeObserverPolyfill={ResizeObserver} />,
-      ).find('rect'),
+      shallow(<Label title="title test" showBackground resizeObserverPolyfill={ResizeObserver} />)
```
williaster, line 38: 🙏  thanks for fixing these

williaster (approved):
thanks for iterating 🙌  docs look good! 





fyi you can check by 
- rebuilding the package of interest `yarn build:workspaces --workspaces=@visx/drag --esm --watch` (the demo site pulls from the `esm/` build, this is documented in `CONTRIBUTING.md` but is subtle)
- start the demo site dev server `yarn dev:demo` (or from `packages/visx-demo` run `yarn dev`)
- visit http://localhost:3000/docs/annotation (or whatever)

I had two minor comments based on playing with it a bit. I might follow up with adding an `svg <> html` toggle in the demo for easier testing, so I'm happy to make those two tweaks in that PR if you'd rather merge as is.

williaster:
Hey @valtism I have some time today so am gonna land this and make those changes in another PR. Awesome job here! 🎉

valtism:
@williaster Perfect. I'm not precious about my code, so thanks for making those fixes. Look forward to using it 😊

PR #1397, opened 2021-12-23
Removed $ sign from Usage in Readme

DantrazTrev:
#### :memo: Documentation

Removed $ sign from Usage in Readme to make the copying getting started a bit easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Let's make a simple bar graph.
 
 First we'll install the relevant packages:
 
-```
+```bash
 npm install --save @visx/mock-data @visx/group @visx/shape @visx/scale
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,7 +70,7 @@ Let's make a simple bar graph.
 First we'll install the relevant packages:
 
 ```
```
williaster, line 72: maybe we can add a markdown formatter if we remove the `$` indicator?

```suggestion
```bash
```

williaster (commented):
lgtm @DantrazTrev , thanks for the improvement! I had one suggestion I'm happy to commit then we can merge.

PR #1423, opened 2022-01-21
fix(responsive): remove global ResizeObserver type declaration

williaster:
#### :bug: Bug Fix

This supersedes https://github.com/airbnb/visx/pull/1253 and fixes https://github.com/airbnb/visx/pull/1104. It removes the global declaration of `ResizeObserver` which breaks consumers ability to upgrade to `typescript@4` (which has its own `ResizeObserver` types).

Somewhat related to https://github.com/airbnb/visx/pull/1422

@gazcn007 @hshoff @kristw 
cc @MSiddharthReddy @HHK1

### DIFF
--- a/packages/visx-responsive/src/components/ParentSizeModern.tsx
+++ b/packages/visx-responsive/src/components/ParentSizeModern.tsx
@@ -2,11 +2,9 @@ import debounce from 'lodash/debounce';
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { ResizeObserver } from '../types';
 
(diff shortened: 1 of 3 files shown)


HHK1:
@williaster Thanks for the fix 🙏 How can I test this properly? I tried to use a git url in my package.json, but I realised after that it had zero chance to work because of the mono-repo structure 😅

hshoff (approved):

williaster:
@HHK1 we don't usually publish a version until after we merge, but this should be out shortly.

PR #1424, opened 2022-01-21
new(annotation): add HtmlLabel demo, add containterStyle prop

williaster:
#### :rocket: Enhancements

These were called out as optional in  #1383 where `HtmlLabel` was added to `@visx/annotation`:
- render `AnnotationLine` **on top of** the html content so its not clipped
- add an optional `containerStyle` prop so that consumers don't have to add an extra wrapper `div` to apply styles (since `HtmlLabel` already renders one).

#### :memo: Documentation

Adds an `Annotation content type` control to the `/annotation` demo to enable toggling between the `SVG <> HTML` label types.

#### :house: Internal

Adds a simple jest test for the new `HtmlLabel`

@valtism @hshoff @gazcn007 @kristw

### DIFF
--- a/packages/visx-annotation/test/HtmlLabel.test.tsx
+++ b/packages/visx-annotation/test/HtmlLabel.test.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
+import { render } from '@testing-library/react';
+
+import { HtmlLabel } from '../src';
+
+describe('<HtmlLabel />', () => {
+  it('should render HTML content', () => {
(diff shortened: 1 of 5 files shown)


packages/visx-annotation/test/HtmlLabel.test.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import ResizeObserver from 'resize-observer-polyfill';
```
williaster, line 2: will conflict with #1422

etr2460 (approved):
lgtm!

PR #1443, opened 2022-02-26
docs: Add https://augora.fr to "In the wild" section

goldenivan:
#### :memo: Documentation

Add https://augora.fr to "In the wild" section


williaster (approved):
Thanks for the addition @goldenivan !!! 

It looks great 🤩  Do I read correctly that you were able to use `@visx/annotation` for the pie labels? 🎉

PR #1460, opened 2022-03-31
fix(xychart): fix typo

tetsugi:
renamed `RenderTooltipGlypProps` to `RenderTooltipGlyphProps`


williaster (approved):
hey @tetsugi, sorry for the delayed review I was OOO for a bit. Thanks for the fix! 🙌

PR #1480, opened 2022-04-26
feat(xychart): add ssr to xychart

nagiek:
#### :rocket: Enhancements

- Allows data to be set upfront in <XYChart> as well as in in child Series components.
- Fixes a chicken-and-egg issue where we delay rendering until data is passed into the children Series components.

### DIFF
--- a/packages/visx-xychart/src/components/XYChart.tsx
+++ b/packages/visx-xychart/src/components/XYChart.tsx
@@ -1,11 +1,12 @@
 /* eslint jsx-a11y/mouse-events-have-key-events: 'off', @typescript-eslint/no-explicit-any: 'off' */
 import React, { useContext, useEffect } from 'react';
 import ParentSize from '@visx/responsive/lib/components/ParentSize';
-import { AxisScaleOutput } from '@visx/axis';
+import { AxisScale, AxisScaleOutput } from '@visx/axis';
 import { ScaleConfig } from '@visx/scale';
 
 import DataContext from '../context/DataContext';
 import { Margin, EventHandlerParams } from '../types';
+import { DataRegistryEntry } from '../types/data';
 import useEventEmitter from '../hooks/useEventEmitter';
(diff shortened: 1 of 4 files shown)


nagiek:
@williaster quick ping for workflow approval, thanks!

williaster:
Thanks for the PR @nagiek 🙏 . My main concern with this approach is that two ways to pass data (to `XYChart` or to `*Series`) may cause confusion.

I just tried implementing the `useComponentWillMount` approach but the `() => unregisterData(dataKeys)` we currently return in `useEffect` for unmounting components is causing problems/I can't quite get it to work right. I'll keep tinkering a bit and think about whether there are any alternatives to your suggested approach.

nagiek:
No problem, I understand. I kept working on my version but I agree it’s a different technique. 

What about splitting the existing `useEffect` call into two calls? One to `useMemo` containing `registerData`, and one to `useEffect` containing `unregisterData`?

nagiek:
Quick follow up, thank you!

ThnxFredrik:
Hello! I’m suffering the same issue over here, how is the progress looking, can I help out some how?

mscottford:
@nagiek Are you using this approach in production? I'd love to see SSR supported.

mscottford:
@williaster
> My main concern with this approach is that two ways to pass data (to `XYChart` or to `*Series`) may cause confusion.

Would a separate component make sense? Similar to how there are separate `Animated` and not-`Animated` components, there could be a `ServerXYChart` and a `ClientXYChart`, which could be aliased as just `XYChart` for backward compatibility. This way the different components could take different approaches to loading the data, and any duplicated functionality could be de-duplicated behind the scenes.

pedroapfilho:
Any updates on this? Doing what @mscottford said, creating a `ServerXYChart`, would fix the issues that you commented about @williaster? Are these still issues on the current react version?

pedroapfilho:
Hey @hshoff, any inputs on this?

Issue #1473, opened 2022-04-14
React 18 Support


AndrewJSchoen:
Would it be possible to check on having React 18 support? This seems like it might be trivial, and it seems like a number of projects are dependent on React 18 support #1470 (including my own). 

(New issue created because the original issue was created due to a `vx` vs `visx` difference and therefore closed).

Thanks so much!

bonesoul:
also looking for react 18 support.

Gaya:
Spent some time updating various dependencies and peer dependencies to support React 18. It looks like everything works fine and the tests are green.

There is however a little work left to update the documentation site (`visx-demo`) to use React 18 too as `render` from `react-dom` has changed.

I can help out if the maintainers would like that.

bonesoul:
any news on this?

Gaya:
> any news on this?

Haven't had the time to continue on this. Don't know if the maintainers are looking into this themselves, but I can work on it some more if I have some free time.

gladysbenitez:
Gaya, can you please explain what you did to get visx to use react 18?

bonesoul:
i'm also looking for a way to upgrade visx to react 18.

Gaya:
You can see the work I've done so far here: https://github.com/airbnb/visx/pull/1483/files
`visx-demo` needs some work before I can finish the PR. I'll try and do that when I have time.

williaster:
Thanks @Gaya ! We're definitely open to contributions to try and get this landed. Your PR looks like a great start, let me know if I can help with any `visx-demo` items. When you're happy with the PR I'll probably pull it locally to validate and we can go from there!

Gaya:
@williaster I’ll check tomorrow. It wasn’t too bad, had some local tooling not working correctly. Mostly need to change the demo pages, but the tests seem to be passing.

Gaya:
@williaster updated the PR and demo pages. The tests are green, everything builds just fine. Only thing I see is that next.js is complaining a bit about the React 18 stuff, but nothing breaking. Maybe updating Next.js would help, but I don't have any experience using it.
I think the rest of the PR is good to go.

orestis:
It looks like this is fixed, thanks everyone involved!

Gaya:
My pleasure

PR #1483, opened 2022-04-28
deps: Update to support React 18

Gaya:
#### :boom: Breaking Changes

- Allows React 18 and updates `visx-demo`'s render methods

#### :rocket: Enhancements

- Allows the use of React 18

See:  #1473


packages/visx-text/Readme.md
```diff
@@ -20,7 +20,7 @@ you'd expect.
 
 ```jsx
 import React from 'react';
-import { render } from 'react-dom';
+import { createRoot } from 'react-dom/client';
```
williaster, line 23: could just note this syntax is specific to v18

```suggestion
// note: react@18 syntax
import { createRoot } from 'react-dom/client';
```

williaster (commented):
Overall this looks great to me! Thanks so much for going through the updates. I'm just going to pull it locally and play with it before approving 👍

williaster:
Ah looks like CI is failing due to prettier formatting in all the sandbox updates. you should be able to auto-fix this by running `yarn format` (or `yarn run prettier --write`) in the root directory.

Gaya:
@williaster thanks for taking a look. Fixed the problems you addressed. Though the `README.md` in root also got updated when I ran `yarn format`, I didn't commit this file as I think it had nothing to do with the other updates I did. Don't know if the CI will fail because of that, but the demo code has been fixed.

Let me know if you need something else.

williaster (approved):
so great! thanks @Gaya ❤️ 

this looks good to me, I'm going to release this as a minor bump but keep it as an alpha to start. then if that looks good published we can do a non-alpha. (releases will be posted to this PR)

Gaya:
Awesome! Glad to help 🙏

PR #1485, opened 2022-04-28
fix(xychart, react-spring): fix duplicate key warning in Axis/Grid

williaster:
#### :bug: Bug Fix

Fixes https://github.com/airbnb/visx/issues/1435 and possibly related to https://github.com/airbnb/visx/issues/1038, https://github.com/airbnb/visx/issues/969 (both closed)

Was able to reproduce @carlsverre 's issue by using a time scale where values only varied on the order of MS – as suggested the value is coerced to a date string which omits the MS and is therefore equal across children

image

I updated the logic to include the index of the tick (and made the same fix for animated grid columns/rows) and the warning disappeared 😎 

@kristw 
cc @carlsverre @mithi

### DIFF
--- a/packages/visx-grid/src/types.ts
+++ b/packages/visx-grid/src/types.ts
@@ -10,7 +10,11 @@ export type GridScale<Output extends GridScaleOutput = GridScaleOutput> =
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   D3Scale<Output, any, any>;
 
-export type GridLines = { from: { x?: number; y?: number }; to: { x?: number; y?: number } }[];
+export type GridLines = {
+  from: { x?: number; y?: number };
+  to: { x?: number; y?: number };
+  index: number;
+}[];
 
 export type CommonGridProps = {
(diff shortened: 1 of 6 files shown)


packages/visx-grid/src/types.ts
```diff
@@ -10,7 +10,11 @@ export type GridScale<Output extends GridScaleOutput = GridScaleOutput> =
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   D3Scale<Output, any, any>;
 
-export type GridLines = { from: { x?: number; y?: number }; to: { x?: number; y?: number } }[];
+export type GridLines = {
+  from: { x?: number; y?: number };
+  to: { x?: number; y?: number };
+  index: number;
```
williaster, line 16: `AnimatedGridRows/Columns` in `@visx/react-spring` (where the bug occurs) extends `@visx/grid`, so we need to pass `index` through in this package.

kristw (approved):

PR #1489, opened 2022-05-01
feat(xychart): add colorAccessor to LineSeries

### LINKED ISSUE #1260: Add `colorAccessor` prop to XYChart LineSeries
A few of the XYChart series components like BarSeries support providing a `colorAccessor`. This is convenient when you want to specify a specific color for each series.

It would be nice if this support could be added to LineSeries as well. 

Currently you either need to modify the theme and track specific indexes to modify the correct colors, or specify a stroke, but the downside of the stroke is it doesn't update glyph you get when hovering.

Zenith00:
#### :rocket: Enhancements

- Add colorAccessor to XYChart LineSeries prop that accepts a function that takes the dataKey and returns a color, replacing the default theme color

resolves #1260

Based on #1005

### DIFF
--- a/packages/visx-xychart/test/components/LineSeries.test.tsx
+++ b/packages/visx-xychart/test/components/LineSeries.test.tsx
@@ -85,6 +85,21 @@ describe('<LineSeries />', () => {
       { showTooltip, hideTooltip },
     );
   });
+
+  it('should use colorAccessor if passed', () => {
+    const {container} = render(
+      <DataContext.Provider value={getDataContext(series)}>
+        <svg>
+          <LineSeries
+            dataKey={series.key}
+            {...series}
+            colorAccessor={(_) => 'banana'}
+          />
+        </svg>
+      </DataContext.Provider>,
+    );
+    expect(container.querySelector('path')).toHaveAttribute('stroke', 'banana')
+  });
 });
 
 describe('<AnimatedLineSeries />', () => {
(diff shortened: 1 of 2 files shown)


williaster (approved):
Thanks for the addition @Zenith00 ! 🙌  

This lgtm, I just merged a react@18 PR and want to test it as an alpha version. So once that's out we can publish this (should be later today if all goes well).

packages/visx-xychart/test/components/LineSeries.test.tsx
```diff
@@ -85,6 +85,21 @@ describe('<LineSeries />', () => {
       { showTooltip, hideTooltip },
     );
   });
+
+  it('should use colorAccessor if passed', () => {
+    const {container} = render(
+      <DataContext.Provider value={getDataContext(series)}>
+        <svg>
+          <LineSeries
+            dataKey={series.key}
```
williaster, line 94: looks like linting is complaining that this isn't a single line

```tsx
<LineSeries dataKey={series.key}·{...series}·colorAccessor={(_)·=>·'banana'} />
```

You could also run `yarn format` in the monorepo root, but sometimes that will update other files unintentionally (you can skip commit-ing them)

Zenith00: yarn format run. Looks like it picked up a `,`->`;` in BaseLineSeries as well

williaster:
Awesome, thanks for the fix and again for the contributions! Should be released shortly 🎉

alexandreabeh:
Excited to see this update - thank you @Zenith00 !

PR #1495, opened 2022-05-09
feat(brush): add renderbrushhandles

arnthor-agustsson:
Add renderBrushHandles to allow users to use render custom handles

#### :boom: Breaking Changes

- None

#### :rocket: Enhancements

- Add renderFunction to `<Brush />` so users can customize the handles

`<Brush
   {...props}
   renderBrushHandles={(brushHandleProps) => <MyCustomHandles {...brushHandleProps}/> }
/>`


arnthor-agustsson:
Bump @williaster, could you take a look at this please? Would appreciate some feedback and insights into this issue.

williaster:
thanks @arnthor-agustsson for the contribution 🙏  can you rebase and force push the branch to see if it triggers CI? I'm not sure why it's not running.

williaster:
yes! sorry I'll try to follow this today/tomorrow and land it if it passes CI.

packages/visx-brush/src/Brush.tsx
```diff
@@ -60,6 +61,8 @@ export type BrushProps = {
   innerRef?: React.MutableRefObject<BaseBrush | null>;
   /** Prevent drag end on mouse leaving from brush stage. */
   useWindowMoveEvents?: boolean;
+  /** Render function for custom brush handles */
```
williaster, line 64: Nit – just adding a period since docs are generated from these.

```suggestion
  /** Render function for custom brush handles. */
```

packages/visx-brush/src/BrushHandle.tsx
```diff
@@ -160,21 +181,38 @@ export default class BrushHandle extends React.Component<BrushHandleProps> {
...
+                onPointerDown={dragStart}
+                onPointerMove={dragMove}
+                onPointerUp={isControlled ? undefined : dragEnd}
+              >
+                {renderBrushHandles({
+                  ...this.props.handle,
+                  height: stageHeight,
+                  className: `visx-brush-handle-${type}`,
+
+                  isBrushActive: brush.extent.x0 !== -1 && brush.extent.x1 !== -1,
```
williaster, line 212: does it make sense to use `brush.isBrushing` here instead of using different logic?

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -43,6 +43,7 @@ export type BaseBrushProps = {
   disableDraggingSelection: boolean;
   resetOnEnd?: boolean;
   useWindowMoveEvents?: boolean;
+  renderBrushHandles?: (props: BrushHandleRenderProps) => JSX.Element | null;
```
williaster, line 46: Since this only renders one handle, I wonder if it makes sense to remove the `s` throughout? also should be able to use `ReactNode` for the return type which includes null.

```suggestion
  renderBrushHandle?: (props: BrushHandleRenderProps) => React.ReactNode;
```

packages/visx-demo/src/sandboxes/visx-brush/Example.tsx
```diff
@@ -194,6 +195,9 @@ function BrushChart({
             onClick={() => setFilteredStock(stock)}
             selectedBoxStyle={selectedBrushStyle}
             useWindowMoveEvents
+            renderBrushHandles={(props) =>
+              props.isBrushActive ? <BrushHandle {...props} /> : null
```
williaster, line 199: it might make sense to _always_ render the `BrushHandle` here, and then let it handle rendering `null` or not

williaster (commented):
@arnthor-agustsson thanks again for the addition! 🙌  had a couple minor suggestions, nothing to big tho.

packages/visx-brush/src/BrushHandle.tsx
williaster: ah I misinterpreted this. it's not whether brushing is happening, it's whether the handle is used as a trigger area?? is this necessary given that only `BrushHandles` for the specified `resizeTriggerAreas` are rendered?

https://github.com/airbnb/visx/blob/master/packages/visx-brush/src/BaseBrush.tsx#L635

arnthor-agustsson: Thanks for the read @williaster I really appreciate it. 
The `isBrushActive` function is just for cosmetics - hide the handles when there is no area selected, otherwise the handles are rendered at `x=0` position. 
Is there a better way to do this or should this be handled by the user?

williaster (approved):
awesome, thanks for iterating on those things! it looks great to me 😎 will merge it once CI completes.

williaster:
happo diff looks good 👍 
image

PR #1522, opened 2022-06-01
fix(brush): vertical brush with useWindowMoveEvents

kangaechigai:
#### :bug: Bug Fix

- Fix vertical resizing of a `Brush` with `useWindowMoveEvents` https://github.com/airbnb/visx/issues/1385

`handleWindowPointerMove` was checking `brushingType === 'bottom'` for both top and bottom, so dragging the top did nothing, and dragging the bottom moved the whole brush (since it was applying the offset to both `y0` and `y1`).


williaster (approved):
thanks for the fix @kangaechigai !

PR #1551, opened 2022-08-03
fix(text, shape): Hide generated text element from assistive technologies

thomashoggard:
Fixes: https://github.com/airbnb/visx/issues/1544

Currently the generated svg used to calculate text and shape size are being read by assitive technologies like screen readers.  This PR hides the generated element so it's now hidden for both screen reader and visual users. 

I tested using the demo app and verified the attribute was being added within Chrome dev tools. 
image


williaster (approved):
lgtm, thanks so much @thomashoggard for the bug report and fix! 🙏  

will get it merged/released once CI finishes.

PR #1553, opened 2022-08-08
Evolution of brush chart

Redskinsjo:
#### :rocket: Enhancements

- Evolution of brush chart. Took brush chart and added area chart (with tooltip) to it. Ends into a combined chart that was not available straight from the package.


williaster:
Thanks for the contribution @Redskinsjo 🙏 do you have an image of the resulting chart? it's a bit harder than this to add a new gallery example – we should make this more easy to do or create a discussion group to share custom examples. for now I will close this but the code will be there for reference, and again sharing this in discussions may be a good avenue.

PR #1563, opened 2022-09-03
Include visx/threshold in visx/visx

### LINKED ISSUE #1556: visx/threshold not included in visx/visx
"@visx/visx": "^2.12.2",

// had to include this separately
    "@visx/threshold": "^2.12.2",

DantrazTrev:
Fixes #1556
Added visx/threshold in visx/visx

### DIFF
--- a/packages/visx-visx/package.json
+++ b/packages/visx-visx/package.json
@@ -58,7 +58,7 @@
     "@visx/scale": "2.2.2",
     "@visx/shape": "2.12.2",
     "@visx/text": "2.12.2",
-    "@visx/threshold": "^2.12.2
+    "@visx/threshold": "2.12.2",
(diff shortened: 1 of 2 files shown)


williaster (commented):
Hey @DantrazTrev 👋  thanks for the fix/contribution!

The `package.json` looks good, but will also need to export the contents of the package from the [`index.ts` file](https://github.com/airbnb/visx/blob/master/packages/visx-visx/src/index.ts). after that we should be good to land 🚀

DantrazTrev:
Hei @williaster apologies didn't look that way . It's fixed thanks for pointing out

williaster (approved):
awesome, thanks for the quick change @DantrazTrev ! lgtm

PR #1573, opened 2022-09-24
fix(demo-geo-mercator): Example alert

vincecao:
#### :bug: Bug Fix

In [codesandbox demo example](https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-geo-mercator?file=/Example.tsx) for geo mercator, the alert in `onClick` is not working as expected. 

This small fix should resolve the issue.


vincecao:
Checked the code again, it looks like the missing on purpose. Because `events` was passed from props. Will close the PR.

PR #1574, opened 2022-09-24
feat(axis): add an ability to specify different paddings for the start and the end of the axis

andreykrupskii:
Hi, everyone! First of all, I want to say, that I'm impressed by the `visx` project, and I'm happy that I discovered it for myself! 

Currently, we are migrating our system from `recharts` to `visx` and found out that there is no way to specify different paddings for the start and the end of the axis. Such ability is present in `recharts` (https://recharts.org/en-US/api/XAxis#padding, https://recharts.org/en-US/api/YAxis#padding), and now we need to work around it using `hideAxisLine` option and our custom line. So it will be nice to have it also in `visx`.  

#### :rocket: Enhancements

- add an ability to specify different paddings for the start and the end of the axis

### DIFF
--- a/packages/visx-axis/src/axis/Axis.tsx
+++ b/packages/visx-axis/src/axis/Axis.tsx
@@ -8,6 +8,7 @@ import getTickPosition from '../utils/getTickPosition';
 import getTickFormatter from '../utils/getTickFormatter';
 import createPoint from '../utils/createPoint';
 import Orientation from '../constants/orientation';
+import getAxisRangePaddingConfig from '../utils/getAxisRangePaddingConfig';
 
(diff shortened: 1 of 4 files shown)


andreykrupskii:
@williaster, I will be happy for your review 😊

andreykrupskii:
Looks like there is no activity yet, I would be happy to get some feedback, even if this change is not going to be merged 😉

andreykrupskii:
@williaster, @hshoff, just a kind ping to maintainers

williaster (approved):
Hi @andreykrupskii 👋  thanks for the contribution and apologies for the delayed review – have been super busy recently so I appreciate the reminder ping 🙏 

This looks great to me! I just poked CI to run and can land it after that passes. Thanks for adding tests as well! 🔥

PR #1579, opened 2022-10-08
feat(shape): expose shape components prop types

andreykrupskii:
This pull request exposes shape components prop types, so it will be easier to wrap the original components into custom components. Currently, our team uses the next workaround to inherit the original component props: 

```typescript
import { BarRounded } from '@visx/shape';

type StyledBar = Omit<Parameters<typeof BarRounded>[0], 'ref'> & {
  // ...our custom props
}
```

While it works well for plain components, things get more complicated when it comes to generic components (like `LinePath`). So, it would be useful to expose the original types from the library itself. 

#### :rocket: Enhancements

- expose shape components prop types


andreykrupskii:
@williaster, I will be happy for your review 😊

andreykrupskii:
@williaster, can you please help me to understand why happo job failed? It reports that type was not exported, while it is: 
https://github.com/airbnb/visx/blob/master/packages/visx-shape/src/shapes/Arc.tsx#L7
image

andreykrupskii:
Unfortunately, it failed again... @williaster, can you please suggest how we can proceed with this PR?

williaster:
thanks for trying @andreykrupskii , looks like it's still having issues.

one question as a possible workaround: can you import types you need directly via deep imports? 
e.g., `import { ArcProps } from '@visx/shape/lib/Arc';`? this is always what we've done when we need access.

andreykrupskii:
Hey, @williaster, thanks for your support 😊 

For sure `import { ArcProps } from '@visx/shape/lib/Arc'` can be used as a workaround, while it looks to me like a violation of the library private interface. I think that better to find the reason of the pipeline failure (so prevent it in the future), rather than expose library's internal file structure to the customer code. 

BTW, feel free to close this pull request if you think that debugging this pipeline issue is not actual for now.

williaster:
I think there's something about explicitly exporting type names. while we export types elsewhere ([example](https://github.com/airbnb/visx/blob/master/packages/visx-scale/src/index.ts#L26)) it’s not a mix of component + types. my only thought at this point is that we generate both `esm` and `cjs` builds I think types (`.d.ts` files) are only exported for the `cjs` build, so if you export anything by name / that’s not `*` you might get a not defined error.

PR #1582, opened 2022-10-12
fix(mock-data): add missing bitcoinPrice to index.ts

filiprejmus:
#### :bug: Bug Fix
bitcoinPrice was missing in the mock-data index and could not be imported


williaster (approved):
ah thank you for noting and fixing this 🙏  lgtm

PR #1583, opened 2022-10-17
fix(tooltip, bounds): replace findDOMNode with refs for TooltipWithBounds

0xmax:
#### :bug: Bug Fix
This pull request extends the withBoundingRects to use optional refs instead of findDomNode. This should solve the deprecation warning in R18 with strictMode as discussed [here](https://github.com/airbnb/visx/issues/737)
To check the fix please enable `reactStrictMode: true` in `next.config.js` in the demo package. 

This _should_ not introduce any breaking changes. If the res is not passed, withBoundingRects falls back to findDOMNode. I would appreaciate pointers on how to correctly type the `nodeRef` in `TooltipWithBounds.tsx`

### DIFF
--- a/packages/visx-bounds/src/enhancers/withBoundingRects.tsx
+++ b/packages/visx-bounds/src/enhancers/withBoundingRects.tsx
@@ -24,6 +24,7 @@ export type WithBoundingRectsProps = {
   getRects?: () => { rect: rectShape; parentRect: rectShape };
   rect?: rectShape;
   parentRect?: rectShape;
+  nodeRef?: React.RefObject<HTMLElement>;
 };
 
 export default function withBoundingRects<Props extends object = {}>(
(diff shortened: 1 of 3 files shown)


0xmax:
Type / lint issues should be resolved.

williaster:
Hey @0xmax thanks so much for fixing this long-standing issue in a non-breaking way! 🙌  love it.

The code looks good to me, I'll pull the branch tomorrow and try adding `reactStrictMode: true` in the dev package 👍 👍 then hopefully we can get this in!

0xmax:
@williaster would be very happy if I can make an useful contribution to such an awsome project! Unfortunately there seems to be an issue with the initialization in withBoundingRect. I will investigate and come back to you shortly.

0xmax:
Ok, now it should work as intended.

williaster (approved):
awesome! locally looks good. let's land this and get it out! thank you again 🙏 ❤️ 

before 😢
image

after 🥲 
image

williaster:
even though this isn't breaking/could be a `patch` release, I'm going to do a `minor` release since it's a new prop on the `@visx/bounds` side.

PR #1599, opened 2022-11-15
fix(demo/radar):  uneven data elements renders spokes and web out of sync

studioVMdev:
#### :bug: Bug Fix

- Update example code

Before
image
After
image


studioVMdev:
@williaster

williaster (approved):
thanks for the fix @studioVMdev ! 🙌  lgtm

williaster:
noting that this doesn't actually affect the demo, but it would make it more generalizable to non-even number of input datum.

demo is unchanged:
image

Issue #1436, opened 2022-02-17
`Brush` does not resize when `width` and `height` props update


wilson-webdev:
This issue was mentioned [here](https://github.com/airbnb/visx/issues/1282#issuecomment-882815020), but I can't see an explicit issue, so here goes...

**Steps to replicate**:

1. Go to the [Brush example](https://codesandbox.io/s/heuristic-bardeen-luzfu4?file=/Example.tsx)
2. Update the Brush so that it is not the default size 
![Screenshot 2022-02-17 at 12 33 16](https://user-images.githubusercontent.com/24278706/154483725-8e131846-1c77-4612-a949-ebf5793932e5.png)
3. Resize chart and see the Brush has not resized ![Screenshot 2022-02-17 at 12 33 28](https://user-images.githubusercontent.com/24278706/154483777-c8896882-f616-4f32-84fc-a0fa8edc7628.png)

williaster:
thanks for reporting this @wilson-webdev 🙏  I don't have bandwidth to address it at the moment but happy to review a PR 👀

williaster:
fixed by https://github.com/airbnb/visx/pull/1536

Issue #1282, opened 2021-07-19
initialBrushPosition does not work on first render


JoeVanGundy:
I have an `AreaClosed` shape set up with a brush on top of it.

When the plots load, the brush does not show up.
However, if I use the code from the example to reset the plot, then the brush shows up.

Any ideas/how would I go about debugging this?
Since the values are hardcoded, what could prevent the brush from displaying?

I have hard coded my initialBrushPosition to a value I know is valid.
```  
const initialBrushPosition = useMemo(
    () => ({
      start: { x: 100 },
      end: { x: 300 },
    }),
    [xBrushScale]
);
```

My brush looks like:
```
 <Brush
    xScale={xBrushScale}
    yScale={yBrushScale}
    width={xBrushMax}
    height={yBrushMax}
    handleSize={8}
    margin={margin}
    innerRef={brushRef}
    brushDirection="horizontal"
    resizeTriggerAreas={['left', 'right']}
    initialBrushPosition={initialBrushPosition}
    onChange={onBrushChange}
    selectedBoxStyle={selectedBrushStyle}
    useWindowMoveEvents
 />
```

My x scale looks like:
```
const xBrushScale = useMemo(
   () =>
      scaleTime({
      range: [0, xBrushMax],
      round: true,
     domain: extent(sortedData, getDate),
   }),
   [xBrushMax, sortedData]
);

JoeVanGundy:
In the panic of trying every possibility to stop from embarrassing myself, I found the issue haha. 
On first render, the width of the plot is coming back as `0`.
So if the plot begins at `0`, then it seems that `initialBrushPosition` no longer works?
I see now there's a check in the demo to stop this from happening: `if (width < 10) return null;`.

However this does surface another issue that seems related.
If you make a brush selection, then resize the plot (via browser resize), the brush maintains the same position.
This limitation is in the demo as well.

Regular plots aren't affected by this, so is there any way around this, or is this just a limitation in the D3 library?

Thanks!
![image](https://user-images.githubusercontent.com/8807246/126218771-311d5a02-78c8-4535-8423-83563f673d25.png)

![image](https://user-images.githubusercontent.com/8807246/126218734-b4e47798-e935-4d30-ac2e-2a6068b00747.png)

williaster:
Hey @JoeVanGundy thanks for opening the issue and doing some digging. The responsive stuff can be tricky at times. 

[...]

PR #1606, opened 2022-12-08
fix(brush): update brush start/end on window resize

datananda:
#### :bug: Bug Fix

Fixes two issues with brush updating on chart resize (see #1282 and #1436):

**1. When moving a brush that has been resized due to chart resize, it regresses to some old sizing**

Fixed by updating brush `start` and `end` when the chart resizes. Previously, only `extent` was updated.

_Repro steps:_
- Use the [Brush example](https://airbnb.io/visx/brush) from the Gallery
- Update the brush so that it is not the default size
- Resize the window width
- Move the brush by dragging

_Current behavior:_
It regresses to some of its old positioning.

Screen Shot 2022-12-08 at 2 53 11 PM

_New behavior:_
It moves as expected in its new coordinate space.

Screen Shot 2022-12-08 at 2 53 50 PM

**2. Resizing when there is no brush selection, creates a phantom brush near 0,0**

Fixed by only updating brush `start`, `end`, and `extent` if there is a brush (i.e. `extent` coordinates are not all -1).

_Repro steps:_
- Use the [Brush example](https://airbnb.io/visx/brush) from the Gallery
- Clear the brush
- Resize the window width

_Current behavior:_
A phantom brush with handles is visible on the left side of the chart. The visx-brush-selection rect has non-zero height and is not positioned at -1,-1.

Screen Shot 2022-12-08 at 2 57 21 PM

_New behavior:_
No handles appear and the visx-brush-selection rect has attributes `x="-1" y="-1" width="0" height="0"` as expected

Screen Shot 2022-12-08 at 3 01 50 PM

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -117,22 +117,28 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     if (this.props.width !== prevProps.width || this.props.height !== prevProps.height) {
       // eslint-disable-next-line react/no-did-update-set-state
       this.setState((prevBrush: BaseBrushState) => {
-        const widthRatio = this.props.width / prevProps.width;
-        const heightRatio = this.props.height / prevProps.height;
+        let { start, end, extent } = prevBrush;
 
-        const start = {
-          x: widthRatio * prevBrush.extent.x0,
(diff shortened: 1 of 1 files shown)


williaster (approved):
Thanks for the fix and contribution @datananda ! 🙌 

This lgtm, will try to land after CI Passes

PR #1612, opened 2022-12-22
fix(demo/pages/docs): use relative imports to fix API docs

williaster:
#### :bug: Bug Fix

This fixes https://github.com/airbnb/visx/issues/1611 where our component types are no longer being shown in our documentation.

BUT this won't be deployable because we missed adding some next config for ESM modules in https://github.com/airbnb/visx/pull/1578

cc @N1gma


williaster:
gonna merge this so we can deploy a working docs page coupled with https://github.com/airbnb/visx/pull/1619

PR #1621, opened 2023-01-03
Revert "Revert breaking "deps(scale): bump `d3-interpolate` and `d3-scale`""

williaster:
### 💥 Breaking

Reverts airbnb/visx#1619 which effectively duplicates https://github.com/airbnb/visx/pull/1578 (which bumps these d3 packages to fix a security vulnerability) to release it as a breaking change since the new d3 packages are `ESM`-only.

(re) closes https://github.com/airbnb/visx/issues/1577

depends on
- [x] https://github.com/airbnb/visx/pull/1620
- [x] https://github.com/airbnb/visx/pull/1622
- [x] https://github.com/airbnb/visx/pull/1626
- [x] https://github.com/airbnb/visx/pull/1627

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "./packages/*"
   ],
   "resolutions": {
+    "@babel/parser": "^7.15.3",
     "handlebars": "4.5.3"
   }
 }
--- a/babel.config.js
+++ b/babel.config.js
@@ -20,7 +20,7 @@ const plugins = ['babel-plugin-typescript-to-proptypes'];
 
 const ignore = [
   'coverage/',
-  'node_modules/',
+  'node_modules/(?!(d3-(array|color|format|interpolate|scale|time|time-format)|internmap)/)',
   'public/',
   'esm/',
   'lib/',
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -114,6 +114,7 @@
     "./packages/*"
   ],
   "resolutions": {
+    "@babel/parser": "^7.15.3",
```
williaster, line 117: https://github.com/d3/d3-array/issues/147#issuecomment-890437860

williaster:
gonna release this tomorrow so I'm around if I break anything :)

williaster:
noting that the `3.0.0` was successful, but the `@visx/demo` publishing failed. I remember this maybe happening for major releases in the past with lerna. Trying to trigger another one now but wanted to make a note

https://github.com/airbnb/visx/actions/runs/3857560425/jobs/6575085943

Xiphe:
so v3.0.0 does indirectly run ESM only now, right?

williaster:
Correct, requiring `@visx` packages which import these `ESM`-only `d3` packages will need special handling which is why we reverted it and re-released as a major version so folks can opt into it.

PR #1626, opened 2023-01-04
internal(github/workflows): add tsconfig + demo site build checks

williaster:
#### :house: Internal

This adds a couple new checks to our `pull_request` github workflow:
- Ensure `tsconfig` references are up to date. Without nimbus, we don't get this for free anymore. I added a script to do this locally but we should run it in CI too.
- Ensure that the `nextjs` demo site builds without error. https://github.com/airbnb/visx/pull/1578 silently broke this so I'd like to avoid it in the future


williaster:
`tsconfig` check works 💯 

image

PR #1578, opened 2022-09-29
deps(scale): bump `d3-interpolate` and `d3-scale`

jreyes33:
Closes #1577

Bumping these packages allows for `d3-color` version 3.1.0 or higher to
be installed, fixing this vulnerability:
https://github.com/advisories/GHSA-36jr-mh4h-2g58

#### :bug: Bug Fix

- Fix [GHSA-36jr-mh4h-2g58](https://github.com/advisories/GHSA-36jr-mh4h-2g58) vulnerability

PR #1627, opened 2023-01-05
deps(demo): upgrade to next@11

williaster:
#### :house: Internal

Before it was reverted, https://github.com/airbnb/visx/pull/1578 broke the nextjs site due to the resulting `d3-*` ESM modules with the error

```bash
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/christopher-williams/dev/visx/packages/visx-demo/node_modules/@visx/scale/node_modules/d3-scale/src/index.js from /Users/christopher-williams/dev/visx/packages/visx-demo/node_modules/@visx/scale/lib/scales/band.js not supported.
```

This PR 
- updates the demo site from `next@9.5.4` to `next@^11` and turns on its experimental support for esm modules
  - no breaking changes for `9=>10`
  - no breaking changes for us for `10=>11` because we're on `react@18`, but some SSR mismatch errors surfaced and I had to refactor some of our next-config + happo config which references `next`'s webpack configuration.
- adds `next-transpile-modules` with config for the problematic `d3-*` packages
  - when trying different combinations I found this was still necessary
- fixed a `@visx/visx` version in `visx-demo` 
- updated the `tsconfig` for `visx-demo` which was trying to extend a base root config which wasn't working great

verified that the site builds locally when rebased on #1578


packages/visx-demo/package.json
```diff
@@ -66,7 +66,7 @@
     "@visx/text": "2.17.0",
     "@visx/threshold": "2.18.0",
     "@visx/tooltip": "2.17.0",
-    "@visx/visx": "2.16.0",
+    "@visx/visx": "2.18.0",
```
williaster, line 69: this was not the latest, which meant a copy of all `@visx/*` packages were in a local `node_modules` and hid some errors.

packages/visx-demo/package.json
```diff
@@ -98,7 +99,8 @@
     "react-markdown": "^4.3.1",
     "react-tilt": "^0.1.4",
     "recompose": "^0.26.0",
-    "topojson-client": "^3.0.0"
+    "topojson-client": "^3.0.0",
+    "webpack": "^5.75.0"
```
williaster, line 103: when upgrading `next` I think we lost a transient `webpack` dependency that `happo` was accidentally depending on.

williaster: not true, we actually configure happo to reference next's bundled version of webpack.

Issue #1638, opened 2023-01-25
Zoom pinch spasming 


kiliancs:
The zoom example in https://airbnb.io/visx/zoom-i demonstrates the issue.

https://user-images.githubusercontent.com/98786/214480791-32a4b3bf-d9bb-4776-82e8-9e13fde43cc5.mov

I'm seeing this in Firefox and Chrome, but not Safari, while using the pinch gesture using the touchpad.

atfera:
I don't know if it's related, but I have the same issue when using wheel + holding ctrl key, in Firefox and Chrome but not Safari.

kiliancs:
I have noticed that the wheel handler (`handleWheel`) is fired when pinching, as is `handlePinch`. Occasionally, the two events scale in the opposite direction, causing the issue.

<img width="486" alt="Screen Shot 2023-01-25 at 11 30 37 AM" src="https://user-images.githubusercontent.com/98786/214620430-1a59c6b2-f27c-4be7-a154-a67de9aee614.png">

kiliancs:
One thing to try is to ignore wheel events while pinching, similar to how it's done in `onDrag`:

```ts
      onDrag: ({ event, pinching, cancel }) => {
        if (pinching) { // <----
          cancel();
          dragEnd();
        } else if (!(event instanceof KeyboardEvent)) {
          dragMove(event);
        }
      },
```

kiliancs:
My attempt at a fix #1639.

PR #1639, opened 2023-01-25
fix(zoom): prevent wheel interfering with pinch

kiliancs:
#### :bug: Bug Fix

- Prevent wheel interfering with pinch 
  
  Using a macbook, a touchpad pinch gesture will trigger `useGesture`'s pinch event, as desired, but it will also trigger the wheel event. Except in Safari.

  With this change, the wheel event handler bails if a pinch gesture is ongoing.
  
  Tested in Firefox, Chrome and Safari on a macbook.

  Fixes: #1638

### DIFF
--- a/packages/visx-zoom/src/Zoom.tsx
+++ b/packages/visx-zoom/src/Zoom.tsx
@@ -329,7 +329,7 @@ function Zoom<ElementType extends Element>({
       onPinch: handlePinch,
       onWheel: ({ event, active, pinching }) => {
         if (
-          // Outside of Safari, the wheel event is fired together with the pinch event.
+          // Outside of Safari, the wheel event is fired together with the pinch event
           pinching ||
           // currently onWheelEnd emits one final wheel event which causes 2x scale
           // updates for the last tick. ensuring that the gesture is active avoids this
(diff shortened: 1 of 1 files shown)


williaster (approved):
Hey @kiliancs thanks so much for the bug report and digging in to fix this nuance 🙌  apologies for the review delay, have been absolutely slammed with work.

This lgtm but I think CI somehow didn't get triggered. Would it be possible for you to push another commit or rebase to trigger? if not I'm happy to do it since you already did the hard work 😅  thanks again

packages/visx-zoom/src/Zoom.tsx
```diff
@@ -327,10 +327,16 @@ function Zoom<ElementType extends Element>({
       },
       onDragEnd: dragEnd,
       onPinch: handlePinch,
-      onWheel: ({ event, active }) => {
-        // currently onWheelEnd emits one final wheel event which causes 2x scale
-        // updates for the last tick. ensuring that the gesture is active avoids this
-        if (!active) return;
+      onWheel: ({ event, active, pinching }) => {
+        if (
+          // Outside of Safari, the wheel event is fired together with the pinch event.
```
williaster, line 332:
```suggestion
          // Outside of Safari, the wheel event is fired together with the pinch event
```

williaster (commented):
looks like I can't push to your remote easily, so just going to make a UI suggestion and commit to nudge CI

Issue #1659, opened 2023-02-25
Cannot find module '@visx/.../src/...' Typescript error


SheaJanke:
### Problem
Imports of the form `import ... from '@visx/.../src/...';` are causing the following typescript error:
`Cannot find module '@visx/.../src/...' or its corresponding type declarations.`

This can be seen in the codesandbox demos from the visx documentation:

- https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-wordcloud?file=/Example.tsx:116-170

- https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-xychart?file=/ExampleControls.tsx:632-711

### Cause of the Issue
When the visx packages are published to npm, the `src` directories seem to get converted to `lib` directories. As a result, any references to the `src` directory are no longer valid.

### Suggested Fix
Convert all imports of the form `import ... from '@visx/.../src/...';` to `import ... from '@visx/.../lib/...';`. After a quick regex search through the codebase, I found 6 import statements that need to be modified. 

I'm willing to create a pull request for this change.

PR #1660, opened 2023-02-25
fix(demo/sandboxes, xychart): Convert '@visx/.../src/...' to '@visx/.../lib/...'

SheaJanke:
#### :bug: Bug Fix
- Replace all instances of `import ... from '@visx/.../src/...';` with `import ... from '@visx/.../lib/...';` 
- Closes: #1659


williaster (approved):
Hey @SheaJanke thanks for the bug report and the fix here! I think there were some paths that were incorrectly updated to`src/` during a major dev refactor at the end of 2022 that caused this.

lgtm!

SheaJanke:
Thanks for the review @williaster! Just a heads up that this also affects @visx/xychart, so maybe it should be added as a label as well.

williaster:
oh yikes, thanks for the callout as that deserves a proper release. I am surprised that passed CI previously 😱

PR #1662, opened 2023-02-28
feat(axis): support object type tickLabelProp

### LINKED ISSUE #1657: Specifying tick label props should not wipe all defaults
Now, when setting something like

```ts
tickLabelProps={() => ({
   stroke: "red"
})};
```

will remove all tick label props that had been set by default, which is probably not what most users want.
To make matters worse, the default props are not exported from `@visx/scale` so users end up having to look through the source code and copy-paste the default props in their own projects.

Lastly, it would help if you'd be allowed to specify `tickLabelProps` as a plain object or a function, instead of just as a function.

Happy to make a PR for this if this is something people can agree on

Robin-Hoodie:
#### :rocket: Enhancements

- Allow specifying `tickLabelProps` as an object
- Keep non-overridden default props when specifying `tickLabelProps` as an object

Resolves #1657


packages/visx-axis/src/utils/getTickLabelProps.ts
```diff
@@ -0,0 +1,15 @@
+import { ScaleInput } from '@visx/scale';
+import { TextProps } from '@visx/text';
+import { TickLabelProps, AxisScale } from '../types';
+
+export function getTickLabelProps<Scale extends AxisScale>(
```
williaster, line 5: generally I like the idea of utils to keep things DRY, but in this case I think it makes things a bit harder to read and the types here become complex (e.g., passing in the generics for `Scale` even if it's just an object, or when using this you still have to coerce to `as Partial<TextProps>`) so I wonder if you should just repeat the following logic in the 4x `Axis*` components. wdyt?

```tsx
typeof tickLabelProps === 'function'
  ? tickLabelProps
  : {
    ...defaultTickLabelProps,
    ...tickLabelProps,
  }
```

williaster (commented):
@Robin-Hoodie thanks for the awesome addition! this will definitely clean up this API / give a better dev experience. 🙌 

Overall looks good to me but I had one suggestion that might make it a bit more readable/reduce complexity.

packages/visx-axis/src/axis/AxisBottom.tsx
```diff
@@ -3,29 +3,29 @@ import cx from 'classnames';
 import Axis from './Axis';
 import Orientation from '../constants/orientation';
 import { SharedAxisProps, AxisScale } from '../types';
+import { getTickLabelProps } from '../utils/getTickLabelProps';
 
-export const bottomTickLabelProps = (/** tickValue, index */) =>
```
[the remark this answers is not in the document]
Robin-Hoodie: Agreed. Consumers should not rely on the internal API

packages/visx-axis/src/utils/getTickLabelProps.ts
Robin-Hoodie: I did think it was quite verbose and hard to read, especially with all the type verbosity. 
I've gone ahead and applied your suggestion in a501cb5d13c18cdfaa1f087d6a170b9345a4e6ee

packages/visx-axis/src/axis/AxisLeft.tsx
```diff
@@ -4,29 +4,35 @@ import Axis from './Axis';
...
 
 export default function AxisLeft<Scale extends AxisScale>({
   axisClassName,
   labelOffset = 36,
-  tickLabelProps = leftTickLabelProps,
   tickLength = 8,
+  tickLabelProps,
   ...restProps
 }: SharedAxisProps<Scale>) {
+  const tickLabelPropsFinal =
```
williaster, line 23: thanks for updating this 🙏

williaster (approved):
lgtm! thanks again @Robin-Hoodie 💯

packages/visx-axis/src/axis/AxisLeft.tsx
Robin-Hoodie: Happy to help!

PR #1686, opened 2023-04-12
fix(xychart): extra space in version definition, that prevents 18.2.0 version t…

codejunkienick:
#### :bug: Bug Fix

Fixes extra space in package.json of @visx/xychart


williaster (approved):
wow thanks for the fix here 🙏

PR #1717, opened 2023-06-21
vendor: detect ESM-only packages

williaster:
#### :house: Internal

After doing some manual work to parse through our `d3-` dependencies and flag those with versions that are ESM-only, or have transitive dependencies which are ESM-only, I decided it's not a great idea to do this manually. 

🫠
image

We actually don't have a huge number of ESM-only dependencies at the moment, **but** if we were to version bump any number of `d3-` dependencies, we could pick up more ESM-only versions and we're back at having to figure out which are ESM-only. Ideally this detection can be automated.

This PR adds a script which we can use in CI to detect and throw errors if any ESM-only dependencies are detected. If any are detected, they can be added to our new/upcoming `@visx/vendor` package.

- [x] write script
- [x] validate
- [x] hook up to CI

Example output
image

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,6 +30,9 @@ jobs:
       - name: 🤖 Ensure TS references are up to date
         run: yarn type:update-refs && git diff --exit-code
 
+      - name: 👀 Ensure we have no ESM-only dependencies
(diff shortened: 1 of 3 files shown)


williaster:
validation in CI

image

PR #1718, opened 2023-06-22
add @visx/vendor

williaster:
#### :house: Internal

This is part of https://github.com/airbnb/visx/pull/1716 and adds the logic for vendoring `esm` and `cjs` versions of ESM-only libs. Again this is heavily based on https://github.com/FormidableLabs/victory/tree/main/packages/victory-vendor

Some key notes for what this package does (some of this is also included in the readme):
- the `package.json` `dependencies` defines which packages are vendored. as noted in https://github.com/airbnb/visx/pull/1717, this is assumed to include **all transitive ESM-only dependencies** (e.g,. ESM-only `d3-scale` depends on another ESM-only `d3-array`) so that we can vendor all of them

- ~dependencies use [`npm` aliases](https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias) of the form `d3-array` => `vendor-d3-array@npm:d3-array`~
  - ~why? one challenge with this repo is that different `visx` packages may use different versions of the same `d3-` package meaning that some dependencies might live in the root `node_modules` or the `node_modules` of a given package. adding a `vendor-` alias is a simple way to **guarantee** that we are vendoring / sourcing the right version of the right package from the root `node_modules`~
  - **EDIT** this added a great deal of complexity (and ultimately would require us to fully-transpile ESM versions of ESM-only libraries to have internally-consistent references), so the approach switched to using `yarn` [`nohoist`](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/)

- for every vendored package `<pkg>` we export the following (note: none of these are ever checked in, they are git-ignored):
  - an ESM version of the package in `esm/<pkg>.js` (this exports the original ESM-only `<pkg>` code)
  - a CJS version of the package in `lib/<pkg>.js`
    - this points to the transpiled version of the **entire** `<pkg>` package in `vendor-cjs/<pgk>/src/**/*.js`
    - `vendor-cjs/<pkg>/LICENSE` contains the upstream license of the vendored package
    - within `vendor-cjs/<pgk>/`, all references to **other** ESM-only packages are replaced by references to the appropriate CJS vendored packages.
      - e.g., `vendor-cjs/d3-scale/src/band.js` references `d3-array`, which is replaced by `"../../d3-array/src/index.js"`
  - TypeScript types from `@types/<pkg>` as root `<pkg>.d.ts` files (when available as specified in the `package.json` `dependencies`)
  - a root `<pkg>.js` file (pointing to the CJS version of the lib) for tooling that doesn't yet
  support `package.json:exports` ([conditional exports](https://nodejs.org/api/packages.html#conditional-exports))

TODO 
- [x] add tests for all vendored packages
- [x] PR to update all references to ESM-only packages to `@visx/vendor` (`@visx/scale` is updated in this PR as a proof of concept) https://github.com/airbnb/visx/pull/1719

<details>
  <summary>Here is a complete set of files generated for `d3-array`</summary>

`packages/visx-vendor/esm/d3-array.js`
```ts
/**
 * `@visx/vendor/d3-array` (ESM)
 * See upstream license: https://github.com/d3/d3-array/blob/main/LICENSE
 *
 * This ESM package re-exports the underlying installed dependencies of 
 * `node_modules/d3-array` (aliased as `d3-array`)
 */
export * from 'd3-array';
```

`packages/visx-vendor/lib/d3-array.js`
```ts
/**
 * `@visx/vendor/d3-array` (CommonJS)
 * See upstream license: https://github.com/d3/d3-array/blob/main/LICENSE
 *
 * This CJS package exports transpiled vendor files in `vendor-cjs/d3-array`
 */
module.exports = require('../vendor-cjs/d3-array/src/index.js');
```

`packages/visx-vendor/vendor-cjs/d3-array/`
(not shown because it's the transpiled lib, includes `src/**/.js` and `LICENSE`)

`packages/visx-vendor/d3-array.d.ts`
```ts
/** `@visx/vendor/d3-array` (TypeScript) 
 *
 * Re-exports the types from `@types/d3-array`
 */
export * from 'd3-array';
```

`packages/visx-vendor/d3-array.js`
```ts
/**
 * `@visx/vendor/d3-array` (CommonJS)
 * See upstream license: https://github.com/d3/d3-array/blob/main/LICENSE
 *
 * This file only exists for tooling that doesn't work yet with package.json:exports
 * by proxying through the CommonJS version.
 */
module.exports = require('./vendor-cjs/d3-array/src/index.js');
```

</details>

### DIFF
--- a/packages/visx-vendor/Readme.md
+++ b/packages/visx-vendor/Readme.md
@@ -7,3 +7,51 @@ some transitive dependencies, which are ESM-only.
 This package is heavily based off of
 [`victory-vendor`](https://github.com/FormidableLabs/victory/tree/main/packages/victory-vendor)
 which aims to solve the same problem.
+
+## Vendored packages
+
+All vendored packages are listed as `dependencies` in the `package.json` of this package.
+
+## How it works
+
+We provide two alternate paths and behaviors -- for ESM and CommonJS
+
+### ESM
+
+When you use a module `import` syntax like the following, it will resolve to a re-exported version
+of `node_modules/d3-interpolate`, the unmodified ESM library from D3.
+
+```ts
+import { interpolate } from '@visx/vendor/d3-interpolate';
+```
+
+### CommonJS
+
+If you use a CJS `require` syntax like the following, it will resolve to an alternate path that
+contains the **transpiled** version of the underlying `d3-*` (or other) library to be found at
+`@visx/vendor/lib-vendor/d3-interpolate/**/*.js`.
+
+```ts
+const { interpolate } = require('victory-vendor/d3-interpolate');
+```
+
+Such transpiled versions have _internally consistent_ import references to other other
+`@visx/vendor/lib-vendor/<pkg-name>` paths that need to be transpiled.
+
+### TODO
+
+- [x] collect d3 requirements (manual) to inform design
+- [ ] automate dependency requirements
+  - yarn install packages
+  - script
+    - crawls packages from the root, finding those in the map
+      - yarn why <pgk> => extract <pgk>@XXX => compare XXX to map => if in map => add to dep
+      -
+  - bump d3-dep anywhere to version that needs vendor => CI should assert that the package.json diff
+    is clean
+  -
+- [ ] create @visx/vendor
+  - [ ]
+- [ ] write transpilation
+- [ ] update `d3-*` imports to reference `@visx/vendor/d3-*`
+  - [ ]
--- a/packages/visx-vendor/Readme.md
+++ b/packages/visx-vendor/Readme.md
@@ -17,9 +17,12 @@ versions of `d3` packages). For each (non-types) package `<pkg>`, we generate th
 - an ESM version of the package in `esm/<pkg>.js`
 - a CJS version of the package in `lib/<pkg>.js`
   - this points to the fully-transpiled version of the package in
-    `vendor-cjs/vendor-<pgk>/src/index.js`
-  - `vendor-cjs/vendor-<pgk>/LICENSE` contains the upstream license of the vendored package
-- TypeScript types from `@types/<pgk>` as root `<pkg>.d.ts` files
+    `vendor-cjs/vendor-<pkg>/src/index.js`
+  - `vendor-cjs/vendor-<pkg>/LICENSE` contains the upstream license of the vendored package
+  - other ESM-only packages (e.g., `<pkg2>`) that are referenced by `<pkg>` are updated to point to
+    `vendor-cjs/vendor-<pkg2>/src/index.js`
+- TypeScript types from `@types/<pkg>` as root `<pkg>.d.ts` files (when available as specified in
+  the `package.json` `dependencies`)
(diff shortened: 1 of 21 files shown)


packages/visx-vendor/scripts/buildVendor/babel.config.js
```diff
@@ -0,0 +1,64 @@
...
+      },
+    ],
+    [
+      'module-resolver',
+      {
+        // Convert all imports for _other_ ESM-only dependencies to the relative
+        // path in our vendor package.
+        resolvePath(sourcePath, currentFile) {
+          // extract the pkg name and detect if there is a deep import path
+          const packagePattern = /^(?<pkg>([^/]))(?<path>.*)/;
```
williaster, line 23: this is still not working correctly

packages/visx-vendor/Readme.md
```diff
@@ -7,3 +7,52 @@ some transitive dependencies, which are ESM-only.
...
+```ts
+const { interpolate } = require('@visx/vendor/d3-interpolate');
+```
+
+Such transpiled versions have _internally consistent_ import references to other other
+`@visx/vendor/vendor-cjs/<pkg-name>` paths that need to be transpiled.
+
+### Root index files & types
+
+In addition to supporting
```
williaster, line 58: finish this

packages/visx-vendor/package.json
```diff
@@ -11,11 +11,54 @@
       "charts"
     ],
     "author": "@williaster",
-    "license": "MIT",
-    "dependencies": {},
-    "devDependencies": {},
+    "license": "MIT and ISC",
+    "dependencies": {
+      "vendor-d3-array@npm:d3-array": "^3.2.1",
```
williaster, line 16: see the PR description for why these are aliased this way

packages/visx-vendor/package.json
```diff
@@ -11,11 +11,54 @@
...
+      "package.json": "./package.json",
+      "d3-*": {
+        "import": "./esm/d3-*.js",
+        "require": "./lib/d3-*.js",
+        "types": "./d3-*.d.ts"
+      },
+      "internmap": {
+        "import": "./esm/internmap.js",
+        "require": "./lib/internmap.js",
+        "types": "./types/internmap.d.ts"
```
williaster, line 61: remove this/doesn't exist

packages/visx-vendor/scripts/buildVendor/utils.ts
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * Re-exports the types from \`${pkg.packageName}\` 
+ */
+export * from '${pkg.npmAlias}';`;
+}
+
+// note: this is how we pass these dynamic variables into the
+// babel config. it's not great but babel config files can't easily
+// import from TS module files like this
+process.env.VENDOR_CJS_PATH = VENDOR_CJS_PATH;
```
williaster, line 160: noting I spent a TON of time trying to figure out how to leverage these inside the `babel.config.js` file ... you can't import the values from an ESM module like this, this was the cleanest way I came up with

PR #1721, opened 2023-06-27
visx*: adopt conditional exports

williaster:
#### :bug: Bug Fix

WIP – unclear what precisely is required here (Relevant discussion https://github.com/airbnb/visx/issues/1637#issuecomment-1608391103)

This is part of the broader https://github.com/airbnb/visx/pull/1716, and updates all `package.json` legacy `main/module/types` fields with the [conditional `exports`](https://nodejs.org/api/packages.html#conditional-exports) entry.


williaster:
abandoning for now

Issue #1726, opened 2023-06-30
XYChart TooltipProvider: "Can't perform a React state update on an unmounted component."


SheaJanke:
## Problem
The XYChart TooltipProvider is throwing  a "Can't perform a React state update on an unmounted component." warning if the chart is unmounted from the DOM before the tooltip is hidden. 

## Reproduction
This issue can easily be reproduced by wrapping the `XYChart` in a `TooltipProvider` with a large `hideTooltipDebounceMs`, and removing the XYChart from the DOM before the timer expires. Here is a [CodeSandbox link](https://codesandbox.io/s/objective-hooks-vcjk4w?file=/src/App.tsx) that can be used to reproduce the issue.

**Note:** This warning only appears in React  <= 17.x

## Proposed Solution
Add an `isMounted` check in the debounced hide tooltip function.

SheaJanke:
After some further thought, I think it makes more sense to cancel the `debouncedHideTooltip` function when the component unmounts instead of checking if the component is mounted. I created a small pull request with this proposed fix.

PR #1728, opened 2023-07-05
fix(xychart): Cleanup async function in TooltipProvider

SheaJanke:
#### :bug: Bug Fix

-  Cleanup the `debouncedHideTooltip` function in the `TooltipProvider` component when it unmounts.

closes: #1726


williaster (approved):
this lgtm, thanks for the fix @SheaJanke! 🙏

Issue #1715, opened 2023-06-21
XYChart GlyphSeries glyphs appear at zero coordinate


sshsphere:
When using GlyphSeries the glyphs don't follow the data, instead all glyphs appear at the (0, 0) coordinate. AnimatedGlyphSeries works correctly.

I encountered this bug in my own project, but you can reproduce it in the XYChart example:
1. go to <https://airbnb.io/visx/xychart>
2. check "render glyphs"
3. uncheck "use animated components"
4. all glyphs appear at top left corner

I tested with the latest versions of Chrome and Firefox

SheaJanke:
It seems like the issue here is with the demo implementation, not the `@visx/xychart` package. The demo is not using passing the `top` and `left` props to the glyph components, causing all the glyphs to be placed at (0,0). I created a potential fix for this above :arrow_up:.

PR #1730, opened 2023-07-07
Fix glyph positioning in the xychart demo. 

SheaJanke:
#### :bug: Bug Fix

- Fix the positioning of the non-animated glyph series in the xychart demo.

closes: #1715


williaster (approved):
thanks for digging in and fixing this @SheaJanke 🙌  lgtm 🚀

Issue #979, opened 2020-12-23
Is there support for multiple Y points on a single X point?


ShaneHudson:
I'm using the XYChart, which seems really nice except for a slight edge case I have.

Using a time scale for the X axis, if a single date has multiple entries they show on the chart but both the tooltips and curve break. You can see the issue with the curve here:

![image](https://user-images.githubusercontent.com/487005/103008350-e8b53780-452c-11eb-90f2-2a658827b14a.png)

Is it just simply not supported? My assumption is separate times may fix it so going to try that as a workaround.

williaster:
Hey @ShaneHudson thanks for checking out `xychart`. I think the issue is the `curve` type as it looks like you figured out. This determines how lines should be drawn / extrapolated between points. In case others come across this I created a [sandbox here](https://codesandbox.io/s/multiple-y-values-per-x-value-c4swu?file=/Example.tsx). This works okay for 2 points, but beyond that (as in your images) it seems like there's no guarantee that the line between two points wouldn't overlap the other lines with the same x-value unless you sorted points with shared x-values by their y-value.

<img src="https://user-images.githubusercontent.com/4496521/103039927-e819a280-4526-11eb-9637-097fbed94f5e.png" width="600" />

Let's leave this issue open as looks like the tooltip logic doesn't quite handle multiple points correctly, as you can see in the demo (left of the point picks the top point, right of the point picks the bottom point). For `LineSeries`, Tooltip logic currently is based on the nearest x- value, without considering the nearest y-value which should be the tie breaker if multiple points have equal x-distance. This should be pretty easy to fix.

ShaneHudson:
I'm unsure if it is the same issue or separate... but I've noticed tooltips also break (shows for max and min but none in the middle) if you have the data sorted backwards.

adamstambouli:
Would it be possible to leverage [`findNearestDatumXY`](https://github.com/airbnb/visx/blob/master/packages/visx-xychart/src/utils/findNearestDatumXY.ts#L6C25-L6C43) to address this? As far as I can understand, using voronoi-based indexing would avoid the issue above but won't be able to snap tooltip to x and/or y datums if the point is too far away?

I'm trying to take a look at how we can update [`findNearestDatumSingleDimension`](https://github.com/airbnb/visx/blob/master/packages/visx-xychart/src/utils/findNearestDatumSingleDimension.ts#L7:25) to fix but I'm not sure where to start. Any ideas?

SagnikPradhan:
[...]

PR #1734, opened 2023-07-25
fix(tooltip): handle multiple points (#979)

jernestmyers:
#### :bug: Bug Fix

Relates to #979 

Adds an optional prop (`findNearestDatumOverride`) to the `XYChart` and `GlyphSeries` components that get passed into the `useEventHandlers` hook for improved tooltip behavior. This felt like a viable path given that [the `useEventHandlers` hook](https://github.com/airbnb/visx/blob/master/packages/visx-xychart/src/hooks/useEventHandlers.ts) already includes the following prop definition:
```js
  /** Optionally override the findNearestDatum logic. */
  findNearestDatum?: (
    params: NearestDatumArgs<XScale, YScale, Datum>,
  ) => NearestDatumReturnType<Datum>;
```

[I tested this approach in a new viz component](https://github.com/VEuPathDB/web-monorepo/pull/381) we are building and the tooltip behaves as desired/expected by using [`findNearestDatumXY`](https://github.com/airbnb/visx/blob/master/packages/visx-xychart/src/utils/findNearestDatumXY.ts) as the override function when consuming `XYChart` and `GlyphSeries`.

Happy to improve upon this implementation or open to alternatives.

### DIFF
--- a/packages/visx-xychart/src/components/XYChart.tsx
+++ b/packages/visx-xychart/src/components/XYChart.tsx
@@ -2,11 +2,11 @@
 import React, { useContext, useEffect } from 'react';
 import ParentSize from '@visx/responsive/lib/components/ParentSize';
 import { ResizeObserverPolyfill } from '@visx/responsive/lib/types';
-import { AxisScaleOutput } from '@visx/axis';
+import { AxisScale, AxisScaleOutput } from '@visx/axis';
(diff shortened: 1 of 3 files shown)


owenbirdsall:
This looks good to me, it would be really helpful if this could be merged in as I imagine it's quite common to want to have multiple points occupying the same x location with varying y locations.

PR #1767, opened 2023-11-09
new(geo, vendor): add `d3-geo@3` to `visx-vendor`, support `geoPath.digits(n)`

kachkaev:
#### :rocket: Enhancements

Closes https://github.com/airbnb/visx/issues/1761

- update `d3-geo` to `3.0.8` to support `geoPath.digits(n)`
- move `d3-geo` dep to `@visx/vendor` because the new version is ESM-only

### `digits={5}`

Screenshot 2023-11-09 at 22 38 54

### `digits={0}`

Screenshot 2023-11-09 at 22 39 10

### no `digits` prop ([defaults to 3](https://github.com/d3/d3-geo/blob/719819bc604eb99438006f922bb586c01283b224/src/path/index.js#L11))

Screenshot 2023-11-09 at 22 39 35


package.json
```diff
@@ -115,6 +115,7 @@
     "ts-node": "9.1.1",
     "typescript": "^3.8.3"
   },
+  "packageManager": "yarn@1.22.19",
```
kachkaev, line 118: I am using [corepack](https://nodejs.org/api/corepack.html) on my computer, which comes with the latest Yarn by default (v4). Specifying `packageManager` helps `corepack` pick the right version and install deps without errors.

packages/visx-geo/src/projections/Projection.tsx
```diff
@@ -149,7 +152,11 @@ export default function Projection<Datum extends GeoPermissibleObjects>({
 
   const path = geoPath().projection(currProjection);
 
-  if (pointRadius) path.pointRadius(pointRadius);
+  // @ts-expect-error Pending `@types/d3-geo` to add digits method (introduced in 3.1.0)
+  // @see https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/67363
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  if (digits !== undefined) path.digits(digits);
```
kachkaev, line 158: Context: https://github.com/d3/d3-geo/releases/tag/v3.1.0

packages/visx-geo/src/projections/Projection.tsx
```diff
@@ -149,7 +152,11 @@ export default function Projection<Datum extends GeoPermissibleObjects>({
...
+  if (pointRadius !== undefined) path.pointRadius(pointRadius);
```
kachkaev, line 159: Side suggestion / question: https://github.com/airbnb/visx/pull/213/files#r1388679342

packages/visx-demo/tsconfig.json
```diff
@@ -14,11 +14,12 @@
     "moduleResolution": "node",
     "composite": true,
     "outDir": "lib",
-    "resolveJsonModule": false,
+    "resolveJsonModule": true,
```
williaster, line 17: ah, I think there's an issue where next forces this when running the demo app locally, but I think it fails the TS generation in CI. let's see if it passes with this change, else I'd have you revert it for merging (and separately need to fix the issue)

packages/visx-geo/src/projections/Projection.tsx
```diff
@@ -93,6 +93,8 @@ export type ProjectionProps<Datum extends GeoPermissibleObjects = GeoPermissible
   graticuleLines?: Omit<GraticuleProps, 'lines'> & { foreground: boolean };
   /** If specified, renders a Graticule outline with the specified props. Specify `graticuleOutline.foreground = true` to be rendered on top of features. */
   graticuleOutline?: Omit<GraticuleProps, 'outline'> & { foreground: boolean };
+  /** Limits the digits after the decimal to the specified number of digits. */
```
williaster, line 96: Wonder if we could tweak the copy slightly to clarify the digits **of what** (not set on this exactly – open to your thoughts!)
```suggestion
  /** Limits the digits for coordinates generated in SVG path strings to the specified number of digits. */
```

packages/visx-geo/src/projections/Projection.tsx
williaster: I think this change is fine (even though `0` shouldn't be a valid input)

packages/visx-geo/src/projections/Projection.tsx
williaster: thanks for starting the thread there 🙏

jest.config.js
```diff
@@ -44,6 +44,6 @@ module.exports = {
   verbose: false,
   testPathIgnorePatterns: ['<rootDir>/packages/visx-demo'],
   transformIgnorePatterns: [
-    'node_modules/(?!(d3-(array|color|format|interpolate|scale|time|time-format)|delaunator|internmap|robust-predicates)/)',
+    'node_modules/(?!(d3-(array|color|format|geo|interpolate|scale|time|time-format)|delaunator|internmap|robust-predicates)/)',
```
williaster, line 47: thanks for finding this – sorry I missed it in my instructions! this isn't ideal to have to do for each ESM-only package, will think on ways to streamline it.

williaster (commented):
Hey @kachkaev this looks great to me overall! No issues from my end except with the one `visx/demo` typescript change I mention (it does look like CI is failing on that so I would try reverting the change and seeing if that fixes it)

packages/visx-demo/tsconfig.json
kachkaev: I reverted this change in https://github.com/airbnb/visx/pull/1767/commits/f7dddefabbf3bca56cfe67191b691c3452776390. Yeah the file was updated by Next when I ran these commands:

```sh
cd packages/visx-demo
NODE_OPTIONS=--openssl-legacy-provider yarn dev
```

(I had to use `NODE_OPTIONS` to avoid `Error: error:0308010C:digital envelope routines::unsupported` in Node 18.18.2)

It would be nice to update `packages/visx-demo/tsconfig.json` but I guess we can do this in a separate DX-focused PR.

kachkaev:
Thanks for your feedback @williaster! I updated the diff, hope the CI will pass this time 🤞

williaster:
Thanks @kachkaev 🙌  I took a pass at updating the PR title + description. Feel free to tweak if you like, only the title is added to the change log (the PR labels determine what it is labeled as there. I think this is just a `minor` version bump / new feature since it shouldn't be a breaking change for consumers – even though the `d3-geo` dep is changing behind the scenes it should only enable net-new functionality)

williaster:
CI passed and Happo diffs look fine!
image

You can't really see diffs by eye, but it's picking up on some (as expected going from random digits to the new default of `3`)
image

williaster:
I think unless you have any other things you want to add, this is probably good to go. Will wait to hear from you before merging/releasing, tho!

kachkaev:
Oh great! Thanks for helping! If the diff and the description look acceptable, let’s get this PR in’ 

Great lib, I’m glad I’ve been able to pay back with this small contribution!

kachkaev:
I guess we can wait for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67384 and remove https://github.com/airbnb/visx/pull/1767#discussion_r1388676698

packages/visx-geo/src/projections/Projection.tsx
kachkaev: Resolved in https://github.com/airbnb/visx/pull/1767/commits/a3f86cb1217221103058e3a9fe3854d1739eb6a3, thanks to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67384

williaster (approved):
this is great, thanks for landing the type change as well!! 🎉 

really appreciate the contribution and hard work. glad to hear you are enjoying `visx` overall :)

PR #1769, opened 2023-11-17
fix: correctly pass through offset from AnimatedGridRows/Columns

henrinormak:
#### :bug: Bug Fix
- Correctly pass through `offset` in AnimatedGridRows
- Correctly pass through `offset` in AnimatedGridColumns


williaster (approved):
thanks for the fix @henrinormak 🙏 ! lgtm, will get this released asap.

Issue #1465, opened 2022-04-05
Brush: "An update was scheduled from inside an update function" warning when using "resetOnEnd"


rarenal:
### Problem
When using [resetOnEnd](https://airbnb.io/visx/docs/brush#Brush_resetOnEnd) in a Brush and after there's a brush end event, the following React warning is thrown
`Warning: An update (setState, replaceState, or forceUpdate) was scheduled from inside an update function. Update functions should be pure, with zero side-effects. Consider using componentDidUpdate or a callback.`

This error is also reproducible if instead of using "resetOnEnd" the brush reference is used inside an "onBrushEnd" to call directly the "reset()" function
```ts
const handleBrushEnd = () => {
    if (brushRef?.current) {
      brushRef.current.reset();
    }
  };
```
### Steps to reproduce
I created [this demo](https://codesandbox.io/s/interesting-feather-dhpfsg?file=/Example.tsx:3468-3603) which basically consists of the Brush example with the "resetOnEnd" attribute.

1. Open CodeSandbox console.
2. Create a new brush area by dragging
3. On the brush end, check how the warning appears in the console.

### Attachments
https://user-images.githubusercontent.com/11938799/161697698-adb7593a-16f5-4cf8-a89c-8918a50995fe.mov

gidesan:
I'm experiencing the same issue

lancefaler:
I also am seeing this same error.  Any insight as to the cause/solution?

kethiri-fm:
I am also facing the same issue. Any updates on this?

rafaelsmgomes:
Also facing the same issue.

Wondering if anyone has got a solution.

GerardGarcia:
Same here, anyone manage to fix it?

VGamezz19:
Same issue here, any update about it?

williaster:
Thanks to @dsdevgit for the fix here! https://github.com/airbnb/visx/pull/1778

PR #1778, opened 2023-12-15
fix(brush): fix issue 'update was scheduled' warning on resetOnEnd #1465

dsdevgit:
#### :bug: Bug Fix

- Fixed issue [#1465](https://github.com/airbnb/visx/issues/1465)

> React warns 'An update was scheduled from inside an update function' in case we set **resetOnEnd = true** or if we call the **reset()** method imperatively.

### The problem:
In short we call **setState** method inside of updater function of another **setState**

```typescript
setState(() => {
    // ...
    //...
    setState(() => {
        //...
    })
})
```

In the [BaseBrush.ts](https://github.com/airbnb/visx/blob/aa8f837c0b69a6de64c9ecaff786cf63e4708437/packages/visx-brush/src/BaseBrush.tsx)
1) **updateBrush()** calls the **setState()** method
2) **handleDragEnd()** or **handleWindowPointerUp()** call the **updateBrush()** with updater function contains **reset()**
3) **reset()** also calls **updateBrush()**

### The Solution:
1) Move brush "idle" state object to **getIdleState()** (state of brush if it is not in use)
2) Avoid calling the **setState** method again by modifying the state passed in the first call (if **resetOnEnd=true** we set 'idle' state from **getIdleState()**)
3) For **reset()** method, we just call **updateBrush()** with  'idle' state

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -113,6 +113,30 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     renderBrushHandles: null,
   };
 
+  getIdleState = () => {
+    const { width, height } = this.props;
+    return {
+      start: { x: 0, y: 0 },
+      end: { x: 0, y: 0 },
+      extent: {
+        x0: -1,
+        x1: -1,
+        y0: -1,
+        y1: -1,
+      },
+      bounds: {
+        x0: 0,
+        x1: width,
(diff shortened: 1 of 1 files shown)


williaster (approved):
hey @dsdevgit thanks for such a great write-up and PR for a long-outstanding bug ❤️ 🙏 

this lgtm, will get it released shortly.

dsdevgit:
Many thanks Chris

Best,
Dennis

PR #1783, opened 2024-01-08
fix: Improve typing in @visx/responsive enhancers

### LINKED ISSUE #1645: withParentSize ugly .d.ts
`withParentSize` is missing a return type annotation:

https://github.com/airbnb/visx/blob/adfa2451cd25e92afe127d23d168895c4b14ff43/packages/visx-responsive/src/enhancers/withParentSize.tsx#L26-L30

and the inferred return type is very ugly:

```tsx
export default function withParentSize<BaseComponentProps extends WithParentSizeProps = {}>(BaseComponent: React.ComponentType<BaseComponentProps & WithParentSizeProvidedProps>): {
    new (props: Readonly<BaseComponentProps & WithParentSizeState>): {
        state: {
            parentWidth: undefined;
            parentHeight: undefined;
        };
        animationFrameID: number;
        resizeObserver: ResizeObserver | undefined;
        container: HTMLDivElement | null;
        componentDidMount(): void;
        componentWillUnmount(): void;
        setRef: (ref: HTMLDivElement) => void;
        resize: import("lodash").DebouncedFunc<({ width, height }: {
            width: number;
            height: number;
        }) => void>;
        render(): JSX.Element;
        context: any;
        setState<K extends "parentWidth" | "parentHeight" | "initialWidth" | "initialHeight">(state: WithParentSizeState | ((prevState: Readonly<WithParentSizeState>, props: Readonly<BaseComponentProps & WithParentSizeState>) => WithParentSizeState | Pick<WithParentSizeState, K> | null) | Pick<WithParentSizeState, K> | null, callback?: (() => void) | undefined): void;
        forceUpdate(callback?: (() => void) | undefined): void;
        readonly props: Readonly<BaseComponentProps & WithParentSizeState> & Readonly<{
            children?: React.ReactNode;
        }>;
        refs: {
            [key: string]: React.ReactInstance;
        };
        shouldComponentUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): boolean;
        componentDidCatch?(error: Error, errorInfo: React.ErrorInfo): void;
        getSnapshotBeforeUpdate?(prevProps: Readonly<BaseComponentProps & WithParentSizeState>, prevState: Readonly<WithParentSizeState>): any;
        componentDidUpdate?(prevProps: Readonly<BaseComponentProps & WithParentSizeState>, prevState: Readonly<WithParentSizeState>, snapshot?: any): void;
        componentWillMount?(): void;
        UNSAFE_componentWillMount?(): void;
        componentWillReceiveProps?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextContext: any): void;
        UNSAFE_componentWillReceiveProps?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextContext: any): void;
        componentWillUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): void;
        UNSAFE_componentWillUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): void;
    };
    new (props: BaseComponentProps & WithParentSizeState, context?: any): {
        state: {
            parentWidth: undefined;
            parentHeight: undefined;
        };
        animationFrameID: number;
        resizeObserver: ResizeObserver | undefined;
        container: HTMLDivElement | null;
        componentDidMount(): void;
        componentWillUnmount(): void;
        setRef: (ref: HTMLDivElement) => void;
        resize: import("lodash").DebouncedFunc<({ width, height }: {
            width: number;
            height: number;
        }) => void>;
        render(): JSX.Element;
        context: any;
        setState<K extends "parentWidth" | "parentHeight" | "initialWidth" | "initialHeight">(state: WithParentSizeState | ((prevState: Readonly<WithParentSizeState>, props: Readonly<BaseComponentProps & WithParentSizeState>) => WithParentSizeState | Pick<WithParentSizeState, K> | null) | Pick<WithParentSizeState, K> | null, callback?: (() => void) | undefined): void;
        forceUpdate(callback?: (() => void) | undefined): void;
        readonly props: Readonly<BaseComponentProps & WithParentSizeState> & Readonly<{
            children?: React.ReactNode;
        }>;
        refs: {
            [key: string]: React.ReactInstance;
        };
        shouldComponentUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): boolean;
        componentDidCatch?(error: Error, errorInfo: React.ErrorInfo): void;
        getSnapshotBeforeUpdate?(prevProps: Readonly<BaseComponentProps & WithParentSizeState>, prevState: Readonly<WithParentSizeState>): any;
        componentDidUpdate?(prevProps: Readonly<BaseComponentProps & WithParentSizeState>, prevState: Readonly<WithParentSizeState>, snapshot?: any): void;
        componentWillMount?(): void;
        UNSAFE_componentWillMount?(): void;
        componentWillReceiveProps?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextContext: any): void;
        UNSAFE_componentWillReceiveProps?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextContext: any): void;
        componentWillUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): void;
        UNSAFE_componentWillUpdate?(nextProps: Readonly<BaseComponentProps & WithParentSizeState>, nextState: Readonly<WithParentSizeState>, nextContext: any): void;
    };
    defaultProps: {
        debounceTime: number;
        enableDebounceLeadingCall: boolean;
    };
    contextType?: React.Context<any> | undefined;
};
```

We don't want our exported components to have this type, so we are forced to cast the result of `withParentSize`.

Could we add a return type annotation such as `React.ComponentType<BaseComponentProps>` to `withParentSize`?

paolostyle:
#### :boom: Breaking Changes

- Technically for TypeScript it does contain a breaking change, you can currently manually pass `parentWidth`/`parentHeight` /`screenWidth`/`screenHeight` prop to the component wrapped in `withParentSize`/`withScreenSize`. Now whether that makes any sense is arguable as it basically invalidates entire purpose of the HOC. The actual runtime behavior did not change, it's just that now TypeScript will scream at you if you try to do this. I can roll this change back if you feel like this is too big of a change.

#### :rocket: Enhancements

- Exported `WithParentSizeProvidedProps` and `WithScreenSizeProvidedProps` to make it easier to properly type props of the component to be enhanced.

#### :memo: Documentation

- Added documentation regarding undocumented config options (i.e. `debounceTime`, `enableDebounceLeadingCall` and their counterparts in `withScreenSize`, `initialWidth`, `initialHeight`) to the readme
- Updated the examples to make it more clear how to properly use it with TypeScript

#### :bug: Bug Fix

- Currently it is not possible to use `withParentSize` and `withScreenSize` HOCs on a component that contains required props different from the ones injected by the HOC without some ugly hacks. Example:
```
import { withParentSize } from '@visx/responsive';

type Props = {
  data: number[];
  parentWidth?: number;
  parentHeight?: number;
};

const Chart = ({
  data, parentWidth, parentHeight,
}: Props) => {
  return null;
};

export default withParentSize(Chart); // error!
// error: Property 'data' is missing in type 'WithParentSizeProps & WithParentSizeState' but required in type 'Props'.ts(2345)

// to make it work we'd have to:
// import { WithParentSizeProps } from '@visx/responsive/lib/enhancers/withParentSize';
// ...
// export default withParentSize<Props & WithParentSizeProps>(Chart);
```

There is no reason why this shouldn't be allowed, so this PR fixes it (i.e. the code above now transpiles correctly). I adjusted a test for this HOC as well to make it more clear that passing other props is now allowed. It also fixes #1645.

#### :house: Internal

- I did not change the actual implementation but I'm happy to contribute `useParentSize` and `useScreenSize` hooks in a separate MR which I believe would be a bit easier to use. These two HOCs could then use these hooks internally instead.
- I dropped `defaultProps` as it makes typing way more difficult and in this particular case it can be easily replaced by `??` operator, it's used in only one place.


packages/visx-responsive/test/withParentSize.test.tsx
```diff
@@ -1,16 +1,18 @@
...
+import { render } from '@testing-library/react';
+import React from 'react';
+import { withParentSize, WithParentSizeProvidedProps } from '../src';
 
-type ComponentProps = {
-  parentWidth?: number;
-  parentHeight?: number;
-};
+interface ComponentProps extends WithParentSizeProvidedProps {
+  // only there to ensure that TS allows enhanced component to have own props, different than the ones passed by the HOC
```
williaster, line 8: thanks for adding tests for this! 🙏

packages/visx-responsive/src/enhancers/withScreenSize.tsx
```diff
@@ -1,7 +1,8 @@
 import debounce from 'lodash/debounce';
 import React from 'react';
+import { Simplify } from '../types';
 
-export type WithScreenSizeProps = {
```
williaster, line 4: wondering if we can still export this, if people are doing deep imports (`import { WithScreenSizeProps } from '@visx/responsive/lib/enhancers/withScreenSize'`) this is theoretically breaking (debatable whether it should be supported or not...)

packages/visx-responsive/src/enhancers/withParentSize.tsx
```diff
@@ -9,33 +9,36 @@ interface PrivateWindow {
   ResizeObserver: ResizeObserverPolyfill;
 }
 
-export type WithParentSizeProps = {
```
williaster, line 12: same comment on continuing to export this

packages/visx-responsive/src/types/index.ts
```diff
@@ -23,3 +23,5 @@ interface ResizeObserverPolyfill {
 }
 
 export { ResizeObserver, ResizeObserverCallback, ResizeObserverPolyfill };
+
+export type Simplify<T> = { [Key in keyof T]: T[Key] } & {};
```
williaster, line 27: curious why this was needed (open to it if it gets things to work nicely, just harder to tell why from the github UI)

packages/visx-responsive/test/withParentSize.test.tsx
```diff
@@ -20,7 +22,7 @@ describe('withParentSize', () => {
 
   test('it should pass parentWidth and parentHeight props to its child', () => {
     const HOC = withParentSize(Component, ResizeObserver);
-    const { getByTestId } = render(<HOC initialWidth={200} initialHeight={200} />);
+    const { getByTestId } = render(<HOC role="img" initialWidth={200} initialHeight={200} />);
```
williaster, line 25: Could also think about an inverse test like this

```suggestion
    // @ts-expect-error ensure unknown types still error
    render(<HOC unknown="prop");
    
    const { getByTestId } = render(<HOC role="img" initialWidth={200} initialHeight={200} />);
```

williaster (commented):
Thanks for the contribution @paolostyle 🙏  This looks great overall, had just a couple minor comments.

> I did not change the actual implementation but I'm happy to contribute useParentSize and useScreenSize hooks in a separate MR which I believe would be a bit easier to use. These two HOCs could then use these hooks internally instead.

Definitely would love to see this! agree hooks are more useful these days (`visx` is showing it's age 🤭)

> 💥 Breaking Changes:  Technically for TypeScript it does contain a breaking change, you can currently manually pass parentWidth/parentHeight /screenWidth/screenHeight prop to the component wrapped in withParentSize/withScreenSize. Now whether that makes any sense is arguable as it basically invalidates entire purpose of the HOC. The actual runtime behavior did not change, it's just that now TypeScript will scream at you

thanks for noting this. also wanted to say I agree this is okay to change as this isn't desired behavior, I wouldn't mark this as a breaking version change.

packages/visx-responsive/src/types/index.ts
paolostyle: I like to put it in library code types for better DX for end users. So the HOC has a return type of `React.ComponentType<WithParentSizeComponentProps<P>>` and without `Simplify` in `WithParentSizeComponentProps<P>` when you hover over a component wrapped in the HOC in the IDE you'd see something like
 ```
const Component: React.ComponentType<WithParentSizeComponentProps<ComponentProps>>
```

With `Simplify`, you'll see 
```
const Component: React.ComponentType<{
    role: string;
    debounceTime?: number | undefined;
    enableDebounceLeadingCall?: boolean | undefined;
    initialWidth?: number | undefined;
    initialHeight?: number | undefined;
}>
```

(`role` is the component's own prop)

File: packages/visx-responsive/src/enhancers/withScreenSize.tsx (hunk 2 above)
Comment (paolostyle) [reply packages/visx-responsive/src/enhancers/withScreenSize.tsx:4]: Yeah, good point. In general, I'm not too fond of deep imports like this and I'd discourage people from using them but you're right, we'd probably break a lot of things with this change considering using this type was the only workaround to the issue I found. I brought it back, annotated it with `@deprecated` JSDoc, and left a to-do comment to remove in the next major version.

File: packages/visx-responsive/src/enhancers/withParentSize.tsx (hunk 3 above)
Comment (paolostyle) [reply packages/visx-responsive/src/enhancers/withParentSize.tsx:12]: Responded there 😄

**Comment by paolostyle** [comment]
Thanks for the review! I'll try to contribute the hooks when I have some free time 😊

**Comment by paolostyle** [comment]
@williaster I updated the PR with `useParentSize` and `useScreenSize` hooks. `useParentSize` is mostly extracted from `<ParentSize>` component, `useParentSize` mostly from `withParentSize`. Some additional notes:
- I did not refactor enhancers to use the hook. I initially did that, but then I realized that the library currently still supports React <16.8 versions, so that would be a breaking change.
- There's a bunch of things I marked as deprecated and frankly I'd really like to get rid of them sooner than later, but as far as I can tell all the changes are backwards compatible
- I added a new way of defining ResizeObserver polyfill, inspired by one of the comments in some merged PR (Would love to credit the person that suggested it but I can't find it anymore...). LMK what do you think. Previous ways of passing it (through props/arguments) are still there.
- Some cosmetic changes to readme
- I wanted to add more comprehensive tests but I just couldn't make it work at all, especially with fake timers and parts that were touching `ResizeObserver`. At this point I think it's just because of the very old version of Jest and/or JSDom, I tried to copy/paste some tests from the @juggle/resize-observer package and they weren't functional (i.e. they were passing, but changing checked values to something else did not break them when it should).

Slightly offtopic - are there some blockers I should be aware of if I wanted to update some very old dependencies (TypeScript, Jest, JSDOM)?

File: packages/visx-responsive/src/hooks/useScreenSize.ts [hunk 6]
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export type UseScreenSizeConfig = DebounceSettings;
+
+const useScreenSize = ({
+  debounceTime = 300,
+  enableDebounceLeadingCall = true,
+}: UseScreenSizeConfig = {}) => {
+  const [screenSize, setScreenSize] = useState<ScreenSize>({
+    width: window.innerWidth,
```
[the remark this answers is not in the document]
williaster: on second read, I think several of the previous `window` refs were inside `componentDidMount` calls, or in `setState` callbacks where the component would have been mounted / on the client. maybe we can go through and change these unprotected refs.

packages/visx-responsive/src/resizeObserver.ts
```diff
@@ -0,0 +1,36 @@
...
+interface PrivateWindow {
+  ResizeObserver: ResizeObserverPolyfill;
+}
+
+let resizeObserverPolyfill: ResizeObserverPolyfill | undefined;
+
+export const getResizeObserver = () =>
+  resizeObserverPolyfill || (window as unknown as PrivateWindow).ResizeObserver;
+
+export const setResizeObserverPolyfill = (polyfill: ResizeObserverPolyfill | undefined) => {
```
williaster, line 34: I have mixed feelings about this approach, I wouldn't want the singleton to cause issues e.g., with SSR/client mismatches or state leaking. Since it's opt-in it could still be okay, but maybe we shouldn't mark the existing config/props as deprecated? Sorry to make you undo that.

williaster (commented):
Hey @paolostyle thanks for taking another pass at all of this and adding the hooks! it looks great overall 😎 

I had a few relatively minor comments, and I think there's a breaking change export that's failing CI. good catch with`react <16.8` support, we use hooks in some packages but mark `^16.8` in the peer dependencies. I wonder if this even compiles now for folks on `16`, even if they don't explicitly use the hooks 🤔 I think because they were there before this change, it's okay.

I don't think there are major things to note regarding older infra that we have. lmk if you hit any issues when tinkering or refactoring, though.

packages/visx-responsive/src/hooks/useParentSize.ts
```diff
@@ -0,0 +1,87 @@
...
+
+          return shouldBail ? existing : incoming;
+        });
+      },
+      debounceTime,
+      { leading: enableDebounceLeadingCall },
+    );
+  }, [debounceTime, enableDebounceLeadingCall, ignoreDimensions]);
+
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
paolostyle: Yes, initially I wanted to ensure that the `initialWidth`/`initialHeight` props are no longer necessary, but I admit I wasn't really thinking about SSR. So I will just revert this change and bring back the original behavior.

packages/visx-responsive/src/hooks/useScreenSize.ts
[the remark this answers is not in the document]
paolostyle: I think I will just bring back the original behavior again, which I believe was simply calling the resize callback on the first render. So these would become 0s/nulls (can't check right now). But I'll check if there are places which require additional checks.

packages/visx-responsive/src/resizeObserver.ts
paolostyle: Yeah I think I will just revert it. The proposed solution is probably quite bad in general, a React context would be a better option but I'm wondering if it's worth the effort, polyfilling ResizeObserver feels like something that might be useful only in tests considering current browser support of this feature, but even then majority of users would probably be happy with a simple mock attached to window with empty methods.

paolostyle:
> good catch with`react <16.8` support, we use hooks in some packages but mark `^16.8` in the peer dependencies. I wonder if this even compiles now for folks on `16`, even if they don't explicitly use the hooks 🤔 I think because they were there before this change, it's okay.

So just to make sure I understand correctly - do you want to me to adjust the enhancers implementations to use the hooks? I have the code stashed already like I mentioned so that doesn't require any effort but just want to make sure that's what you mean.

packages/visx-responsive/src/resizeObserver.ts
williaster: yeah agree, hopefully this is temporary and we can remove it entirely soon 👍

williaster:
> So just to make sure I understand correctly - do you want to me to adjust the enhancers implementations to use the hooks? I have the code that uses the hooks stashed so that doesn't require any effort but just want to make sure that's what you mean.

sorry for not being clear. I just verified that if you deep import `withParentSize` using `react@16` (no hook support), that the app compiles okay. so just to be safe, I would leave the enhancer implementations as they are / don't refactor to hooks in case there is usage like that out there.

paolostyle:
Updated the PR, I think there shouldn't be any build errors and I addressed your feedback.

paolostyle:
Shameless bump - @williaster I would really appreciate if you could take a look at the PR, thanks!

williaster (approved):
thanks for the ping @paolostyle 🙏 and again for all the hard work and iteration here ❤️  lgtm, should be released soon 🚀

packages/visx-responsive/src/index.ts
```diff
@@ -1,4 +1,6 @@
-export { default as ScaleSVG } from './components/ScaleSVG';
 export { default as ParentSize } from './components/ParentSize';
-export { default as withParentSize } from './enhancers/withParentSize';
-export { default as withScreenSize } from './enhancers/withScreenSize';
+export { default as ScaleSVG } from './components/ScaleSVG';
+export { default as withParentSize, WithParentSizeProvidedProps } from './enhancers/withParentSize';
+export { default as withScreenSize, WithScreenSizeProvidedProps } from './enhancers/withScreenSize';
+export { default as useParentSize, UseParentSizeConfig } from './hooks/useParentSize';
+export { default as useScreenSize, UseScreenSizeConfig } from './hooks/useScreenSize';
```
dumbmatter, line 6: These lines seem to be causing an error in TypeScript (if you have it scan node_modules) because those new types are not exported from the transpiled versions of these 4 files.

paolostyle: Ah damn. Sorry about that. Created a PR that should fix this issue.

williaster: oof, thank you. also found a blind spot in our CI, thanks for the quick report and fix 🙏

PR #1791, opened 2024-01-19
fix(axis): include defaultTextProps in Axis labelProps

### LINKED ISSUE #1790: <Axis /> labelProps is documented as Partial, but overwrites values
Setting a fill color like so:

```tsx
<Axis
    labelProps={{
      fill: AXIS_TEXT_COLOR,
    }}
/>
```

overwrites all other label props.

RyKilleen:
#### :bug: Bug Fix

- Enables passing of partial labelProps per the docs

Closes #1790

### DIFF
--- a/packages/visx-axis/src/axis/AxisRenderer.tsx
+++ b/packages/visx-axis/src/axis/AxisRenderer.tsx
@@ -26,7 +26,7 @@ export default function AxisRenderer<Scale extends AxisScale>({
   label = '',
   labelClassName,
   labelOffset = 14,
-  labelProps = defaultTextProps,
+  labelProps,
   orientation = Orientation.bottom,
   scale,
   stroke = '#222',
@@ -42,6 +42,10 @@ export default function AxisRenderer<Scale extends AxisScale>({
   ticks,
   ticksComponent = Ticks,
 }: AxisRendererProps<Scale>) {
+  const labelPropsDefault = {
+    ...defaultTextProps,
+    ...labelProps
+  }
   const tickLabelPropsDefault = {
     ...defaultTextProps,
     ...(typeof tickLabelProps === 'object' ? tickLabelProps : null),
@@ -89,13 +93,13 @@ export default function AxisRenderer<Scale extends AxisScale>({
           className={cx('visx-axis-label', labelClassName)}
           {...getLabelTransform({
             labelOffset,
-            labelProps,
+            labelPropsDefault,
             orientation,
             range: scale.range(),
             tickLabelFontSize: maxTickLabelFontSize,
             tickLength,
           })}
-          {...labelProps}
+          {...labelPropsDefault}
         >
           {label}
         </Text>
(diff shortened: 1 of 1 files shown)


packages/visx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -42,6 +42,10 @@ export default function AxisRenderer<Scale extends AxisScale>({
   ticks,
   ticksComponent = Ticks,
 }: AxisRendererProps<Scale>) {
+  const labelPropsDefault = {
+    ...defaultTextProps,
+    ...labelProps
+  }
```
williaster, line 48: I think linting will complain about the missing `,`. Could potentially tweak the name here as well (super nit, wouldn't have suggested without the `,` fix)

```suggestion
  const combinedLabelProps = {
    ...defaultTextProps,
    ...labelProps,
  }
```

williaster (commented):
Thanks for the fix here/making it consistent with `tickLabelProps`! (that behavior was added relatively recently in https://github.com/airbnb/visx/pull/1662)

Technically this could be thought of as a breaking change as it would change existing behavior if consumers expect to provide all props. However for `tickLabelProps` we just made it a `minor` version bump so we can do that here. I agree this is a better dev x.

packages/visx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -89,13 +93,13 @@ export default function AxisRenderer<Scale extends AxisScale>({
           className={cx('visx-axis-label', labelClassName)}
           {...getLabelTransform({
             labelOffset,
-            labelProps,
+            labelPropsDefault,
```
williaster, line 96: Looks like TS is failing because this is a named argument

```suggestion
            labelProps: combinedLabelProps,
```

RyKilleen:
Thanks @williaster! Suggested changes applied.

packages/visx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -89,13 +93,13 @@ export default function AxisRenderer<Scale extends AxisScale>({
...
             labelOffset,
-            labelProps,
+            labelProps: combinedLabelProps,
             orientation,
             range: scale.range(),
             tickLabelFontSize: maxTickLabelFontSize,
             tickLength,
           })}
-          {...labelProps}
+          {...labelPropsDefault}
```
williaster, line 102:
```suggestion
          {...combinedLabelProps}
```

williaster (commented):
sorry missed one spot that I think needs to include the superset of props. else lgtm! thanks for the followup 🙏

packages/visx-axis/src/axis/AxisRenderer.tsx
```diff
@@ -42,6 +42,10 @@ export default function AxisRenderer<Scale extends AxisScale>({
   ticks,
   ticksComponent = Ticks,
 }: AxisRendererProps<Scale>) {
+  const combinedLabelProps = {
+    ...defaultTextProps,
+    ...labelProps,
+  }
```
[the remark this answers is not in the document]
RyKilleen: No worries, I could have been more thorough in my first pass! Pure hubris to think I could do it through GH UI alone 😄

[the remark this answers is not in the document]
williaster: haha np 😅  thanks for the multiple cycles

williaster (approved):
thanks again @RyKilleen ! 🙏

Issue #1782, opened 2024-01-08
XYChart has a zero width or height, bailing - console.log


mamo-muj:
Hi there, I see why this issue happens although my XYChart appears to show just fine (haven't set a width intentionally). I was wondering if there's a way to remove this log as this shows up in production as well. The source shows `XYChart.js`. 

Happy to provide additional context. Thank you

williaster:
Hey @mamo-muj thanks for checking out `visx`. My guess is that this happens during an initial render of the chart, and after that it has a width and renders correctly. 

definitely open to removing this in a prod env, would you be interested in a PR for that?

mamo-muj:
@williaster yes please that would be great!

sleonia:
@williaster @mamo-muj 
I've made some research. What do you think about adding `babel.config` into `visx-xychart` and transform there env variable. So code might be like this

```javascript

  if (width <= 0 || height <= 0) {
    if (__DEV__) {
      console.info('XYChart has a zero width or height, bailing', { width, height });
    }
    return null;
  }

```

But I'm not sure about babel.config. Should I change global babel.config or create the local one?

https://babeljs.io/docs/babel-plugin-transform-inline-environment-variables

williaster:
I don't think we should add env variables like `__DEV__` to the visx babel config, that should be done by the consuming web application. we could potentially reference `process.env.NODE_ENV === 'development'` which is what [`react` does](https://github.com/search?q=repo%3Afacebook%2Freact%20process.env.NODE_ENV&type=code). 

checking that it equals `development` errs on the side of _not_ logging if the variable is not defined, checking that it equals `production` errs on the side of logging if it's not defined.

hshoff:
fixed in https://github.com/airbnb/visx/pull/1792

PR #1792, opened 2024-01-20
feat(xychart): print console.info only in dev mode

sleonia:
#### :rocket: Enhancements
Print console.info only in developer #1782


williaster (approved):
lgtm, thanks @sleonia ! 🙏

PR #1803, opened 2024-02-22
Proposes to add in the wild to hosted visx site

mcioffi:
#### :rocket: Enhancements

- PR proposes to expose _"in the wild"_ not just in README.md, but also on the hosted site at `airbnb.io/visx/`
- Code changes included creates new nav tab nearly identical to `airbnb.io/visx/gallery`
- Screenshots to external site assets maintained identically to Docusarus [here](https://github.com/facebook/docusaurus/blob/main/website/src/pages/showcase/_components/ShowcaseCard/index.tsx) and also [here](https://www.zachleat.com/web/screenshots/) from 11ty folks — as Docusaurus is actively maintained by **meta/facebook**, should be safe to use these services
- Any new sites can be added in the list in the small data file under `packages/visx-demo/src/showcase/sites.js`

**NOTE**: entire list of _"in the wild"_ here is not 100% comprehensive, and any footnotes by `visx` maintainers should be added to this new page since this is maintained by **airbnb**

Screenshot 2024-02-22 at 17-54-24 visx in the wild

### DIFF
--- a/packages/visx-demo/src/components/GalleryTile.tsx
+++ b/packages/visx-demo/src/components/GalleryTile.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import Link from 'next/link';
+import classNames from 'classnames';
 import ParentSize from '@visx/responsive/lib/components/ParentSize';
 import { WidthAndHeight } from '../types';
 
@@ -13,6 +14,7 @@ type Props<ExampleProps extends WidthAndHeight> = {
   exampleUrl?: string;
   tileStyles?: React.CSSProperties;
   title?: string;
+  tileShort?: boolean;
 };
 
 const renderLinkWrapper = (url: string | undefined, node: React.ReactNode) =>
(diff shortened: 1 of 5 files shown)


williaster (commented):
Hi @mcioffi , thanks for all of the work here and for this interesting direction for in the wild. After discussing with the maintainers we decided we don't want to take on the extra maintenance burden of the `slorber-api-screenshot.netlify.app` API – despite other projects like react using them.

so at this time we won't be merging this. again, thank you for the effort and idea 🙏

mcioffi:
@williaster makes sense 👍. Perhaps you and the visx maintainers might be open to an airbnb self-hosted similar service in the future, as [11ty](https://www.11ty.dev/docs/services/screenshots/) (maintains `https://v1.screenshot.11ty.dev/`) and [docusarus](https://docusaurus.io/) (which maintains `slorber-api-screenshot.netlify.app`)? I believe the latter is actually a carbon copy of the first.

PR #1808, opened 2024-03-08
Revert "new(axis): add shapeRendering prop to Axis"

williaster:
Reverts airbnb/visx#1739 . this actually made the tick line rendering **more blurry** as it overrode https://github.com/airbnb/visx/pull/840

image

by removing this, you can still set it via `axisLineClassName`, and you can set it for tick lines via `tickLineProps`

cc @Lonami


Lonami (commented):
Thanks :)

PR #1809, opened 2024-03-08
fix: explicit type exports in @visx/responsive

paolostyle:
#### :bug: Bug Fix
#1783 introduced a bug where bundlers were treating type exports as non-type exports - I'm quite surprised that happened but also I didn't know that `export type` was introduced in 3.8, which is great because that makes it an easy fix.


paolostyle:
@williaster Sorry to bother you again but that last release broke the package pretty hard - this should help though.

williaster (approved):
thanks for the quick fix here @paolostyle 🙏

PR #1817, opened 2024-04-09
Fix useParentSize not updating width and height

Ti-tanium:
#### :bug: Bug Fix
Fixes: https://github.com/airbnb/visx/issues/1816

#### Description
The `parentRef` is initially `null`, so the observer is not registered on the ref element (`observer.observe(parentRef.current);` not executed) in the initial render. If we don't add `parentRef.current ` to the dependency array, the `useEffect` won't be executed after `parentRef` is assigned.

### DIFF
--- a/packages/visx-responsive/src/hooks/useParentSize.ts
+++ b/packages/visx-responsive/src/hooks/useParentSize.ts
@@ -80,7 +80,7 @@ export default function useParentSize<T extends HTMLElement = HTMLDivElement>({
       observer.disconnect();
       resize.cancel();
     };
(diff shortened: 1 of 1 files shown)


hshoff:
Hi @Ti-tanium thanks for checking out visx and for the PR but we don't want to add `parentRef.current` to the `useEffect` dependency list because mutating the ref doesn't trigger a re-render. See https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/ for more info. I'll follow up for more info in the issue.

Issue #1828, opened 2024-05-09
LegendThreshold: Domain starting with a 1 returns undefined color


ZachBarbre:
Hi y'all, 

I found in ```<LegendThreshold />``` that if you have the domain array of ```scaleThreshold``` start with a 1 the "Less than 1" ```rect``` doesn't color correctly.  While testing it I also found that if you have only two negative items in the domain array you can end up with the same color in the middle legend item and the "more than" legend item. 

You can see it in action here: https://codesandbox.io/p/sandbox/determined-sanne-4j7x4y?file=%2Fsrc%2FApp.js%3A11%2C20

I have a fix if you would like me to make a PR on it.

hshoff:
Hi @ZachBarbre thanks for reporting this bug. Happy to review a PR

ZachBarbre:
@hshoff Awesome. Thank you. PR is up.

hshoff:
Confirmed fixed with `@visx/legend@3.10.3` https://codesandbox.io/p/sandbox/determined-sanne-forked-k8zkm8?file=%2Fpackage.json%3A12%2C5-12%2C29

PR #1831, opened 2024-05-14
Fix LegendThreshold

ZachBarbre:
#### :bug: Bug Fix

- Changed how the labelTransform get an upper and lower value to get the color out of the threshold scale.  
This should guarantee a larger or smaller number than the upper and lower bounds is passed into the scale function to retrieve the correct color.  
- Remove the logical OR from the scale() input. It should now never be undefined and this was the cause of a bug if the the lower bound was 1 making `value` falsy.  
- Because I can't help but make eslint happy, added a type to the domain.  

Issue #1828


hshoff (approved):
LGTM, thanks @ZachBarbre!

PR #1836, opened 2024-05-21
brush: fix brush jumping around after mouseup

tfineberg4:
#### :bug: Bug Fix

**How to induce?** 
resize the brush window by dragging and releasing one handle quickly, or quickly drag the brush window and release (harder to induce this way)

**Impact?**
Handle ends up quite far away from the mouse pointer's location. Sometimes the location of _both_ handles will be updated, resulting in a drastic resize of the brush window.

Two relevant issue have been linked below- I believe this PR should resolve both of them!

[Issue 1](https://github.com/airbnb/visx/issues/1804)
[Issue 2](https://github.com/airbnb/visx/issues/1736)

**Root cause** 
`mouseup` action is _supposed_ to disable subsequent dragging of the handle upon `mousemove`. This is achieved by updating the brush's state. However, when moving quickly, `mousemove` actions can trigger `handleWindowPointerMove` _before_ the state change is complete. This leads to handle movement _after_ `mouseup`. Additionally, the stale state causes the occasionally erratic placement of the handle(s).
 
#### Before 

https://github.com/airbnb/visx/assets/166535835/a1c82e0e-edae-449b-88ca-1259c40eb65a

#### After

Can't induce the bug anymore, even with very quick use of the brush.

https://github.com/airbnb/visx/assets/166535835/53bfc639-feab-4034-8e41-ac34817b13ba

### DIFF
--- a/packages/visx-brush/src/BaseBrush.tsx
+++ b/packages/visx-brush/src/BaseBrush.tsx
@@ -16,6 +16,7 @@ import {
   BrushPageOffset,
 } from './types';
 import { getPageCoordinates } from './utils';
+import debounce from 'lodash/debounce';
 
 type PointerHandlerEvent = React.PointerEvent<SVGRectElement>;
 
@@ -224,7 +225,7 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     }
   };
 
-  handleWindowPointerMove = (event: MouseEvent) => {
+  handleWindowPointerMove = debounce((event: MouseEvent) => {
     const { useWindowMoveEvents } = this.props;
     const { brushingType, isBrushing, brushPageOffset, start } = this.state;
 
@@ -318,7 +319,7 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
         return newState;
       });
     }
-  };
+  }, 1);
 
   getExtent = (start: Partial<Point>, end: Partial<Point>) => {
     const { brushDirection, width, height } = this.props;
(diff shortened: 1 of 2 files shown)


wassgha:
thanks 👍

hshoff:
Hi @tfineberg4, thanks for the PR! The before/after videos look good. Nice work. 

The fix for the root problem looks right to me, but would you mind updating this PR without bringing in the @lodash/debounce dependency? Fine to just add a local debounce function in the file itself. We try to avoid dependencies if we can to keep package size down and reduce maintenance burden.

Thank you again, excited to close these long standing issues!

tfineberg4:
Hey @hshoff ! 

Oops, my bad! Totally thought lodash was already a dependency here. I slipped the debounce function into `utils.ts`. If you don't prefer that, of course I can move it into `BaseBrush.tsx`.

Thanks so much for an awesome set of packages, as well as your quick response here 👍

packages/visx-brush/src/BaseBrush.tsx
```diff
@@ -320,6 +320,8 @@ export default class BaseBrush extends React.Component<BaseBrushProps, BaseBrush
     }
   };
 
+  debouncedHandleWindowPointerMove = debounce(this.handleWindowPointerMove, 1);
```
williaster, line 323: could potentially add a comment for why we need this debounce since it was such a cryptic issue before!

williaster (commented):
+1 thanks for looking into these longstanding issues. 

(side note: love your visuals and the use of `brush` as a ✨  range slider! 😎 )

hshoff (approved):
I pulled your branch down and ran the demo brush component locally and compared it to https://airbnb.io/visx/brush. Can confirm bug fixed.

Interesting to discover the bug isn't present in Safari on https://airbnb.io/visx/brush, but present in Chrome.

Fix looks good to me, thanks again for contributing to visx!

PR #1842, opened 2024-05-28
feat(xychart): improve performance when hovering with tooltip

nikaspran:
#### :rocket: Enhancements

- Improves performance drastically when hovering over XYChart with tooltips on. This is achieved by no longer re-rendering unless any of the props change, as renders are quite expensive with `d3-interpolate-path`

#### :memo: Documentation

- Added an gallery control to select data size for testing

This could also be achieved by doing a comparison for `previousD.current === d` - let me know if that's preferable over using React `memo`.


hshoff (approved):
Hi @nikaspran, love it. Thanks for the PR. Looks good to me

Issue #1845, opened 2024-05-31
Zoom: Disable or override wheel gesture


RamonEspinosa:
I'm working on a page that shows a map similar to [Geo Custom Demo](https://airbnb.io/visx/geo-custom). 

The map projection takes a good chunk of the page's width and height.
This makes it super hard for users to scroll down to see the rest of the page because the map will intercept the wheel event. 
Basically, to see the rest of the page, users need to scroll outside the map.

We thought of two solutions to this problem:
1. Use ctrl + scroll to scroll in the map, similar to what [google maps ](https://developers.google.com/maps/documentation/javascript) do for embedded maps:
![image](https://github.com/airbnb/visx/assets/47266830/2f66757a-d36b-405d-b1ec-3f720c9e8d23)
2. Disable wheel handler and only allow zooming by using the controls ( + / - / reset) 

However, i don't think the `@visx/zoom` package supports either.

PR #1846, opened 2024-06-01
Allow custom gesture handlers in Zoom

### LINKED ISSUE #1798: Zoom: Enable customisation of event handlers
We're using @visx/zoom for a project to zoom/pan a combined visualisation of charts. Basically, it's a list of bar charts which can be scrolled. Additionally, the view can be zoomed/panned. 

So, to overcome the shared interaction of the scroll wheel to scroll or zoom the view, we would like to use the control-key as key modifier to enable the zooming/panning. As the control-key is used by @use-gesture to trigger the pinch-event, we need to modify the settings used by @visx/zoom to initialise useGesture:

``` typescript
useGesture(
    {
      onDragStart: ({ event }) => {
        if (!(event instanceof KeyboardEvent)) dragStart(event);
      },
      onDrag: ({ event, pinching, cancel }) => {
        if (pinching) {
          cancel();
          dragEnd();
        } else if (!(event instanceof KeyboardEvent)) {
          dragMove(event);
        }
      },
      onDragEnd: dragEnd,
      onPinch: handlePinch,
      onWheel: ({ event, active, pinching }) => {
        if (
          // Outside of Safari, the wheel event is fired together with the pinch event
          pinching ||
          // currently onWheelEnd emits one final wheel event which causes 2x scale
          // updates for the last tick. ensuring that the gesture is active avoids this
          !active
        ) {
          return;
        }
        handleWheel(event);
      },
    },
    { target: containerRef, eventOptions: { passive: false }, drag: { filterTaps: true } },
  );
```

So, I would like to suggest to enable the customisation of the useGesture-options as well as the event handlers (e.g. `handleWheel`) through adding additional properties to the Zoom-component, so that we can provide custom handlers/options.

Basically doing something like:

``` jsx
<Zoom<SVGSVGElement>
  height={height}
  width={width}
  onWheel={customHandleWheel}
  useGestureConfig={customUseGestureConfig}
>
  {(zoom) => {
 ...
}}
</Zoom>
```

RamonEspinosa:
#### :rocket: Enhancements

- Introduce 'createGestureHandlers' function prop in the `Zoom` component to enable developers define custom gestures.
This allows for flexibility in disabling wheel events or overriding default pinch gestures, improving customization options.

#### :memo: Documentation

- Added basic description for the new `createGestureHandlers` prop.

Should close #1798 and #1845

### DIFF
--- a/packages/visx-zoom/src/Zoom.tsx
+++ b/packages/visx-zoom/src/Zoom.tsx
@@ -18,6 +18,7 @@ import {
   ScaleSignature,
   ProvidedZoom,
   PinchDelta,
+  CreateGestureHandlers,
 } from './types';
 
 // default prop values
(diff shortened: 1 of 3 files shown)


packages/visx-zoom/src/Zoom.tsx
```diff
@@ -38,6 +39,40 @@ const defaultPinchDelta: PinchDelta = ({ offset: [s], lastOffset: [lastS] }) =>
   scaleY: s - lastS < 0 ? 0.9 : 1.1,
 });
 
+const defaultCreateGestureHandlers: CreateGestureHandlers = <ElementType,>({
```
RamonEspinosa, line 42: Any thoughts on exporting this so devs can do extend the defaults? Example:
```jsx
<Zoom 
    width={400} 
    height={400}  
    createGestureHandlers={(zoom) => ({
      ...defaultCreateGestureHandlers(zoom),
      ...customGestureHandlers(zoom)
    })}
/>
```

PR #1847, opened 2024-06-02
fix: don't include all of lodash

RobinClowers:
#### :boom: Breaking Changes

- None

#### :rocket: Enhancements

- Visx packages now only depend on specific lodash function packages, rather than all of lodash. This affects `visx-responsive`, `visx-text`,  and `visx-xychart`.

#### :memo: Documentation

- None

#### :bug: Bug Fix

- None

#### :house: Internal

- None

### DIFF
--- a/packages/visx-responsive/package.json
+++ b/packages/visx-responsive/package.json
@@ -28,9 +28,9 @@
   },
   "homepage": "https://github.com/airbnb/visx#readme",
   "dependencies": {
-    "@types/lodash": "^4.14.172",
+    "@types/lodash.debounce": "^4.0.9",
     "@types/react": "*",
-    "lodash": "^4.17.21",
+    "lodash.debounce": "^4.0.8",
     "prop-types": "^15.6.1"
   },
   "peerDependencies": {
--- a/packages/visx-responsive/src/enhancers/withParentSize.tsx
+++ b/packages/visx-responsive/src/enhancers/withParentSize.tsx
@@ -1,4 +1,4 @@
-import debounce from 'lodash/debounce';
+import debounce from 'lodash.debounce';
 import React from 'react';
 import {
   DebounceSettings,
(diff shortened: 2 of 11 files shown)


packages/visx-shape/yarn.lock
```diff
@@ -0,0 +1,293 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
williaster, line 1: we avoid committing lock files in specific packages, and just have them in the monorepo root. could you remove these and then re-run `yarn` in the root just to make sure the root lock file is good?

williaster (commented):
thanks for the improvement here @RobinClowers ! 🙌  had a quick comment on the non-root lock files.

packages/visx-shape/yarn.lock
RobinClowers: Ah good call, done.

RobinClowers:
@williaster can we get this merged please? Let me know if there is anything else outstanding.

VIKTORVAV99:
@RobinClowers have you considered replacing lodash entirely with https://es-toolkit.slash.page/?

It should be faster and smaller while being more or less a drop in replacement for lodash functions.

RobinClowers:
@VIKTORVAV99 I hadn't heard of es toolkit, looks like an awesome project!

Since many of these projects only depend on one or two functions, I'm not sure the savings would be significant (compared to all of lodash). Also, there might be subtle differences in the behavior between the libraries. That said, it sounds like a good improvement, but I'm not personally motivated to make the change. Especially since this low risk PR hasn't been merged, I worry that it would be wasted effort.

hshoff:
Lodash dep has been removed in visx 4 release. https://github.com/airbnb/visx/pull/2006

PR #1855, opened 2024-06-21
fix: Add missing `tickValues` prop type

f1multiviewer:
#### :bug: Bug Fix
Fixes an issue where the `tickValues` property wasn't properly being typed, even though it works when being passed.

Related: https://github.com/airbnb/visx/pull/376


packages/visx-grid/src/types.ts
```diff
@@ -33,6 +33,8 @@ export type CommonGridProps = {
   strokeDasharray?: string;
   /** Approximate number of grid lines. Approximate due to d3 alogrithm, specify `tickValues` for precise control. */
   numTicks?: number;
+  /** An array of values that determine the number and values of the ticks. Falls back to `scale.ticks()` or `.domain()`. */
+  tickValues?: ScaleInput<Scale>[];
```
hshoff, line 37: Looks like we need import `ScaleInput` and `Scale` for this.

hshoff (changes requested):
Nice, an F1 app sending PRs during FP2. Needs a small change and we should be good to land this.

packages/visx-grid/src/types.ts
f1multiviewer: I added the import for `ScaleInput`, `Scale` should be a Generic, so should not require an import.

f1multiviewer: Nevermind, it's not generic here, let me check a little more, it might need more changes so it aligns with `AxisProps`

rnintai:
Any updates so far?
I'm truly looking forward to this changes.

rnintai:
> Okay, so I've found that the `tickValues` is explicitly excluded from the types (even though it spreads just fine). Instead, I'm working now on exposing the `rowTickValues` and `columnTickValues` props, which are exposed elsewhere.

@f1multiviewer 
One more, i agree to your suggestion because we may need it for the cases which we want separate the `tickValues`.
Then, i think we need to split `numTicks` to `rowNumTicks` and `colNumTicks` for better use and covering more cases.
(or migrating the prop type by `Array` for both these. e.g. `tickValues={[row, col]}`, `numTicks={[row, col]}`)

PR #1863, opened 2024-07-28
Fix small error in the code example in the "Theming" section

maurer2:
#### :memo: Documentation

Small fix in the documentation on theming.

Cheers


williaster (approved):
thanks @maurer2 !

PR #1889, opened 2024-12-06
Add react@19 to peer dependencies of all packages

darthmaim:
This adds ` || ^19.0.0-0` to all `react` and `react-dom` peer dependencies.

This still uses react@18 in this repo, since the tests don't yet support react 19 (see https://github.com/airbnb/visx/issues/1883#issuecomment-2523069005). I've had to use `18.0.0` instead of the latest 18 release `18.3.1`, because some tests fail because of deprecation warnings with `18.3.1` and `18.0.0` was the version resolved by yarn before, so this doesn't change anything.

> [!CAUTION]
> I've only tested a few packages with react@19 so far, some of them might not work yet.


williaster (approved):
lgtm, will do a pre-release of this. thanks again @darthmaim!

PR #1894, opened 2024-12-19
internal: upgrade to TypeScript ^4.8

williaster:
#### :house: Internal

WIP


config-eslint/typescript.js
```diff
@@ -80,6 +80,7 @@ module.exports = {
           'error',
           { assertionStyle: 'as', objectLiteralTypeAssertions: 'allow-as-parameter' },
         ],
+        '@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports' }],
```
williaster, line 83: enabled auto-fixing `import { FooType } from './foo'` to `import type { FooType } from './foo'`

packages/visx-demo/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "next build && next export",
     "deploy": "rm -rf out && yarn build && cd out && touch .nojekyll && git init && git add . && git commit -m \"Deploy commit\" && git remote add origin git@github.com:airbnb/visx.git && git push -f origin master:gh-pages",
-    "dev": "next",
+    "dev": "NODE_OPTIONS=--openssl-legacy-provider next",
```
williaster, line 9: may need to update next, minimally need to add this same option to the build. don't fully understand why there is an error without this, because our version of next should be using webpack 5 by default 🤔

darthmaim: I've tried to upgrade next to the latest version using `npx @next/codemod upgrade`  back when I made #1889 in an attempt to verify all components with react 19, but when I tried `babel-plugin-typescript-to-proptypes` did not work anymore (everything else did though), so I didn't continue working on it.

I can try again in the next few days based on this branch, maybe some other dependency upgrade made it work in the meantime, or maybe there is an easy replacement for that dependency...

config-eslint/typescript.js
darthmaim: Typescript should not require this afaik unless `verbatimModuleSyntax` is enabled 🤔

But it's a good practice anyway, so 👍

packages/visx-axis/src/axis/Axis.tsx
```diff
@@ -2,12 +2,13 @@ import React from 'react';
...
 import { getTicks, coerceNumber } from '@visx/scale';
-import { SharedAxisProps, AxisScale } from '../types';
+import type { SharedAxisProps, AxisScale } from '../types';
 import AxisRenderer from './AxisRenderer';
 import getTickPosition from '../utils/getTickPosition';
 import getTickFormatter from '../utils/getTickFormatter';
 import createPoint from '../utils/createPoint';
-import Orientation, { OrientationType } from '../constants/orientation';
+import type { OrientationType } from '../constants/orientation';
+import Orientation from '../constants/orientation';
```
darthmaim, line 11: FYI, instead of splitting those lines up you can also do this:

```suggestion
import Orientation, { type OrientationType } from '../constants/orientation';
```

config-eslint/typescript.js
williaster: I think I was getting `*.d.ts is not a module` errors, but actually fixed that in tsconfig with `"moduleResolution": "node16"`

but agree this is best practice/we should do it anyway 👍

packages/visx-axis/src/axis/Axis.tsx
williaster: had just used the eslint autofix here. this doesn't seem to be configurable https://typescript-eslint.io/rules/consistent-type-imports/. no strong opinion from me.

williaster:
everything is good except the typescript -> docs is broken. may not land this until after the holidays.

darthmaim:
> everything is good except the typescript -> docs is broken. may not land this until after the holidays.

I briefly tried to look into fixing the docs and wrote everything down in #1897.

williaster:
Yes apologies - we are blocked by the broken documentation which I haven't had time to investigate further unfortunately. if someone from the community has the expertise and interest in helping it would be very much appreciated!

BearCooder:
What about usind Typedoc: https://typedoc.org/ as documentation generator for TypeScript projects instead of react-docgen-typescript-loader or react-docgen-typescript-plugin? Typedoc is active and well maintained

jbrill:
Is there an update on this? Are we still gated on the typescript generation?

wy19900814:
Any updates on this? Almost a year no update for this library, it is so good, but Airbnb abandoned it?

BearCooder:
Hello @williaster , @skevy could you please take a look at the atttempt from @pergamonster that wanted to solve your issue? So we can finally get a new version with react 19
https://github.com/airbnb/visx/issues/1883#issuecomment-2942578250

hshoff:
Hi all, I'm working on getting this across the finish line.

PR #1895, opened 2024-12-19
Fix typescript error when accessing `error` in `scripts/updateTsReferences.ts`

darthmaim:
#### :house: Internal

- Fix typescript error when accessing `error` in `scripts/updateTsReferences.ts`, this fixes running `yarn type:update-refs` 

---
Targets #1894 by @williaster


williaster (approved):
thanks @darthmaim !!!

PR #1896, opened 2024-12-20
perf(demo): lazy render example tiles

williaster:
#### :rocket: Enhancements

Improve the performance of the initial gallery page load by lazy rendering visible tiles

| Before | After | 
| -- | -- |
| slower  | faster  |

### DIFF
--- a/packages/visx-demo/src/components/GalleryTile.tsx
+++ b/packages/visx-demo/src/components/GalleryTile.tsx
@@ -30,23 +30,24 @@ function useEverVisible() {
     const observer = new IntersectionObserver(
       (entries) => {
         entries.forEach((entry) => {
-          // ever visible
           setIsVisible((visible) => visible || entry.isIntersecting);
         });
       },
       {
         root: null, // viewport is the root
-        threshold: 0.05, // 10% visibility threshold
+        threshold: 0.01,
       },
     );
 
+    let curr;
     if (ref.current) {
+      curr = ref.current;
       observer.observe(ref.current);
     }
 
     return () => {
-      if (ref.current) {
-        observer.unobserve(ref.current);
+      if (curr) {
+        observer.unobserve(curr);
       }
     };
   }, [ref]);
--- a/packages/visx-demo/src/components/GalleryTile.tsx
+++ b/packages/visx-demo/src/components/GalleryTile.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import Link from 'next/link';
 import ParentSize from '@visx/responsive/lib/components/ParentSize';
 import { WidthAndHeight } from '../types';
--- a/packages/visx-demo/src/components/GalleryTile.tsx
+++ b/packages/visx-demo/src/components/GalleryTile.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import Link from 'next/link';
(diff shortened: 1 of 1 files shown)


hshoff (approved):

packages/visx-demo/src/components/GalleryTile.tsx
```diff
@@ -18,6 +18,43 @@ type Props<ExampleProps extends WidthAndHeight> = {
...
+      curr = ref.current;
+      observer.observe(ref.current);
+    }
+
+    return () => {
+      if (curr) {
+        observer.unobserve(curr);
+      }
+    };
+  }, [ref]);
```
hshoff, line 53: Oh, might not want to use a ref as an effect dep. It'll be stable across re-renders and not trigger the effect. https://legacy.reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

packages/visx-demo/src/components/GalleryTile.tsx
```diff
@@ -18,6 +18,43 @@ type Props<ExampleProps extends WidthAndHeight> = {
...
+
+    return () => {
+      if (curr) {
+        observer.unobserve(curr);
+      }
+    };
+  }, []);
+
+  return { everVisible, ref };
+}
```
indiesquidge, line 56: What do you think about this hook instead?

```ts
function useIntersectionObserver() {
  const [entry, setEntry] = useState<IntersectionObserverEntry | null>(null);
  const prevObserver = useRef<IntersectionObserver | null>(null);

  const ref = (node: Element) => {
    if (prevObserver.current) {
      prevObserver.current.disconnect();
      prevObserver.current = null;
    }

    if (node.nodeType === Node.ELEMENT_NODE) {
      const observer = new IntersectionObserver(
        ([entry]) => {
          setEntry(entry);
        },
        { threshold: 0.01, root: null },
      );

      observer.observe(node);
      prevObserver.current = observer;
    }
  };

  return { entry, ref };
}
```


This could be a better approach for a few reasons:
1. I think `useIntersectionObserver` as a name feels like it more directly reflects its functionality, while `useEverVisible` is more abstract.
2. I think using `disconnect` is better because it completely removes the observer and stops watching all elements, ensuring proper cleanup, whereas `unobserve` only stops watching a specific element without cleaning up the observer itself, which can lead to memory leaks if not handled properly.
3. We can return the full `IntersectionObserverEntry`, giving access to more useful visibility data (maybe you don't want or care about this though).
4. The hook is easier to use with a single `customRef` function and fewer concerns about state tracking and effects.

[1]: https://github.com/uidotdev/usehooks/blob/90fbbb4cc085e74e50c36a62a5759a40c62bb98e/index.js#L512

packages/visx-demo/src/components/GalleryTile.tsx
```diff
@@ -28,21 +65,24 @@ export default function GalleryTile<ExampleProps extends WidthAndHeight>({
...
-            <ParentSize>
-              {({ width, height }) =>
-                React.createElement(exampleRenderer, {
-                  width,
-                  height: height + (title || description ? detailsHeight : 0),
-                  ...exampleProps,
-                } as ExampleProps)
-              }
-            </ParentSize>
+            {everVisible && (
```
indiesquidge, line 75: If you went with my [suggestion][1], you'd want to change this to `entry?.isIntersecting`.

[1]: https://github.com/airbnb/visx/pull/1896/files#r1894403694

packages/visx-demo/src/components/GalleryTile.tsx
williaster: I dig it! this is a great suggestion for all the reasons you listed, I was just being lazy (hah!) :)

PR #1903, opened 2025-02-12
feat(xychart): add colorAccessor to AreaSeries

vretamal:
#### :rocket: Enhancements

* Add colorAccessor to XYChart AreaSeries prop that accepts a function that takes the dataKey and returns a color, replacing the default theme color

Based on https://github.com/airbnb/visx/pull/1489


evansimons-chrono:
Hi @hshoff! Thanks for your work on visx.
I'm wondering whether minor patches / feature additions like this one are in a holding pattern for release under the v4 line? I don't spy a v3 release branch so imagining that's the case, but appreciate any insight.

anandfresh (approved):

PR #1907, opened 2025-04-17
Apply margin to Brush

### LINKED ISSUE #1711: Margin has no effect on Brush
I've been having a try with the Brush component and noticed that the margin does not have any effect on the Brush area or calculations.  Even looking at the [Brush-Demo](https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-brush) you can see that the `const brushMargin = { top: 10, bottom: 15, left: 50, right: 20 };` is not moving anytihing inside the `<rect />` for the brush but instead is being used only by the `<g />` element for the `<ChartArea />`

This is a bit problematic and forces you to wrap the charts in a `<g />` in order to correctly align them rather then being able to build the scale using `margin.left` and `innerWidth` like is sometimes common with D3 patterns.

### To Recreate
1. open the [demo in CodeSandbox](https://codesandbox.io/s/github/airbnb/visx/tree/master/packages/visx-demo/src/sandboxes/visx-brush)
2. Go to `line 189` `margin={brushMargin}`
3. Delete the `margin` prop from the `<Brush />`
4. See nothing change in the brush 🤷

valtism:
Fixes https://github.com/airbnb/visx/issues/1711

Applies margin to Brush so that it will shrink to fit inside the margins instead of ignoring them

#### :boom: Breaking Changes

- This probably is a breaking change as it will apply margins that were previously ignored.

#### :rocket: Enhancements

- Brush can use margin now

#### :memo: Documentation

- N/A

#### :bug: Bug Fix

- Fixes margin with Brush

### DIFF
--- a/packages/visx-brush/src/Brush.tsx
+++ b/packages/visx-brush/src/Brush.tsx
@@ -207,8 +207,8 @@ class Brush extends Component<BrushProps> {
     if (brushRegion === 'chart') {
       left = 0;
       top = 0;
-      brushRegionWidth = width;
-      brushRegionHeight = height;
+      brushRegionWidth = width - margin.left - margin.right;
+      brushRegionHeight = height - margin.top - margin.bottom;
     } else if (brushRegion === 'yAxis') {
       top = 0;
       brushRegionHeight = height;


valtism:
Do not merge this just yet. I've found an issue that needs addressing first.

anandfresh (approved):

PR #1947, opened 2025-11-01
Migrate to jsdom

hshoff:
#### :house: Internal

- Replace happy-dom with jsdom (27.0.0) for better DOM compatibility
- Add @vitest/eslint-plugin (^1.4.0) with legacy-recommended preset
- Add vitest/globals types to test tsconfig files to fix TypeScript errors
- Update all vitest configs to use jsdom environment
- Fix TooltipWithBounds test for jsdom style computation


hshoff:
Pinned jsdom to `27.0.0` due to `jsdom@27.0.1` [introducing a breaking change](https://github.com/jsdom/jsdom/blob/main/Changelog.md#2701). Need to upgrade node to upgrade jsdom from here.

PR #1978, opened 2026-01-06
Fix: include index in glyphs prop

TychosDrunkElk:
#### :bug: Bug Fix

- The glyphprops type already has an index, but this doesn't currently get set in the BaseGlyphSeries. Adds the index, and  ensures glyphs satisfy the type


TychosDrunkElk:
Hi @hshoff 👋 This is a minor issue I ran into, currently getting around it by parsing `key`. Since the type for GlyphProps has the index, I'm assuming it just needs to be added here?

anandfresh (approved):

PR #1985, opened 2026-03-10
fix(legend): render Rect shape as SVG for print compatibility

### LINKED ISSUE #1984: Bug: LegendOrdinal `shape="rect"` does not appear in print / print preview
## Summary

When using `LegendOrdinal` with `shape="rect"`, the legend shape (coloured rectangle) is not visible in browser print preview or when printing to PDF. The legend labels still appear; only the rectangular shape indicators are missing. Using `shape="circle"` (or `shape="line"`) works correctly in print.

## Environment

- **Package:** `@visx/legend`
- **Browser:** Chrome (others likely affected)
- **Reproduction:** Use any LegendOrdinal example with `shape="rect"` and open print preview (e.g. [legends docs example](https://visx.airbnb.tech/gallery?pkg=legend)).

## Expected behavior

The legend rectangle shapes should be visible when printing or in print preview, consistent with how circle and line shapes render.

## Actual behavior

The rect shapes do not appear in print preview or in printed output. Circle and line shapes do appear.

Image

Image

## Cause

In `@visx/legend`:

- **Rect** (`shapes/Rect.js`) is implemented as a **`<div>`** with colour applied via CSS **`background`**.
- **Circle** (`shapes/Circle.js`) is implemented as an **`<svg>`** with a **`<circle>`** element using the **`fill`** attribute.

Browsers typically do not print CSS background colours unless the user enables “Background graphics” / “Print background colors and images”. SVG fills are treated as document content and print regardless. So the rect disappears in print while the circle does not.

## Suggested fix

Implement the rect legend shape as an SVG `<rect>` with a `fill` attribute (similar to how Circle uses `<svg>` + `<circle fill={fill}>`), so it prints consistently with the other shapes and does not rely on CSS `background`.

## Additional context

- No custom CSS or print styles are required to reproduce; the issue appears with default visx legend usage.
- This affects any use case where the legend is shown in print or PDF (reports, exports, etc.).

pranay0703:
## Summary

Fixes #1984

The `Rect` legend shape (`shape="rect"`) was implemented as a `<div>` with `background` CSS color. Browsers suppress CSS backgrounds during printing unless the user explicitly enables "Print background colors". SVG `fill` attributes are treated as document content and always print regardless of this setting.

## Changes

- **`packages/visx-legend/src/shapes/Rect.tsx`** — Changed from `<div style={{ background: fill }}>` to `<svg><rect fill={fill}></svg>`, consistent with how `Circle` and `Line` shapes are already implemented.
- **`packages/visx-legend/test/LegendThreshold.test.tsx`** — Updated assertions to query the `<rect>` element and check its `fill` attribute instead of querying `<div>` and checking `background` style.

## Before / After

| | Before | After |
|---|---|---|
| Element | `<div style={{ background: fill }}>` | `<svg><rect fill={fill}></svg>` |
| Screen | ✅ Visible | ✅ Visible |
| Print/PDF | ❌ Invisible | ✅ Visible |

## Testing

All 18 existing tests pass with the updated implementation.

### DIFF
--- a/packages/visx-legend/src/shapes/Rect.tsx
+++ b/packages/visx-legend/src/shapes/Rect.tsx
@@ -12,14 +12,11 @@ export type ShapeRectProps = {
 };
 
 export default function ShapeRect({ fill, width, height, style }: ShapeRectProps) {
+  const cleanWidth = typeof width === 'string' || typeof width === 'undefined' ? 15 : width;
+  const cleanHeight = typeof height === 'string' || typeof height === 'undefined' ? 15 : height;
   return (
-    <div
-      style={{
-        width,
-        height,
-        background: fill,
-        ...style,
-      }}
-    />
+    <svg width={cleanWidth} height={cleanHeight}>
+      <rect width={cleanWidth} height={cleanHeight} fill={fill} style={style} />
+    </svg>
   );
 }
--- a/packages/visx-legend/test/LegendThreshold.test.tsx
+++ b/packages/visx-legend/test/LegendThreshold.test.tsx
@@ -32,7 +32,9 @@ describe('<LegendThreshold />', () => {
     range.forEach((color, index) => {
       const legendItem = thresholdLegend[index];
       const legendShape = legendItem?.querySelector('.visx-legend-shape');
-      expect(legendShape?.querySelector('div')).toHaveStyle(`background: ${color}`);
+      const rectEl = legendShape?.querySelector('rect');
+      expect(rectEl).not.toBeNull();
+      expect(rectEl).toHaveAttribute('fill', color);
     });
   });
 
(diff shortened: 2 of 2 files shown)


packages/visx-legend/src/shapes/Rect.tsx
```diff
@@ -12,14 +12,11 @@ export type ShapeRectProps = {
 };
 
 export default function ShapeRect({ fill, width, height, style }: ShapeRectProps) {
+  const cleanWidth = typeof width === 'string' || typeof width === 'undefined' ? 15 : width;
```
mreizvikh, line 15: What's the magic number 15 and why it's needed here? Previously we passed width and height directly to the element without a fallback value.

PR #2006, opened 2026-06-11
refactor(deps): remove lodash package usage

hshoff:
#### :boom: Breaking Changes

- None

#### :rocket: Enhancements

- Remove direct `lodash` package usage from first-party visx packages and demos.
- Replace `lodash/debounce` usage with a small local debounce helper in `@visx/responsive`.
- Replace `lodash/memoize` usage in `@visx/text` with a small local memoize helper.

#### :memo: Documentation

- Fix the XYChart demo to pass its known `width` into `XYChart`, avoiding a zero-height responsive wrapper that prevented the chart from rendering correctly on the docs page.

#### :bug: Bug Fix

- Fix the XYChart demo/docs page rendering issue found during browser smoke testing.

#### :house: Internal

- Remove direct `lodash` / `@types/lodash` dependencies from affected package manifests and update `yarn.lock`.
- Update the split-line-path demo to use a local `chunk` helper instead of `lodash`.
- Update `@visx/xychart` tests to resolve the local `@visx/responsive` source alias.
- Update package size baseline after rebuilding.
- Verified with `yarn install --immutable`, `yarn lint`, focused `vitest` runs for `@visx/responsive`, `@visx/text`, and `@visx/xychart`, and browser smoke tests for `/gallery`, `/responsive`, `/text`, `/splitlinepath`, and `/xychart`.


hshoff:
One note on the size report: `packages/sizes.json` only tracks each package’s built JS output, not install size or dependency footprint. Since this PR replaces lodash imports with small local helpers, `@visx/responsive` and `@visx/text` include a bit more own-code in their package builds, while `@visx/xychart` is roughly flat/slightly smaller.

The savings show up in install/runtime dependency footprint instead: consumers no longer need `lodash` for these paths. Locally, `lodash` is about ~1.5 MB compressed in the Yarn cache and ~4.9 MB unpacked. For apps that were only pulling lodash through these visx imports, the min+gzip runtime comparison is also favorable: replacing `lodash/debounce` and `lodash/memoize` with the local helpers is roughly ~8.3 KB minified / ~3.4 KB gzip smaller.

So the package size diff is expected: own package JS gets slightly larger because the helper code now lives in visx, but the overall install/runtime story improves by dropping the lodash dependency.

Issue #1856, opened 2024-06-25
Type annotation incomplete for visx/geo fitExtent/fitSize


aendra-rininsland:
These lines for the `fitExtent` and `fitSize` props have `ExtendedFeatureCollection` commented out for some reason, which is preventing me from passing a `FeatureCollection` to my Projection component in TypeScript.

https://github.com/airbnb/visx/blob/master/packages/visx-geo/src/projections/Projection.tsx#L71
https://github.com/airbnb/visx/blob/master/packages/visx-geo/src/projections/Projection.tsx#L76

AFAICT it works fine to pass a FeatureCollection to; I'd submit a PR fixing those lines but given they're commented out was thinking there might be another reason why I'm missing.

Issue #1865, opened 2024-08-02
[BoxPlot] Outlier always draw behind other elements


raphaelspimenta:
Hello everyone!

It seems that the outliers are always draw behind other elements if they cross each other. Is it possible to change it so the outliers are draw after so they can overlap other elements?

![Screenshot 2024-08-02 at 16 41 53](https://github.com/user-attachments/assets/60a0d2c7-7efa-4ef1-9842-246c25edb8e3)

PR #2024, opened 2026-07-13
fix(stats): handle zero-IQR samples in computeStats

### LINKED ISSUE #1772: computeStats throws exception when all statistics align
Some reproductions:
```
computeStats([-1, 0, 0, 0, 0, 0, 1]);
```
```
computeStats([0]);
```

Thrown error:
```
RangeError: Invalid array length
at computeStats (../../../node_modules/@visx/stats/lib/util/computeStats.js:39:14)
```

ATKasem:
## Summary

`computeStats` crashed with `RangeError: Invalid array length` on samples where IQR is 0 (or the inlier range collapses), e.g. `[-1, 0, 0, 0, 0, 0, 1]`, `[0]`, or `[5, 5, 5, 5]`.

Freedman–Diaconis gives `binWidth = 0` in those cases, so `binCount` becomes `NaN` and `new Array(binCount + 2)` throws. This guards that path and returns a single-bin histogram instead. For `n < 2`, quartile medians fall back to the sample median so single-value inputs stay well-defined.

Fixes #1772

## Test plan

- [x] Added regression tests for the issue reproductions
- [x] Existing computeStats cases still pass

### DIFF
--- a/packages/visx-stats/src/util/computeStats.ts
+++ b/packages/visx-stats/src/util/computeStats.ts
@@ -15,12 +15,13 @@ export default function computeStats(numericalArray: number[]) {
   // calculate median of first half i.e. firstQuartile
   const lowerHalfLength = Math.floor(sampleSize / 2);
   const lowerHalf = points.slice(0, lowerHalfLength);
-  const firstQuartile = calcMedian(lowerHalf);
+  // For n < 2 the half-slices are empty; fall back to the sample median.
+  const firstQuartile = lowerHalf.length ? calcMedian(lowerHalf) : median;
 
-  // calculate median of first half i.e. secondQuartile
+  // calculate median of second half i.e. thirdQuartile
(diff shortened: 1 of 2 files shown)


ATKasem:
@hshoff — gentle ping on a few small self-contained fixes I opened on July 13. Happy to close any that aren't a direction you want.

- **#2024** (this one, +59/-20) — `computeStats` throws `RangeError: Invalid array length` on zero-IQR samples like `[5, 5, 5, 5]` or `[0]`; Freedman–Diaconis gives `binWidth = 0`, so `binCount` becomes `NaN`. Fixes #1772.
- **#2025** (+9/-3) — allow `FeatureCollection` in the `fitExtent`/`fitSize` types. Fixes #1856.
- **#2026** (+49/-18) — `BoxPlot` paints outlier circles before the box, so overlapping outliers hide behind the opaque fill (#1865).

They're independent, so they can be taken in any order. No rush — just flagging in case they slipped past.

